PR #1, opened 2020-06-05
feat(test-runner-core): first setup

LarsDenBakker:
This adds the WTR core package. Please check if things are set up correctly for the monorepo.

Some open tasks:
- [ ] Run build and tests in CI
- [ ] Add script to run tests and build for the entire workspace. Lerna used to do this for us, what should we use now?


packages/test-runner-core/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "test-runner-core",
```
bennypowers, line 2: Publishing unscoped?

packages/test-runner-core/src/test-session/TestSessionStatus.ts
```diff
@@ -0,0 +1,13 @@
+export type TestSessionStatus = 'SCHEDULED' | 'INITIALIZING' | 'STARTED' | 'FINISHED';
```
bennypowers, line 1: Why not a const enum?

packages/test-runner-core/package.json
LarsDenBakker: good one

packages/test-runner-core/src/test-session/TestSessionStatus.ts
LarsDenBakker: I avoid enums because they cause trouble when using tools that use single file transforms, and when it's part of the public API I want people to be able to use it just as a string (ex. in a config).

packages/test-runner-core/package.json
LarsDenBakker: Fixed!

packages/test-runner-core/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@web/test-runner-core",
+  "version": "0.0.1",
```
jorenbroekema, line 3: you should probably start this as 0.0.0 and let changesets handle releasing 0.0.1 properly. If you add a changeset to this PR with a patch, this should happen, and you will have 0.0.1 initial release added to your changelog too.

jorenbroekema:
Proposal, have a separate github actions workflow per package.

So we have a publish workflow for the mono repo, and verify workflows for each separate package that only runs based on whether there are changes in the right package.

For this package it would look like:
web-test-runner.yml

```yaml
name: Verify web-test-runner

on:
  pull_request:
    paths:
    - 'packages/web-test-runners/**'

jobs:
  verify:
    name: Verify changes
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [12.x, 13.x]
    steps:
    - uses: actions/checkout@v2
    - name: Setup Node ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - name: Install dependencies
      run: yarn --frozen-lockfile
    - run: cd packages/web-test-runner
    - name: Build
      run: yarn build
    - name: Lint
      run: yarn lint
    - name: Test
      run: yarn test
```
So we have a specific workflow and we only spend build-time on packages where changes have been made.

If you want to run linting always in  the root, you can also have a root workflow that you run additionally:

```yml
name: Verify @web 

on: pull_request

jobs:
  verify:
    name: Verify changes
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [12.x, 13.x]
    steps:
    - uses: actions/checkout@v2
    - name: Setup Node ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - name: Install dependencies
      run: yarn --frozen-lockfile
    - name: Lint
      run: yarn lint
```

jorenbroekema:
In order to release this package @LarsDenBakker  you can run `yarn changeset` locally and go through the changeset menu, which will generate what is called a `changeset` which is basically a generated markdown file inside `.changeset` folder. If you merge that to master, the publish workflow will start running because you pushed to master and something changed inside `.changeset/*.md`, it will run changeset version (to bump as needed and create changelog etc.) and publish to npm, so we should make sure we added the web-test-runner package to npm first https://www.npmjs.com/settings/web/packages . I sadly don't have permission to do this

packages/test-runner-core/package.json
LarsDenBakker: ok

PR #119, opened 2020-06-20
Feat/demo

LarsDenBakker:
Adds demo projects for lit-element and lit-element-ts. Shows example of using esbuild plugin.

Fixes a publishing error in esbuild plugin.

### DIFF
--- a/demo/projects/lit-element-ts/package.json
+++ b/demo/projects/lit-element-ts/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "@web/demo-lit-element-ts",
+  "version": "0.0.0",
+  "description": "Webcomponent lit-element following open-wc recommendations",
+  "license": "MIT",
+  "author": "modern web",
+  "main": "index.js",
+  "module": "index.js",
+  "scripts": {
+    "start": "es-dev-server --app-index demo/projects/lit-element-ts/demo/index.html --node-resolve --open --watch",
+    "test": "wtr test/**/*.test.ts",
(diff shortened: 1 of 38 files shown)


demo/projects/lit-element-ts/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@web/demo-lit-element-ts",
+  "version": "0.0.0",
```
daKmoR, line 3: private: true missing?

demo/projects/lit-element/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "@web/demo-lit-element",
+  "version": "0.0.0",
+  "description": "Webcomponent lit-element following open-wc recommendations",
+  "license": "MIT",
+  "author": "my-element",
+  "main": "index.js",
+  "module": "index.js",
+  "scripts": {
+    "start": "es-dev-server --app-index demo/index.html --node-resolve --open --watch",
+    "test": "wtr test/**/*.test.js --coverage",
```
daKmoR, line 11: I would prefer to using web-test-runner in our demos and documentation as acronyms are always tough especially for new users

demo/projects/lit-element-ts/package.json
LarsDenBakker: yep

demo/projects/lit-element/package.json
LarsDenBakker: done

PR #121, opened 2020-06-24
fix(test-runner): removed duplicate entry in the commands table

Pieterv24:
I removed a duplicate entry of config in the commands table


LarsDenBakker:
Thanks!

PR #122, opened 2020-06-24
fix(test-runner-cli): fixed error when files is defined as string in the config file

Pieterv24:
When a string was used in the 'files' field the validator would throw an error.
I've changed the boolean expression for the error to be thrown to accept both Arrays and Strings that are not empty.

### Truth table (of fixed expression)
Boolean expression: `!(IsString || IsArray) || IsEmpty`
| IsString | IsArray | IsEmpty | Result |
| --- | --- | --- | --- |
| FALSE | FALSE | FALSE | **TRUE** | 
| FALSE | FALSE | TRUE | **TRUE** |
| FALSE | TRUE | FALSE | **FALSE** |
| FALSE | TRUE | TRUE | **TRUE** | 
| TRUE | FALSE | FALSE | **FALSE** |
| TRUE | FALSE | TRUE | **TRUE** |  
| TRUE | TRUE | FALSE | **FALSE** | 
| TRUE | TRUE | TRUE | **TRUE** | 

In this table you can see the only time the error won't be thrown, is when either the IsArray or IsString are true and the String or array is not empty.

### DIFF
--- a/.changeset/warm-tigers-complain.md
+++ b/.changeset/warm-tigers-complain.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-cli': patch
+---
+
+validate config now accepts a string for as files field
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
Thanks!

PR #196, opened 2020-07-14
chore: modify navigation

daKmoR:
- alternative menu with click to open and only show first level

Preview: https://reverent-spence-d26436.netlify.app/


daKmoR:
changed approach

PR #208, opened 2020-07-17
feat: handle inline scripts

thepassle:
I have some duplicate code, need to see how to clean it up a bit

### DIFF
--- a/packages/dev-server-rollup/src/rollupAdapter.ts
+++ b/packages/dev-server-rollup/src/rollupAdapter.ts
@@ -9,6 +9,8 @@ import {
   PluginSyntaxError,
   Context,
 } from '@web/dev-server-core';
+import { queryAll, predicates, getTextContent } from '@web/dev-server-core/dist/dom5';
+import { parse as parseHtml } from 'parse5';
 import { URL, pathToFileURL, fileURLToPath } from 'url';
 import { Plugin as RollupPlugin, TransformPluginContext } from 'rollup';
 import { InputOptions } from 'rollup';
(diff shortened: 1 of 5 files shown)


packages/dev-server-rollup/src/rollupAdapter.ts
```diff
@@ -215,8 +217,57 @@ export function rollupAdapter(
...
+            transformedCode = result;
+          }
+
+          if (typeof result === 'object' && typeof result?.code === 'string') {
+            transformedCode = result.code;
+          }
+
+          if (transformedCode) {
+            transformedFiles.add(context.path);
+            return transformedCode;
```
thepassle, line 262: @LarsDenBakker Im not sure how I should go about this here, returning the transformedCode for a single file makes sense, but here I'm wanting to do the transform on potentially multiple inline scripts, so if I return here, if theres a 2nd inline script, it wont get the transform, because the function has already returned

LarsDenBakker: If you return here the HTML file becomes a JS file :D 

I think you need to replace the content of the HTML script node with the transformed code text. And then send back the updated HTML.

thepassle: > If you return here the HTML file becomes a JS file :D

Its a feature!

thepassle: @LarsDenBakker Alright, I *think* this does what it should be doing now, could you take another look?

Issue #133, opened 2020-06-30
[test-runner-core] reporting


bennypowers:
custom reporting is a useful feature in many cases.

wtr should be able to output TAP-formatted results, which would allow users to pipe into any number of existing tools which consume TAP

PR #228, opened 2020-07-21
Feat/reporters

LarsDenBakker:
Fixes https://github.com/modernweb-dev/web/issues/133


bennypowers:
That's cool, so for TAP or JUnit we'll write new reporters?

LarsDenBakker:
Yep

PR #289, opened 2020-08-02
feat/test-runner-cli/reporter/junit-reporter

bennypowers:
This WIP PR adds a junitReporter

Some notes for the preliminary version:

- ~~here are some missing properties on `TestRunner`~~ implemented these in PR
  - ~~current master is missing a means to identify skipped tests, which is [useful for junit reports](https://help.catchsoftware.com/display/ET/JUnit+Format)~~ this was added on branch. if it lands in master prior to merge, we'll adjust here
  - ~~similarly, missing timing information for each test~~ see previous point
  - ~~PR expects `test.suiteName` to be provided~~ see previous point
- ~~I've created a test fixture that [shells out to execute `wtr`](https://github.com/modernweb-dev/web/pull/289/files#diff-1f718ddf9fe50fc957db80060b760e8fR27), it may be preferable to run the cli from node, but I'm not sure if we have a node api.~~ ~~see https://github.com/modernweb-dev/web/blob/master/packages/test-runner-chrome/test/chromeLauncher.test.ts#L47~~ a more comprehensive solution is pending, once `commands` is implemented. For now, this PR sticks with `child_process`
- ~~see the current diff between expected output and actual~~ tests currently passing (at least locally)
- ~~I added `junitReporter` as a sibling of `defaultReporter`. We may wish to create a separate package~~ done
- ~~"Works on My Machine™️" even though CI is failing. looks like something to do with the delicate build order.~~
  - ~~`/bin/sh: 1: web-test-runner: not found`~~
  - ~~JUnit XML can present key-value "properties" per-test-suite. I'm currently adding `browser.name`, `browser.userAgent`, and `test.fileName`. I made those keys up 🤷 ~~
- ~~(tangential): our eslint config disallows `@ts-expect-error`, which I recommend allowing, as it has specific semantics which are useful (e.g. in test files)~~

- I'm currently doing simple string comparison for the test fixtures. If we know of something like `semantic-dom-diff` that can parse xml, that may be preferable
  - also FYI i'm doing a regex replace to replace unique ids (like wtr session, timestamp, browser ua string) with some known token
- there are some compromises to get tests passing
  - the predicate for "has this test failed" is `!test.passed && test.error`, since as of 3d059c0, all tests are coming back with `passed: false`

## TODO
- [x] Add README
- [x] Polyfill Array#flatMap
- [x] Recurse over nested `describe`


packages/test-runner-cli/src/reporter/junitReporter.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type TestSessionMetadata =
+  Omit<TestSession, 'tests'>;
+
+type TestResultWithMetadata =
+  TestResult & TestSessionMetadata;
+
+type TestResultsWithMetadataByBrowserName = {
+  [key: string]: TestResultWithMetadata[];
```
LarsDenBakker, line 21: I prefer `Record<string, TestResultWithMetadata[]>`, what do you think?

bennypowers: done

bennypowers:
Notes in the PR description have been updated

tsconfig.node-base.json
```diff
@@ -61,6 +61,15 @@
...
-    "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
+    "forceConsistentCasingInFileNames": true, /* Disallow inconsistently-cased references to the same file. */
+    "lib": [
+      "DOM",
+      "DOM.Iterable",
+      "ES6",
+      "ES2017",
+      "ES2019",
+      "ES2019.array",
+      "ScriptHost"
```
LarsDenBakker, line 72: We still support node v10 right now. And I'm not sure node v12 supports es2019 features.

bennypowers: flatMap starts at node v11

I can polyfill at the top of the file, it's only a few lines

bennypowers:
Currently getting this error, but only when I run _both_ test files. If I slap a `.only` on the second test (in junitReporter.test.js), It passes.

```
  {
    "message": "Protocol error (Emulation.setDeviceMetricsOverride): Session closed. Most likely the page has been closed.",
    "stack": "Error: Protocol error (Emulation.setDeviceMetricsOverride): Session closed. Most likely the page has been closed.\n    at CDPSession.send (/Users/bennyp/Documents/web/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:195:35)\n    at EmulationManager.emulateViewport (/Users/bennyp/Documents/web/node_modules/puppeteer-core/lib/cjs/puppeteer/common/EmulationManager.js:20:26)\n    at Page.setViewport (/Users/bennyp/Documents/web/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:952:58)\n    at ChromeLauncherPage.runSession (/Users/bennyp/Documents/web/packages/test-runner-chrome/dist/ChromeLauncherPage.js:49:34)\n    at ChromeLauncher.startSession (/Users/bennyp/Documents/web/packages/test-runner-chrome/dist/ChromeLauncher.js:92:20)\n    at TestScheduler.startSession (/Users/bennyp/Documents/web/packages/test-runner-core/dist/runner/TestScheduler.js:72:42)\n    at TestScheduler.runNextScheduled (/Users/bennyp/Documents/web/packages/test-runner-core/dist/runner/TestScheduler.js:55:18)\n    at TestScheduler.schedule (/Users/bennyp/Documents/web/packages/test-runner-core/dist/runner/TestScheduler.js:39:14)\n    at TestRunner.runTests (/Users/bennyp/Documents/web/packages/test-runner-core/dist/runner/TestRunner.js:76:34)\n    at TestRunner.start (/Users/bennyp/Documents/web/packages/test-runner-core/dist/runner/TestRunner.js:50:18)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at launchTestRunner (/Users/bennyp/Documents/web/packages/test-runner-junit-reporter/test/junitReporter.test.ts:61:3)"
  },
```

PR #292, opened 2020-08-02
chore: make use of changeset action and cache deps

jorenbroekema:
The action works as follows.

- When you create a changeset and push it to master, it does not immediately publish. 
- First, the action creates a PR which contains all the information about what will be released, which dependent packages get a bump, etc. etc.
- When another PR is merged to master also containing a changeset, the auto-created PR is updated automatically by the action.
- When you merge this auto-generated PR, it will publish all the things for you 🎉 so it's a final check basically, and an easy and overview-able way to stack up intents to release.

I also added some caching to the dependencies, which in lion already sped things up significantly (>50% faster).

Lastly, I checked about linked packages. Long story short is that you don't need linked packages for technical reasons usually, the patch bumps of dependents is usually enough. It is more for your users to keep linked packages' versions more "coupled" so that your majors are always aligned. Because one can make the assumption that e.g. "test-runner-core" 2.0.0 cannot possibly work with "test-runner-launcher" 1.5.7 for example so then linked packages can make sense. 

See [here](https://github.com/atlassian/changesets/issues/415) for the issue discussion and [here](https://github.com/ing-bank/lion/pull/839#discussion_r460960837) for additional comment by Andarist who uses/works on changesets a lot

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches:
       - master
-    paths:
-      - '.changeset/*.md'
 
 jobs:
   release:
+    # Prevents changesets action from creating a PR on forks
+    if: github.repository == 'modernweb-dev/web'
     name: Release
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
If I understand things correctly, if I make a patch of `@web/test-runner-core` it will patch bump `@web/test-runner` as well because it has a dependency on it?

jorenbroekema:
Yes, because test-runner is a dependent yep

PR #331, opened 2020-08-08
docs: rename family to about

LarsDenBakker:
It wasn't clear to me initially what the "Family" page was. I think "About" is a more established term, when people want to know more about the people behind the website. We could also perhaps call it "Organization".


daKmoR:
fair enough

PR #366, opened 2020-08-15
chore: add dark mode toggle

daKmoR:


daKmoR:
we postponed the dark mode toggle for now

PR #385, opened 2020-08-20
docs: use HTML test for responsive learn section

LarsDenBakker:


docs/learn/test-runner/responsive.md
```diff
@@ -60,53 +59,87 @@ describe('isMobile', () => {
...
+          expect(color).to.equal('rgb(255, 0, 0)');
+        });
+      });
+    </script>
+  </body>
+</html>
 ```
 
-_In this tests we combine js and css so we will need to put the css into the test site_
+To run HTML tests, we just need to include it in the files we pass on to the test runner in our `package.json` scripts:
```
daKmoR, line 134:
```suggestion
To run HTML tests, we need to include it in the files we pass on to the test runner in our `package.json` scripts:
```

just, easy, simply, ... are words that should be avoided in learn/docs

it sort of has the indication that people reading this are not smart enough as it's just that simple so why you need to follow this docs here? 😅

LarsDenBakker: Makes sense!

PR #388, opened 2020-08-20
docs: various improvements to the learn section

LarsDenBakker:
Some improvements to the learn section, adding some new content and filling some todos.

I'd like to propose naming the "Standards based" menu item "Web development". I think it makes it clearer what this section is about. Our website mentions standards based development all over the place, so it's pretty clear the web development section covers that too.


docs/learn/web-development/es-modules.md
```diff
@@ -2,11 +2,11 @@
 title: ECMAScript Modules
 eleventyNavigation:
   key: ES Modules
-  parent: Standards-Based
+  parent: Web Development
   order: 60
 ---
 
-All modern browsers support standard es modules, using `import` and `export` statements:
+All modern browsers support standard es modules. These are javascript files using `import` and `export` statements:
```
Westbrook, line 9: Does it matter that it's still a module _even if_ it has side effects?

docs/learn/web-development/es-modules.md
```diff
@@ -16,16 +16,49 @@ export function doBar() {
 }
 ```
 
-You can load them from an HTML page using a module script:
+In contrast to regular scripts, variables declared inside modules are scoped to the file itself. This makes them especially useful for writing modular code, explicitly declaring the API surface using exports.
```
Westbrook, line 19: Is there a better word that "regular"?

docs/learn/web-development/es-modules.md
```diff
@@ -16,16 +16,49 @@ export function doBar() {
...
+
+```js
+function loadComponent() {
+  return import('./components/my-component.js');
+}
+```
+
+## File extensions
+
+You will often see modules using the `.mjs` file extension. For the browser, the file extension doesn't matter, but for node js `.mjs` is an indication to execute this code as a module. If the `package.json` has `type="module"` set, node js it will execute `.js` file as a module automatically as well.
```
Westbrook, line 60: Is it worth discussing file extension support at the server level here?

docs/learn/web-development/es-modules.md
```diff
@@ -49,7 +82,7 @@ Both [@web/test-runner](../../docs/test-runner/overview.md) and [@web/dev-server
 
 ## CommonJS modules
 
-CommonJS is the module system of NodeJS and predates the standard es module system. You can recognize them by the use of `require` and `module.exports`:
+CommonJS is the original module system of NodeJS and predates the standard es module system. You can recognize them by the use of `require` and `module.exports`:
```
Westbrook, line 85:
```suggestion
CommonJS is the original module system of NodeJS and predates standard es modules. You can recognize them by the use of `require` and `module.exports`:
```
This reads weird to me, but might not actually need changing.

docs/learn/web-development/es-modules.md
```diff
@@ -59,7 +92,7 @@ module.exports.doBar = function doBar() {
 };
 ```
 
-This doesn't work in the browser, as it doesn't know what to do with CommonJS modules. A lot of code is still written as CommonJS modules. If you want to use a library written in CommonJS, there are a couple of options to look into.
+This doesn't work in the browser, as it doesn't know what to do with CommonJS modules. While node now supports es modules natively, a lot of code is still written as CommonJS modules. If you want to use a library written in CommonJS, there are a couple of options to look into.
```
Westbrook, line 95:
```suggestion
This doesn't work in the browser, as it doesn't know what to do with CommonJS modules. Node now supports es modules natively, however, a lot of code is still written as CommonJS modules. If you want to use a library written in CommonJS, there are a couple of options to look into.
```
Shifts the negative implication to CommonJS and away from es modules?

docs/learn/web-development/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Web Development
+eleventyNavigation:
+  key: Web Development
+  order: 10
+---
+
+Modern browsers are a powerful platform for building websites and applications. We should work with what's available in the browser first before reaching for custom solutions.
```
Westbrook, line 8: "We should" or "We can"...because they are so powerful? Not sure the underlying goals of the content.

docs/learn/web-development/es-modules.md
LarsDenBakker: Then we have to explain side effects too, and it gets a bit wordy. I think it's fine without it.

docs/learn/web-development/es-modules.md
LarsDenBakker: Perhaps "classic"?

docs/learn/web-development/es-modules.md
LarsDenBakker: Yep, I added this in.

docs/learn/web-development/es-modules.md
LarsDenBakker: Adjusted the wording slightly.

docs/learn/web-development/index.md
LarsDenBakker: Perhaps this should be "Our goal is to"? This is taken from the blog post actually.

PR #468, opened 2020-08-29
feat: add @web/rollup-plugin-copy

daKmoR:
Adds a rollup plugin that copies files while retaining the relative folder structure

=> Preview of Docs https://deploy-preview-468--modern-web.netlify.app/docs/building/rollup-plugin-copy/


packages/rollup-plugin-copy/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@web/rollup-plugin-copy",
+  "version": "0.0.0",
```
LarsDenBakker, line 3: Maybe start at v1?

daKmoR: I wanna have at least one beta package to see if everything is alright... e.g. types, all files are included, etc.

bennypowers (approved):

PR #470, opened 2020-08-29
fix: output TypeScript definitions into src for js packages

daKmoR:
What I did:
- output TypeScript definitions into src for js packages
- main entry for commonjs pointed to non-existing file in plugin-rollup-copy
- type exports should be from src in config-loader
- execute linting step before build (so that it won't fail on generated definition files)


daKmoR:
superseded by https://github.com/modernweb-dev/web/pull/479

PR #482, opened 2020-08-30
docs(test-runner): cover ignoring uncovered lines

Westbrook:
Add documentation of `/* c8 ignore next */` to "Guides" and "Docs".

### DIFF
--- a/docs/guides/test-runner/code-coverage/index.md
+++ b/docs/guides/test-runner/code-coverage/index.md
@@ -212,6 +212,10 @@ We now have a failing test but still 100% code coverage.
 
 You should, therefore, see code coverage as a tool that only gives you guidance and help on spotting missing tests, rather than a hard guarantee of code quality.
 
+## Ignoring uncovered lines
+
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
LGTM, should we also add a link to the c8 docs themselves?

docs/guides/test-runner/code-coverage/index.md
```diff
@@ -212,6 +212,10 @@ We now have a failing test but still 100% code coverage.
 
 You should, therefore, see code coverage as a tool that only gives you guidance and help on spotting missing tests, rather than a hard guarantee of code quality.
 
+## Ignoring uncovered lines
+
+In more complex applications, it is likely that you will find yourself creating difficult, if not impossible, to test branches of functionality. While this can absolutely be a pointer to logic that you might look into breaking down into more approachable parts, there will be cases where this is not feasible. In this case you may chose to [ignore lines of code](../../../docs/test-runner/writing-tests/code-coverage.md#ignoring-uncovered-lines) while computing code coverage.
```
daKmoR, line 217: can we somehow include the code snippet that most will need... could be in this sort of thinking
```
In most cases `/* c8 ignore next */` will do if you wanna know more then...
```

Westbrook: Good idea. I was struggling with how to make a reference on both pages without it being the same.

Westbrook:
The c8 docs are pretty repetitive of this content, but let me know what yall think about a link to there. I originally thought to say directly "we use `v8-to-istanbul`", but it seemed like content that could easily go stale and wasn't really adding much.

LarsDenBakker:
@Westbrook I think it's useful for people to know about the underlying tools, because they could look at the original documentation as well.

LarsDenBakker (approved):

PR #506, opened 2020-09-03
fix(config-loader): export map entry for import needs to be valid

daKmoR:
- fix path for module users in node

### DIFF
--- a/packages/config-loader/package.json
+++ b/packages/config-loader/package.json
@@ -44,7 +44,7 @@
   "types": "dist/index.d.ts",
   "exports": {
     ".": {
-      "import": "./index.,js",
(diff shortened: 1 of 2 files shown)


packages/config-loader/package.json
```diff
@@ -44,7 +44,7 @@
   "types": "dist/index.d.ts",
   "exports": {
     ".": {
-      "import": "./index.,js",
```
LarsDenBakker, line 47: `,` is right next to `m` :) :) :)

daKmoR: indeed it was right next to working 🤣

LarsDenBakker: lol

PR #556, opened 2020-09-11
Docs: Fix invalid SauceLabs configuration

marvinhagemeister:
According to the source code we don't need a wrapping `capabilities` object:

https://github.com/modernweb-dev/web/blob/c4b00bba39a35e0e1776c541f6ebea30f065c6e4/packages/test-runner-saucelabs/test-remote/saucelabsLauncher.test.js#L53-L58


LarsDenBakker:
Thanks! Probably a leftover from the browserstack docs :)

We have an issue deploying the website right now, so the update will show up when that's fixed 👍

PR #591, opened 2020-09-17
docs: update test timeout config options

web-padawan:
Looks like the docs need to be updated to reflect actual config.


LarsDenBakker:
Thanks!

PR #616, opened 2020-09-22
fix(rollup-plugin-import-meta-assets): add ESM entrypoint with .mjs file

hsablonniere:
I was having this error:

```
[!] Error: Cannot find module '/home/hsablonniere/dev/clever-components/node_modules/@web/rollup-plugin-import-meta-assets/dist/index.js'. Please verify that the package.json has a valid "main" entry
```

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/package.json
+++ b/packages/rollup-plugin-import-meta-assets/package.json
@@ -14,6 +14,7 @@
   "author": "modern-web",
   "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/rollup-plugin-import-meta-assets",
   "main": "src/index.js",
+  "module": "index.mjs",
   "engines": {
     "node": ">=10.0.0"
   },


daKmoR:
what node version are you using? the existing "exports" field in package.json should be picked up by node 12 and 14 🤔

LarsDenBakker:
Indeed, the module field is for bundlers

hsablonniere:
* Node.js 14.6.0
* npm 6.14.6

daKmoR:
> Indeed, the module field is for bundlers

yeah I keep forgetting that rollup still does not support the exports field 🙈

hsablonniere:
The project in which I'm using the module has `"type": "module"`.

hsablonniere:
OK, my bad, the bug is fixed with this: https://github.com/modernweb-dev/web/commit/68e24b6900cea9a3eb1d2f6767b2a278ce13d451

It's just that the fix is not released yet.

LarsDenBakker:
My bad, I submitted the fix but didn't merge the version change PR from changesets. It's running now!

PR #663, opened 2020-09-30
chore: update rocket to no longer require patching

daKmoR:
## What I did:

1. Update Rocket & remove patch-package + patch file (as it's no longer needed 💪)


LarsDenBakker:
can we also remove https://github.com/modernweb-dev/web/blob/master/patches/%40lion%2Boverlays%2B0.16.6.patch ?

daKmoR:
oups I wrote I removed it but I didn't 😅

now it is 👍

PR #668, opened 2020-10-01
docs: fix copy-paste from another page

web-padawan:
Replacing "browserstack" (which appears to be a copy paste) with Sauce Labs.


LarsDenBakker:
Thanks 👍

PR #684, opened 2020-10-04
feat(dev-server): export option definitions & readCliArgs for Node API

jorenbroekema:
## Use case

In my extension of web-dev-server I have my own CLI arguments that need to work on top of the web-dev-server CLI arguments.

E.g. `my-dev-server --port 8080 --foo="bar"`

```js
if (opts.argv) {
  const myServerDefinitions = [
    {
      name: 'foo',
      type: String,
      description:
        'Foobar',
    },
  ];

  myConfig = {
    ...myConfig,
    ...commandLineArgs(myServerDefinitions, { argv: opts.argv }),
  };
}
```
This would error because it does not recognize `port` as a valid CLI argument.

The fix would be importing the WDS option definitions:
```js
if (opts.argv) {
  const myServerDefinitions = [
    ...optionDefinitions, // <-- adding the WDS definitions
    {
      name: 'foo',
      type: String,
      description:
        'Foobar',
    },
  ];

  myConfig = {
    ...myConfig,
    ...commandLineArgs(myServerDefinitions, { argv: opts.argv }),
  };
}
```

Although not necessary if you use command-line-args `partial` and pass the `_unknown` args to the `argv` prop of startDevServer options param.

### DIFF
--- a/docs/docs/dev-server/node-api.md
+++ b/docs/docs/dev-server/node-api.md
@@ -73,6 +73,45 @@ async function main() {
 main();
 ```
 
+## Combine with your own CLI definitions
+
+If you extend the dev-server and want to use `command-line-args` to add your own CLI definitions, it is recommended to use the [`partial` option](https://github.com/75lb/command-line-args/wiki/Partial-parsing).
+
+```js
+const myDefinitions = [
+  {
+    name: 'foo',
+    type: String,
+    description: 'Bar',
+  },
+];
+
+const myConfig = commandLineArgs(myServerDefinitions, { partial: true });
+```
+
+This will allow you to do:
+
+```sh
+my-dev-server --port 8080 --foo="bar"
+```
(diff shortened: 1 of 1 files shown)


docs/docs/dev-server/node-api.md
```diff
@@ -73,6 +73,46 @@ async function main() {
...
+const myDefinitions = [
+  {
+    name: 'foo',
+    type: String,
+    description:
+      'Bar',
+  },
+];
+
+const myConfig = ...commandLineArgs(myServerDefinitions, { partial: true });
```
LarsDenBakker, line 90: why the `...` here?

jorenbroekema: Erhh oops that's a copy paste error, since in my config in my code I merge it with defaults ;). Changed it

PR #715, opened 2020-10-09
feat(dev-server-core): expose websockets server on the manager

jorenbroekema:
## What I did

1. make the webSocketServer property of the WebSocketsManager public, so it can be used by users of Node API for WDS
2. add documentation about this

### DIFF
--- a/docs/docs/dev-server/node-api.md
+++ b/docs/docs/dev-server/node-api.md
@@ -133,6 +133,25 @@ async function main() {
 main();
 ```
 
+## Websockets Server
+
+The WebSocketsManager is exposed in case you need more fine-grained control. It contains three helpers:
+
+- `sendImport`: imports the given path, executing the module as well as a default export if it exports a function
+- `sendConsoleLog`: logs a message to the browser console of all connected web sockets.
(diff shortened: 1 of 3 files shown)


docs/docs/dev-server/node-api.md
```diff
@@ -133,6 +133,25 @@ async function main() {
 main();
 ```
 
+## Websockets Server
+
+The WebSocketsManager is exposed in case you need more fine-grained control. It contains three helpers:
+
+- `sendImport`: imports the given path, executing the module as well as a default export if it executes a function
```
LarsDenBakker, line 140:
```suggestion
- `sendImport`: imports the given path, executing the module as well as a default export if it exports a function
```

jorenbroekema: fixed

jorenbroekema:
I'm not sure what could be the reason for a sudden failing test in the test-runner-core, is Saucelabs known to be flaky like this?

I can't really debug this locally without a saucelabs account I think right?

LarsDenBakker:
haven't seen it be flaky, but with a remote service it's possible

PR #721, opened 2020-10-10
fix(dev-server-esbuild): upgrade mdn-browser-compat-data to @mdn/brow…

wmadfaa:
…ser-compat-data

## What I did

1. upgrade mdn-browser-compat-data to @mdn/browser-compat-data


LarsDenBakker:
Thanks! I want to double check if everything still works correctly.

PR #729, opened 2020-10-12
fix(rollup-plugin-copy): adds copied files to watch list

benjamind:
Adds the copied files to the rollup watch list, this enables 'watched' rollup bundle scripts.


packages/rollup-plugin-copy/src/copy.js
```diff
@@ -24,6 +24,12 @@ function copy({ patterns = [], rootDir = process.cwd() }) {
   const resolvedRootDir = path.resolve(rootDir);
   return {
     name: '@web/rollup-plugin-copy',
+    async buildStart() {
+      const files = await patternsToFiles(patterns, rootDir);
```
LarsDenBakker, line 28: Can we store the files in a variable and reuse them in `generateBundle`?

benjamind: Good idea, done.

PR #757, opened 2020-10-21
Feature/javascript start test runner

TomzBench:
I am writing a plugin and would like to be able to start the test runner from javascript instead of using the cli.  

The public export to users outside of this package will exports the startTestRunner method and also a TestRunnerConfig interface. But the startTestRunner does not accept the TestRunnerConfig interface. (It accepts some cli like params).

I re-factored startTestRunner to accept a config object in a backward compatible way. (startTestRunner() and additionally startTestRunnerWithOptions().)  Where startTestRunner reuses the startTestRunnerWithOptions.

### DIFF
--- a/packages/test-runner/src/index.ts
+++ b/packages/test-runner/src/index.ts
@@ -1,4 +1,4 @@
 export * from '@web/test-runner-core';
 export { chromeLauncher } from '@web/test-runner-chrome';
 export { defaultReporter } from '@web/test-runner-cli';
-export { TestRunnerConfig, startTestRunner } from './startTestRunner';
+export { TestRunnerConfig, startTestRunner, TestRunnerOptions } from './startTestRunner';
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
Need to look into this one a bit. We have some planned refactorings here, since we're merging the CLI package into this one. I still need to decide what we want to do in terms of public API.

TomzBench:
@LarsDenBakker - Ya there are assumptions in this PR about that and wasn't really sure ...but glad the intent is to  eventually export the routine.

TomzBench:
@LarsDenBakker 

Spawning web-test-runner using spawn seems to work.

```javascript
    const args = [
      './node_modules/.bin/web-test-runner',
      o.testPattern,
      `--config=${o.webTestRunnerConfig}`,
      '--node-resolve',
    ];
    const node = spawn('node', args);
    node.stdout.on('data', (data) => context.logger.info(data.toString()));
    node.stderr.on('data', (data) => context.logger.error(data.toString()));
    node.on('error', (error) => {
      context.logger.error(`[${error.name}] ${error.message}`);
      if (error.stack) context.logger.error(error.stack);
    });
    node.on('exit', (code) => subscriber.next({ success: true, code }));
```

Is this the best way to run the test runner from javascript for now? Or is there a way I can avoid spawning?

TomzBench:
Looks like this feature is added now (with the node api https://modern-web.dev/docs/dev-server/node-api/)

PR #759, opened 2020-10-21
test(dev-server-core): add test for dynamic import with space

ChristianMurphy:
This adds a test case which demonstrates an issue where spacing characters within a dynamic import, causes it to not resolve as expected.
In it's current form this is a bug report, with an included test case.

minimal example of https://github.com/elmsln/issues/issues/420

### DIFF
--- a/packages/dev-server-core/test/plugins/transformModuleImportsPlugin.test.ts
+++ b/packages/dev-server-core/test/plugins/transformModuleImportsPlugin.test.ts
@@ -63,6 +63,7 @@ describe('transformImports()', () => {
       [
         'function lazyLoad() { return import("my-module-2"); }',
         'import("my-module");',
+        'import(\n\t"my-module"\n);',
         'import("./local-module.js");',
       ].join('\n'),
       defaultFilePath,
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Thanks for the PR! If anyone is interested in fixing this, we need to update this code branch to properly handle whitespace: https://github.com/modernweb-dev/web/blob/master/packages/dev-server-core/src/plugins/transformModuleImportsPlugin.ts#L146

There was a PR for this on es-dev-server for inspiration: https://github.com/open-wc/open-wc/pull/1336

ChristianMurphy:
@LarsDenBakker would you be comfortable with an AST parser (like: acorn, babel, or typescript) being leveraged to simplify this?
The number of edge cases, from comments, to spacing, to variations to template strings and string concatenation are pretty complex.
Having an AST or a CST to work with would simplify the process significantly, and all of the mentioned parsers aim to maintain https://github.com/tc39/test262 compliance, and should by extension be spec complaint.

LarsDenBakker:
We avoid these parsers on purpose because it will slow down the dev server a lot. We use https://github.com/guybedford/es-module-lexer as a light WASM parser and this already does the detection of the imports. 

For static imports the rules are very straightforward, there can only be strings inside and es-module-lexer already detects the start and end for us. Dynamic imports are more complex with more edge cases like you mention. However bare imports in dynamic imports aren't as common, so this is a tradeoff we have to make.

LarsDenBakker:
FYI I'm checking to see if we can use the lexer for this https://github.com/guybedford/es-module-lexer/issues/56

LarsDenBakker:
I've fixed handling of at least spaces in https://github.com/modernweb-dev/web/pull/951

PR #779, opened 2020-10-26
fix cli command for typescript init

ChristianUlbrich:
`npx typescript --init` doesn't work; it is `npx tsc --init`

## What I did

1. fix CLI command in docs


LarsDenBakker:
Thanks!

PR #798, opened 2020-11-01
fix(rollup-plugin-html): prevent remote URLs from counting as assets

Westbrook:
## What I did

1. Updated the `isAsset` test to return `false` when the path to the file pointed to a remote location.


LarsDenBakker:
Looks good. We need to add a changeset

LarsDenBakker (approved):

PR #820, opened 2020-11-06
feat(test-runner-visual-regression): selenium support

### LINKED ISSUE #806: [test-runner-visual-regression] Add selenium launcher and SauceLabs support
At Vaadin we use SauceLabs for visual regression tests, so I would like to add `selenium` support.

It seems like this might need the following changes:

1. Expose `driver` publicly and use [WebDriver.findElement](https://github.com/SeleniumHQ/selenium/blob/63d986984f9c5d3bd78f41ef16ff7664cac8763b/javascript/node/selenium-webdriver/lib/webdriver.js#L950):

```js
await driver.findElement(); // add locator function
```

2. Take a screenshot using [WebElement.takeScreenshot](https://github.com/SeleniumHQ/selenium/blob/63d986984f9c5d3bd78f41ef16ff7664cac8763b/javascript/node/selenium-webdriver/lib/webdriver.js#L2316):

```js
const screenshot = await element.takeScreenshot();
```

So I tried to make these changes but apparently `IFrameManager` makes it more complicated than I though.

I'm going to continue investigating how to do it, but would be very grateful for a piece of advice 🙂

web-padawan:
Fixes #806 

## What I did

1. Added Selenium support to `@web/test-runner-visual-regression`

There is now a `takeScreenshot` method which contains all the logic. I'm not so happy about the implementation as it calls `driver.switchTo()` every time to focus the iframe and then switch back to the parent frame.

Any suggestions how this could be improved?

### DIFF
--- a/packages/test-runner-selenium/src/IFrameManager.ts
+++ b/packages/test-runner-selenium/src/IFrameManager.ts
@@ -1,5 +1,5 @@
 import { CoverageMapData, TestRunnerCoreConfig } from '@web/test-runner-core';
-import { WebDriver } from 'selenium-webdriver';
+import { WebDriver, WebElement } from 'selenium-webdriver';
 
 interface BrowserResult {
   testCoverage?: CoverageMapData;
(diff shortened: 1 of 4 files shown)


web-padawan:
Closing in favor of #1054

PR #824, opened 2020-11-07
fix: don't run actions that need secrets on pull requests

stephenwade:
Any pull request from a fork will fail to run the BrowserStack and Sauce Labs tests, because secrets are not shared with actions run on forks.

## What I did

Change the behavior of the BrowserStack and Sauce Labs tests to run only when the PR is from the `modernweb-dev/web` repository.

### DIFF
--- a/.github/workflows/verify-browserstack.yml
+++ b/.github/workflows/verify-browserstack.yml
@@ -1,6 +1,10 @@
 name: Verify Browserstack
 
-on: pull_request
+on:
+  # Run the workflow whenever commits are pushed to the main branch
+  push:
+    branches:
+      - master
 
 jobs:
   verify-browserstack:
--- a/.github/workflows/verify-browserstack.yml
+++ b/.github/workflows/verify-browserstack.yml
@@ -1,15 +1,23 @@
 name: Verify Browserstack
 
 on:
-  # Run the workflow whenever commits are pushed to the main branch
+  # Run the workflow when commits are pushed to the main branch
   push:
     branches:
       - master
+  # Also run when a PR is opened
+  pull_request:
 
 jobs:
   verify-browserstack:
     name: Verify Browserstack
+
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
+
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/verify-browserstack.yml
+++ b/.github/workflows/verify-browserstack.yml
@@ -1,12 +1,6 @@
 name: Verify Browserstack
 
-on:
-  # Run the workflow when commits are pushed to the main branch
-  push:
-    branches:
-      - master
-  # Also run when a PR is opened
-  pull_request:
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
I'd like to keep running those tests from our own pull requests. They're important validations for us. Can we detect forks from the github actions?

abdonrd:
> I'd like to keep running those tests from our own pull requests. They're important validations for us. Can we detect forks from the github actions?

Yes!

https://github.com/Polymer/lit-html/blob/4395e3d45d68cf704155b507b2e0607fc7759a0c/.github/workflows/benchmarks.yml#L10-L13

stephenwade:
@LarsDenBakker I've updated my PR to run the workflows if they're on a PR from the same repository.

Thanks @abdonrd!

.github/workflows/verify-browserstack.yml
```diff
@@ -1,11 +1,23 @@
 name: Verify Browserstack
 
-on: pull_request
+on:
+  # Run the workflow when commits are pushed to the main branch
+  push:
+    branches:
+      - master
+  # Also run when a PR is opened
+  pull_request:
```
daKmoR, line 9:
```suggestion
on: pull_request
```

if they run for our own pull request then it's fine... on master this would only add noise if they are flaky

.github/workflows/verify-saucelabs.yml
```diff
@@ -1,11 +1,23 @@
 name: Verify Saucelabs
 
-on: pull_request
+on:
+  # Run the workflow when commits are pushed to the main branch
+  push:
+    branches:
+      - master
+  # Also run when a PR is opened
+  pull_request:
```
daKmoR, line 9:
```suggestion
on: pull_request
```

.github/workflows/verify-browserstack.yml
stephenwade: My reason was that if you accept a PR from someone else, you never run these tests at all. If that is okay, then we can skip running on `master`.

stephenwade:
@daKmoR I've committed your suggestions.

daKmoR:
nice - thank you 🙇‍♂️

.github/workflows/verify-browserstack.yml
```diff
@@ -5,7 +5,13 @@ on: pull_request
 jobs:
   verify-browserstack:
     name: Verify Browserstack
+
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
abdonrd, line 11: Sorry to revive this after so long, but...
On what occasion does `github.event.pull_request == null` occur?
We just have `on: pull_request` as trigger.

stephenwade: Hi @abdonrd,

I don't know. I actually [copied this from lit](https://github.com/lit/lit/blob/4395e3d45d68cf704155b507b2e0607fc7759a0c/.github/workflows/benchmarks.yml#L10-L13) on your suggestion in https://github.com/modernweb-dev/web/pull/824#issuecomment-723416910, and lit uses `on: [pull_request]` also. Probably it would be fine to just use:

```yaml
    if: github.event.pull_request.head.repo.full_name == github.repository
```

PR #829, opened 2020-11-07
fix(rollup-plugin-html): handle none module script tags

daKmoR:
## What I did

- do not touch `<script>` tags with inline content/code
- treat `<script src="...">` tags as assets

### DIFF
--- a/packages/rollup-plugin-html/src/assets/utils.ts
+++ b/packages/rollup-plugin-html/src/assets/utils.ts
@@ -19,7 +19,7 @@ function isAsset(node: Node) {
       }
       break;
     case 'script':
-      if (getAttribute(node, 'type') !== 'module') {
+      if (getAttribute(node, 'type') !== 'module' && getAttribute(node, 'src')) {
         path = getAttribute(node, 'src') ?? '';
       }
       break;
--- a/packages/rollup-plugin-html/src/assets/utils.ts
+++ b/packages/rollup-plugin-html/src/assets/utils.ts
@@ -18,6 +18,11 @@ function isAsset(node: Node) {
         path = getAttribute(node, 'href') ?? '';
       }
       break;
+    case 'script':
+      if (getAttribute(node, 'type') !== 'module') {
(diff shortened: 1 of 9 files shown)


packages/rollup-plugin-html/src/assets/utils.ts
```diff
@@ -18,6 +18,11 @@ function isAsset(node: Node) {
         path = getAttribute(node, 'href') ?? '';
       }
       break;
+    case 'script':
+      if (getAttribute(node, 'type') !== 'module') {
```
LarsDenBakker, line 22:
```suggestion
      if (getAttribute(node, 'type') !== 'module' && getAttribute(node, 'src')) {
```

LarsDenBakker: no need to extract inline scripts I think?

packages/rollup-plugin-html/src/input/extract/extractModules.ts
```diff
@@ -16,7 +16,10 @@ function createContentHash(content: string) {
 
 export function extractModules(params: ExtractModulesParams) {
   const { document, htmlDir, rootDir } = params;
-  const scriptNodes = findElements(document, e => getTagName(e) === 'script');
+  const scriptNodes = findElements(
+    document,
+    e => getTagName(e) === 'script' && getAttribute(e, 'type') === 'module',
```
LarsDenBakker, line 21: oops 👍

packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
```diff
@@ -243,4 +243,24 @@ describe('extractAssets', () => {
 
     expect(assets.length).to.equal(0);
   });
+
+  it('does treat none module script tags as assets', () => {
```
LarsDenBakker, line 247:
```suggestion
  it('does treat non module script tags as assets', () => {
```

PR #831, opened 2020-11-07
docs: group example naming consistency

nmpcunha:
Guess this makes the example clearer. Hope I'm seeing it in the correct way :)
Otherwise we could have only one group `package-a` and improving the glob to something like `'packages/{a,b}/test/**/*.test.js'`

## What I did

1. Renamed the group name to `package-b` in order to match the directory example.


LarsDenBakker:
Thanks!

PR #838, opened 2020-11-08
Feat/rollup html social urls

daKmoR:
## What I did

1. Identify `<meta property="og:image"` as an asset
2. Allow for a `absoluteBaseUrl` setting which will convert absolute URLs to full absolute URLs for the following tags.

```html
<!-- FROM -->
<meta property="og:image" content="./images/image-social.png">
<link rel="canonical" href="/guides/">
<meta property="og:url" content="/guides/">

<!-- TO -->
<meta property="og:image" content="https://domain.com/assets/image-social-xxx.png">
<link rel="canonical" href="https://domain.com/guides/">
<meta property="og:url" content="https://domain.com/guides/">
```

it's 2 separate commit as `1` is only 6 line changes while `2` is adding a few more things...


docs/docs/building/rollup-plugin-html.md
```diff
@@ -205,6 +205,41 @@ export default {
 
 The HTML plugin does not do any minification by default. You can use a transform function to use a minifier for your HTML or assets.
 
+### Social Media Tags
+
+Some tags/URLs do need to use full absolute URLs (e.g. https://domain.com/guide/).
```
LarsDenBakker, line 210:
```suggestion
Some social media tags require full absolute URLs (e.g. https://domain.com/guide/).
```

docs/docs/building/rollup-plugin-html.md
```diff
@@ -105,7 +105,7 @@ export default {
 
 ### Bundling assets
 
-The HTML plugin will bundle assets referenced from `img` and `link` elements in your HTML. The assets are emitted as rollup assets, and the paths are updated to the rollup output paths.
+The HTML plugin will bundle assets referenced from `img` and `link` and `meta property="og:image"` elements in your HTML. The assets are emitted as rollup assets, and the paths are updated to the rollup output paths.
```
LarsDenBakker, line 108:
```suggestion
The HTML plugin will bundle assets referenced from `img` and `link` and social media tags in your HTML. The assets are emitted as rollup assets, and the paths are updated to the rollup output paths.
```

The code examples shows it handles more than just `og:image`

docs/docs/building/rollup-plugin-html.md
```diff
@@ -234,6 +269,10 @@ export interface RollupPluginHTMLOptions {
   transformHtml?: TransformHtmlFunction | TransformHtmlFunction[];
   /** Whether to extract and bundle assets referenced in HTML. Defaults to true. */
   extractAssets?: boolean;
+  /** Define a full absolute url to your site (e.g. https://domain.com) */
+  siteUrl?: string;
+  /** Whether to set full absolute urls for ['meta[property=og:image]', 'link[rel=canonical]', 'meta[property=og:url]'] or not. Requires a siteUrl to be set. Default to true. */
+  fullAbsoluteUrls?: boolean;
```
LarsDenBakker, line 275:
```suggestion
  absoluteSocialMediaPaths?: boolean;
```

docs/docs/building/rollup-plugin-html.md
```diff
@@ -234,6 +269,10 @@ export interface RollupPluginHTMLOptions {
   transformHtml?: TransformHtmlFunction | TransformHtmlFunction[];
   /** Whether to extract and bundle assets referenced in HTML. Defaults to true. */
   extractAssets?: boolean;
+  /** Define a full absolute url to your site (e.g. https://domain.com) */
+  siteUrl?: string;
```
LarsDenBakker, line 273: I don't think this is a great name. It should at least be website but not everything is a website. It's rather the deployed origin and base path... is there a common name for that?

packages/rollup-plugin-html/src/output/adjustToFullAbsoluteUrls.ts
```diff
@@ -0,0 +1,47 @@
...
+      ) {
+        return true;
+      }
+      return false;
+    default:
+      return false;
+  }
+}
+
+function joinSiteUrlAndPart(siteUrl: string, part: string) {
```
LarsDenBakker, line 25: I think you can use `new URL(url, pathOrUrl).href`

packages/rollup-plugin-html/src/output/adjustToFullAbsoluteUrls.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (part.startsWith('http')) {
+    return part;
+  }
+  if (part === '/') {
+    return siteUrl + part;
+  }
+  return `${siteUrl}/${part}`;
+}
+
+export function adjustToFullAbsoluteUrls(document: Document, siteUrl: string) {
```
LarsDenBakker, line 35:
```suggestion
export function injectAbsoluteUrls(document: Document, siteUrl: string) {
```

Keeps it consistent with the other functions. Perhaps it should be called from `injectedUpdatedAssetPaths` as well.

LarsDenBakker:
I'm not sure about this change, it doesn't feels unintuitive and hard to explain properly from the configuration. But I see that it's useful to have... 

What happens when we extract those social meta tags by default without resolving the URL? Should we throw an error in that case? Or perhaps we should put this option behind a flag and throw when there is no base url:

```js
{
  extract: {
    socialMedia: true
  }
}
```

docs/docs/building/rollup-plugin-html.md
LarsDenBakker: What about just:

```suggestion
  baseUrl?: string;
```

daKmoR: it could be a URL with subfolder as well... e.g. domain.com/my-sub-folder

so it's a full URL to a site/website...

siteUrl or websiteUrl... 
don't really have an idea about a different name 🤔

packages/rollup-plugin-html/src/output/adjustToFullAbsoluteUrls.ts
daKmoR: thought about it - injecting means to me to add new nodes... that's why I used a different name... but I can change it to whatever

LarsDenBakker: You're injecting different paths. But we could call it rewriting as well. At least both should use the same naming.

daKmoR:
it's not about extract... 😅 It's a new last step in outputting the html - it putting full qualified urls (as those are required for those 3 tags)

right now it's a flag that is true by default but it does nothing if there is no site/baseUrl...

so to make those social links work you need to set the site/baseUrl... if you don't only those urls will not be made absolute but maybe you already have it useful in your html... so I think throwing would not be wise...

docs/docs/building/rollup-plugin-html.md
daKmoR: they are urls and not pathes?
how about `absoluteSocialMediaUrls`?

packages/rollup-plugin-html/src/output/adjustToFullAbsoluteUrls.ts
daKmoR: thx worked 👍

docs/docs/building/rollup-plugin-html.md
daKmoR: used `absoluteBaseUrl`

docs/docs/building/rollup-plugin-html.md
daKmoR: used `absoluteSocialMediaUrls`

docs/docs/building/rollup-plugin-html.md
daKmoR: changed

docs/docs/building/rollup-plugin-html.md
daKmoR: changed

packages/rollup-plugin-html/src/output/adjustToFullAbsoluteUrls.ts
daKmoR: used `injectAbsoluteBaseUrl`

LarsDenBakker:
I meant you toggle the extract/bundle behavior, and we always use fully resolved urls. But I guess we can just document the requirement

abdonrd:
While I was creating this issue... https://github.com/IBM/pwa-lit-template/issues/229

Makes sense to also update the `link[rel="canonical"]` and `meta[property="og:url"]`?

daKmoR:
@abdonrd I'm not sure what you mean?

abdonrd:
> @abdonrd I'm not sure what you mean?

The `link[rel="canonical"]` / `meta[property="og:url"]` meta tags also need to have the absolute url.

abdonrd:
Oh, sorry! The new version already does!
I had misread, and I thought it was only about `meta[property="og:image"]`.

PR #906, opened 2020-11-16
(web-dev-hmr) - Windows compatability

JoviDeCroock:
In the `_dependencyTree` you seem to rely on paths being converted to how they look on the OS for Windows this is look a bit different, hence why we can't use posix here.


LarsDenBakker:
Thanks!

Issue #883, opened 2020-11-15
[dev-server-hmr] - Seems to register wrong modules


JoviDeCroock:
Hey 👋 

I was developing the plugin to integrate [`Prefresh`](https://github.com/JoviDeCroock/prefresh/pull/215) with the dev-server now that the hmr spec is there. I was running into some issues when developing on my Windows machine, I noticed that reloads just wouldn't happen, this made me dive into the internals and found the following to happen:

Dependency tree entry:

```
  '/src/app.jsx' => {
    hmrAccepted: true,
    hmrEnabled: true,
    dependencies: Set(0) {},
    dependents: Set(0) {}
  },
```

Update request route:

```
/../C:\Users\jovid\Desktop\prefresh\test\fixture\web-dev-server\src\app.jsx
```

This meant that the hmr-module wouldn't find an update for what i requested and it seemed stuck (nothing was happnening on my screen).

At first I thought this was just a Windows problem but when running the tests inside of the CI, a similar issue seems to be happening.

In case you want to use the fixture from that branch to test with do note that the dependencies are pointing wrong, this is because the puppeteer tests move these to the root of the project. To use it do:

- run `yarn && yarn build` at the root
- Change deps [here](https://github.com/JoviDeCroock/prefresh/pull/215/files#diff-7ab1a45fec1e13b0ef0536cc6707e6b667ed27cb8d56cc090af349325e07971bR12) and add a `../`
- run `yarn && yarn dev` in the fixture dir

Feel free to give pointers where I should look in case you want me to debug somewhere.

43081j:
I tried it locally on windows with WSL and the `accept()` callback does get called, but the flush of updates has no updates, so nothing happens.

presumably you're not using WSL (given the path you saw) so ill try it outside that too if i can.

JoviDeCroock:
Given that it also errors out on CI I wouldn't exclusively tie it to native windows, hmm maybe the error is located somewhere differently. I'll debug further.

43081j:
do you have a link to an example failed CI build?

also, its a bit of a pain right now as there's no way to turn debug logging on. but you can manually alter the server to force it that way... if you edit `node_modules/@web/dev-server/dist/startDevServer.js` and change `debugLogging: false` to `true`, you'll see more useful output.

i can see `app.jsx` does get registered, has `hmrEnabled=true` and ultimately calls the `accept()` callback when it changes. 

in your case, does it never call the callback?

JoviDeCroock:
[...]

PR #908, opened 2020-11-16
(dev-server-hmr) - fix hmr bubbling and a syntax issue in the hmrClientScript

JoviDeCroock:
Fixes a syntax issue in the client script and bubbling by not eagerly clearing deps

Fixes: https://github.com/modernweb-dev/web/issues/883

### DIFF
--- a/packages/dev-server-hmr/src/hmrClientScript.ts
+++ b/packages/dev-server-hmr/src/hmrClientScript.ts
@@ -44,7 +44,7 @@ export class HotModule {
 
     sendMessage({ type: 'hmr:accept', id: this.id });
     this[moduleState] = HmrState.Accepted;
-    this[acceptHandlers].add(callback ?? () => {});
+    this[acceptHandlers].add(callback || function() {});
   }
 
   dispose(handler) {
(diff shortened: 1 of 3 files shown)


packages/dev-server-hmr/src/hmrClientScript.ts
```diff
@@ -44,7 +44,7 @@ export class HotModule {
 
     sendMessage({ type: 'hmr:accept', id: this.id });
     this[moduleState] = HmrState.Accepted;
-    this[acceptHandlers].add(callback ?? () => {});
+    this[acceptHandlers].add(callback || function() {});
```
43081j, line 47: i'd still rather we use the null coalesce operator here, its just missing parentheses:

```ts
add(callback ?? (() => {}));
```

43081j:
good catch, funny because i knew something was wrong around there but couldn't put my finger on what 😂 

also does point out the issue that our client script isn't type checked. but there were reasons that was difficult originally (as it is injected so needs to be a string).

packages/dev-server-hmr/src/hmrClientScript.ts
JoviDeCroock, line 47:
```suggestion
    this[acceptHandlers].add(callback ?? (() => {}));
```

PR #943, opened 2020-11-19
fix(test-runner-cli): avoid under 1 concurrency

benoitlecluse:
Ensures that minimal concurrency is 1. On small VM (with 1 cpu), max concurrency could be 0.5 so the test suite never starts.


LarsDenBakker:
Good catch, thanks a lot!

Issue #942, opened 2020-11-18
(web-dev-server-hmr) - Doesn't register module-dependencies/dependents correctly


JoviDeCroock:
Hey 👋 

When I was testing the current hmr runtime I noticed that bubbling wasn't working correctly, this because of this [line](https://github.com/modernweb-dev/web/blob/master/packages/dev-server-hmr/src/HmrPlugin.ts#L94) this seems specific to UNIX.

Source here is of the following shape: `./greeting.jsx` which makes the regex always bail out.

I can take this up later this week(end) but if any of you have an idea feel free to point me in the right direction! Thanks in advance.

LarsDenBakker:
That specific line is intended to skip over dynamic imports only. Since it's scanning browser code, the line separators are expected t be unix right?

JoviDeCroock:
Yes, but that regex triggers on `./greeting.jsx` when I was debugging it.

LarsDenBakker:
Sorry I meant to say that's the intention of this line. I see that's it's matching that it's skipping `./greeting.jsx` indeed but not `"./greeting.jsx"`. @43081j I think right now this is only checking correctly for dynamic imports but not static imports?

43081j:
yup i wasn't aware at the time that static imports wouldn't have quotes in their path but dynamic imports would.

we should have a check to only assert against the regex when the import is a dynamic one. good catch

JoviDeCroock:
It does seem to work correctly on windows, since hooks bubble up correctly

PR #947, opened 2020-11-19
feat(dev-server-hmr): improve dependency handling

LarsDenBakker:
## What I did

Fixes https://github.com/modernweb-dev/web/issues/942


LarsDenBakker:
@43081j what do you think?

PR #951, opened 2020-11-19
feat(dev-server): improve handling of dynamic imports

LarsDenBakker:
## What I did

Improves handling of dynamic imports, covering more edge cases with whitespace and concatenated strings.

Fix for https://github.com/modernweb-dev/web/pull/759


ChristianMurphy (approved):

packages/dev-server-core/src/plugins/transformModuleImportsPlugin.ts
```diff
@@ -44,6 +46,11 @@ async function resolveConcatenatedImport(
   let pathToResolve = importSpecifier;
   let pathToAppend = '';
 
+  if (pathToResolve.startsWith('/') || pathToResolve.startsWith('./')) {
```
43081j, line 49: would this miss `../foo.js`?

LarsDenBakker: Indeed!

PR #956, opened 2020-11-20
feat(test-runner-selenium): add heartbeat timeout

### LINKED ISSUE #945: [test-runner-saucelabs] Implement heartbeat to prevent timeout in 90 seconds
I have more than 900 tests in `vaadin-grid` repository and they fail in Safari in SauceLabs most of time.

When I tried to debug the tunnel, I got the following error in SauceLabs UI:

```There is a problem with your test.
Test did not see a new command for 90 seconds. Timing out.
```

And in the WTR output the following error is shown:

```
test/scroll-to-index.test.js:
 :x: Browser tests did not start after 120000ms You can increase this timeout with the testsStartTimeout option. Check the browser logs or open the browser in debug mode for more information. 
 :x: UnsupportedOperationError: executeScript: ERROR The test with session id ad4efc3223fa406e88423a1a876ecd0e has already finished, and can't receive further commands. 
      You can learn more at https://app.saucelabs.com/tests/ad4efc3223fa406e88423a1a876ecd0e
      For help, please check https://wiki.saucelabs.com/display/DOCS/Common+Error+Messages
      at parseHttpResponse (/home/runner/work/***/***/node_modules/selenium-webdriver/lib/http.js:580:11)
      at Executor.execute (/home/runner/work/***/***/node_modules/selenium-webdriver/lib/http.js:491:26)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at async WebDriver.execute (/home/runner/work/***/***/node_modules/selenium-webdriver/lib/webdriver.js:700:17)
      at async IFrameManager.stopSession (/home/runner/work/***/***/node_modules/@web/test-runner-selenium/dist/IFrameManager.js:101:29)
```

The Karma launcher has heartbeat logic: https://github.com/karma-runner/karma-sauce-launcher/pull/223

I'm not sure but most likely adding heartbeat would help to solve this problem.

web-padawan:
## What I did

1. Added heartbeat timeout to `test-runner-selenium` for keeping connection alive in SauceLabs

Fixes #945 

Note: it would make more sense to implement this in `@web/test-runner-saucelabs` but that would require some refactoring, e.g. we need to expose `driver` instance, which is currently private, to setup the heartbeat interval.

### DIFF
--- a/packages/test-runner-selenium/src/seleniumLauncher.ts
+++ b/packages/test-runner-selenium/src/seleniumLauncher.ts
@@ -16,6 +16,7 @@ export class SeleniumLauncher implements BrowserLauncher {
   private iframeManager?: IFrameManager;
   private __iframeManagerPromise?: Promise<IFrameManager>;
   private isIE = false;
+  private pendingHeartbeat!: NodeJS.Timeout;
 
   constructor(private driverBuilder: Builder) {}
 
@@ -47,6 +48,7 @@ export class SeleniumLauncher implements BrowserLauncher {
 
   async startSession(id: string, url: string) {
     await this.ensureIframeManagerInitialized();
+    this.heartbeat();
     return this.iframeManager!.queueStartSession(id, url);
   }
 
@@ -62,6 +64,10 @@ export class SeleniumLauncher implements BrowserLauncher {
   }
 
   async stopSession(id: string) {
+    // If there is still pending heartbeat, clear the timeout
+    if (this.pendingHeartbeat) {
+      clearTimeout(this.pendingHeartbeat);
+    }
     return this.iframeManager!.queueStopSession(id);
   }
 
@@ -95,6 +101,25 @@ export class SeleniumLauncher implements BrowserLauncher {
     this.iframeManager = new IFrameManager(this.config, this.driver, this.isIE);
 
     return this.iframeManager;
+  }
+
+  private heartbeat() {
+    const driver = this.driver;
+
+    // Heartbeat function to keep alive sessions on Sauce Labs via WebDriver calls
+    this.pendingHeartbeat = setTimeout(async () => {
+      if (!driver) {
+        return;
+      }
+      try {
+        await driver.getTitle();
+        this.heartbeat();
+      } catch (e) {
+        // Do nothing, just clear the timeout
+        clearTimeout(this.pendingHeartbeat);
+      }
+      return;
+    }, 60000);
   }
 }
 
--- a/packages/test-runner-selenium/src/seleniumLauncher.ts
+++ b/packages/test-runner-selenium/src/seleniumLauncher.ts
@@ -114,7 +114,7 @@ export class SeleniumLauncher implements BrowserLauncher {
         await this.driver.getTitle();
       } catch (e) {
         // Do nothing, just clear the timeout
-        clearTimeout(this.pendingHeartbeat);
+        clearInterval(this.pendingHeartbeat);
       }
       return;
     }, 60000);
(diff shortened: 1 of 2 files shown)


packages/test-runner-selenium/src/seleniumLauncher.ts
```diff
@@ -96,6 +102,25 @@ export class SeleniumLauncher implements BrowserLauncher {
 
     return this.iframeManager;
   }
+
+  private heartbeat() {
+    const driver = this.driver;
+
+    // Heartbeat function to keep alive sessions on Sauce Labs via WebDriver calls
+    this.pendingHeartbeat = setTimeout(async () => {
```
LarsDenBakker, line 110: Should this be a setInterval?

packages/test-runner-selenium/src/seleniumLauncher.ts
```diff
@@ -16,6 +16,7 @@ export class SeleniumLauncher implements BrowserLauncher {
   private iframeManager?: IFrameManager;
   private __iframeManagerPromise?: Promise<IFrameManager>;
   private isIE = false;
+  private pendingHeartbeat!: NodeJS.Timeout;
```
LarsDenBakker, line 19: It's not picking up the nodejs types, perhaps we need to configure the environment. We could also type this as `number | NodeJS.Timeout`

packages/test-runner-selenium/src/seleniumLauncher.ts
```diff
@@ -96,6 +102,25 @@ export class SeleniumLauncher implements BrowserLauncher {
 
     return this.iframeManager;
   }
+
+  private heartbeat() {
+    const driver = this.driver;
+
+    // Heartbeat function to keep alive sessions on Sauce Labs via WebDriver calls
+    this.pendingHeartbeat = setTimeout(async () => {
+      if (!driver) {
```
LarsDenBakker, line 111: I would query the driver from `this` fresh each time in the setTimeout callback. This way we properly pick up reassignment of the property.

packages/test-runner-selenium/src/seleniumLauncher.ts
web-padawan: IIRC the recursive `setTimeout` is a bit more precise but in this case it probably doesn't matter. Changed.

packages/test-runner-selenium/src/seleniumLauncher.ts
web-padawan: Updated to use `ReturnType<typeof setInterval>` which appears to work.

packages/test-runner-selenium/src/seleniumLauncher.ts
web-padawan: Done.

.changeset/loud-toes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web/test-runner-selenium': patch
```
LarsDenBakker, line 2:
```suggestion
'@web/test-runner-selenium': patch
'@web/test-runner-browserstack': patch
'@web/test-runner-saucelabs': patch
```

packages/test-runner-selenium/src/seleniumLauncher.ts
LarsDenBakker: Yeah recursive would be fine too, but it wasn't recursive from what I saw. It was started when a session starts, but if a session takes a few minutes to start it would only trigger one heartbeat.

.changeset/loud-toes-cheer.md
web-padawan: I just force pushed the PR to add another changeset 🙈

PR #982, opened 2020-11-21
docs(html-tests): fix error when following the docs

Krutsch:
## What I did

Add a package to the imports so that `expect` will be defined.


docs/docs/test-runner/writing-tests/html-tests.md
```diff
@@ -9,6 +9,7 @@ With mocha, you need to define your tests inside the `runTests` function:
   <body>
     <script type="module">
       import { runTests } from '@web/test-runner-mocha';
+      import { expect } from "@esm-bundle/chai";
```
LarsDenBakker, line 12:
```suggestion
      import { expect } from '@esm-bundle/chai';
```

Linter is complaining about this :)

Issue #984, opened 2020-11-21
[dev-server-hmr] update mtime for replaced modules


LarsDenBakker:
When we replace a module because of one of it's dependencies changed, we should update the imports to these dependencies to include the mtimes that a new version of the dependency is downloaded as well.

CC @43081j @JoviDeCroock

PR #994, opened 2020-11-22
feat(dev-server-hmr): update bubbled dependencies

LarsDenBakker:
## What I did

Fixes https://github.com/modernweb-dev/web/issues/984.


packages/dev-server-hmr/src/HmrPlugin.ts
```diff
@@ -228,7 +247,18 @@ export class HmrPlugin implements Plugin {
     }
 
     // Nothing left to try
-    this._broadcast({ type: 'hmr:reload' });
+    if (dependents.size === 0) {
```
LarsDenBakker, line 250: @43081j I added this to avoid bubbled modules to always trigger a reload, what do you think?

PR #1008, opened 2020-11-24
feat(rollup-plugin-html): include video and source assets

jorenbroekema:


packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -650,10 +653,13 @@ describe('rollup-plugin-html', () => {
...
       'image-b.svg',
+      'image-d.png',
+      'video.mp4',
     ];
 
     for (const name of expectedAssets) {
       const asset = getAsset(output, name);
+      console.log(asset, name);
       expect(asset).to.exist;
       expect(asset.source).to.exist;
```
daKmoR, line 664: console?

jorenbroekema: Yeah just for debugging purposes, this test is failing in a very strange way, suddenly it cannot find the image-c asset as soon as I add image-d. And I have no idea why 😭  I tried to play around with it and change stuff around but this bug keeps happening.

jorenbroekema:
I think I need some help with this PR, I cannot figure out why the tests are failing.

As soon as I add image-d (poster attr for video), suddenly it cannot find the image-c asset, and I can't figure out why it behaves that way :\

I kinda need this PR for my site which uses <video> & <source> elements

gdbaldw:
@jorenbroekema maybe this PR https://github.com/modernweb-dev/web/pull/1541 meets your needs for `video` and `source` tags?

jorenbroekema:
@gdbaldw yep, very nice. I still think this PR should be done at some point to make those tags work by default without needing to specify them manually, since they are native html tags that contain a reference to an asset, but I've given up making the flaky test work :P so your PR will at least enable me to extract those video assets

PR #1024, opened 2020-11-28
fix: don't clear scrollback history

stephenwade:
## What I did

- Updated the ANSI escape codes in `dev-server` and `test-runner-cli` so it doesn't clear the scrollback history.

When I'm developing, I sometimes need to scroll up to see the results of a previous command. This isn't possible if `dev-server` clears the scrollback history. In addition, my terminal emulator (iTerm) blocks this escape code from working by default and displays a warning message. I think `dev-server` should not try to clear the scrollback history.

Screenshot of the iTerm terminal app with an alert message: 'A control sequence attempted to clear scrollback history. Allow this in the future?'

The ANSI escape codes used are the following:
- `\x1B[2J`: clear entire screen
- `\x1B[3J`: clear entire screen and erase scrollback history
- `\x1B[H`: move the cursor to the top-left corner

I deleted the second escape code. This still clears the screen and moves the cursor, but it doesn't erase the scrollback history.

I verified that this change works on my Mac with both iTerm and Terminal.app.

### DIFF
--- a/.changeset/unlucky-ads-share.md
+++ b/.changeset/unlucky-ads-share.md
@@ -1,0 +1,6 @@
+---
+'@web/dev-server': patch
+'@web/test-runner-cli': patch
+---
+
+don't clear scrollback buffer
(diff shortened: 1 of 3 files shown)


stephenwade:
@LarsDenBakker Can you take a look at this?

LarsDenBakker:
I think it's a good change, just want to test it a bit on a few different systems.

LarsDenBakker:
I had to revert this in https://github.com/modernweb-dev/web/pull/1024 because it was creating double logs when scrolling up. I notice that both Jest and TSC also don't allow scrolling up either. Perhaps they use a different method, something to look into.

PR #1027, opened 2020-11-30
Extract IFrameManager to separate package

web-padawan:
## What I did

1. Added `@web/test-runner-iframe-manager` package to be used by Selenium and WebdriverIO launchers
2. Updated `@web/test-runner-selenium` launcher to use new `IFrameManager` API.


web-padawan:
As discussed in person, I will submit a separate PR for `@web/test-runner-webdriverio` which will include this logic.

PR #1038, opened 2020-12-01
Chore/update selenium

LarsDenBakker:
## What I did

1.


web-padawan:
The dependency is now updated on master so this PR can be closed.

PR #1066, opened 2020-12-07
docs: add dev server guides

LarsDenBakker:
## What I did

This adds a guides section in the docs for Web Dev Server.


LarsDenBakker:
@daKmoR I moved the "using plugins" from the test runner to the dev server section, and for the "writing plugins" part there is now a guide to use a rollup plugin to do the same.

docs/docs/dev-server/overview.md
```diff
@@ -7,7 +7,7 @@ eleventyNavigation:
   order: 1
 ---
 
-Dev Server for web applications, ideal for buildless es module workflows. Optionally supports simple code transformations.
+Web dev server helps developing for the web, using native browser features like es modules. It is ideal for buildless workflows, and has a plugin architecture for light code transformations.
```
Westbrook, line 10: Do we want to use/establish a "how we talk about web dev server" standard? Is is `Web dev server`, `Web Dev Server`, `@web/dev-server`, etc...this will help the people who wright State of JS 2021.

docs/docs/dev-server/plugins/storybook.md
```diff
@@ -1,12 +1,10 @@
 # Dev Server >> Plugins >> Storybook ||7
 
-> This project is currently in alpha. Expect updates and breaking changes.
```
Westbrook, line 3: We're comfortable with the API here?

docs/docs/dev-server/plugins/storybook.md
```diff
@@ -98,13 +45,13 @@ Follow the regular storybook docs to learn how to create docs.
 
 ### Project types
 
-We currently supported `preact` and `web-components` project types. This corresponds to the Storybook "Framework".
+We currently support `preact` and `web-components` project types. This corresponds to the Storybook "Framework".
 
 Other project types could be supported, let us know if you are interested in this.
 
-### preview.js
+### main.js preview.js
```
Westbrook, line 52: Should we use a comma?

Westbrook: Or `and`?

Westbrook (commented):
Didn't realize the size of the scrollbar, will need to come back. Couple of small things.

docs/guides/dev-server/typescript-and-jsx.md
```diff
@@ -0,0 +1,65 @@
...
+  ],
+};
+```
+
+## Typescript
+
+The official typescript compiler (TSC) is the most predictable when it comes to compiling typescript, we generally recommend it for most users.
+
+Esbuild can be used to compile code on the fly before serving to the browser. The benefit is that you don't need to serve from generated files, and the transformation is faster because it doesn't do any type checking. To keep type checking during development, you could run keep TSC running in the background with the `--noEmit`.
+
```
daKmoR, line 50: optionally add a note on how to do that via an package.json script

docs/docs/dev-server/overview.md
LarsDenBakker: I think Web Dev Server for regular sentences, and `@web/dev-server` when refering to the package.

docs/docs/dev-server/plugins/storybook.md
LarsDenBakker: Yes, any concerns? :)

docs/docs/dev-server/plugins/storybook.md
LarsDenBakker: Added

docs/guides/dev-server/typescript-and-jsx.md
LarsDenBakker: I don't think it adds a lot more in this context, I don't know what other flags they're using so it's just `tsc --noEmit`

docs/docs/dev-server/overview.md
LarsDenBakker: I've updated all the references now

daKmoR (approved):

Issue #1039, opened 2020-12-01
Problems with symlinked dependencies in monorepos and phantom dependencies


javier-garcia-meteologica:
Since @open-wc/testing-karma is deprecated I'm migrating to @web/test-runner.

There are 2 big incompatibilities with monorepos.

### esbuild duplicates symlinked lit-html instances

This problem is related to https://github.com/Polymer/lit-element/issues/390. Lit-html module should only be imported once, otherwise it will render `[object Object]`. The module loader should cache the first 'lit-html' call and server always the same module.

Note that `preserveSymlinks: true` is necessary. Without it web-test-runner will complain that @open-wc/testing is outside of the project directory.

### web-test-runner referencees phantom dependencies

web-test-runner assumes that `node_modules` contains all explicit and implicit dependencies of the project. But this is incompatible with `pnpm`.

### Reproduction

The problems can be reproduced in this repo:

https://github.com/javier-garcia-meteologica/test-runner-lit-hml-monopero-symlink-issue

```
$ git clone https://github.com/javier-garcia-meteologica/test-runner-lit-hml-monopero-symlink-issue
$ cd test-runner-lit-hml-monopero-symlink-issue
$ rush install
$ rush build
$ cd lit-element-ts-esbuild
$ yarn run test
```

I modified the code to log `this.shadowRoot.innerHTML` to the console. The tests will fail because lit-html is imported more than once and it's rendering those pesky `[object Object]`.

```
 🚧 Browser logs:
      <!---->[object Object]<!---->
      <!---->[object Object]<!---->
      <!---->[object Object]<!---->
      <!---->[object Object]<!---->

 ❌ MyElement > increases the counter on button click
      at: test/my-element.test.ts:16:44
      TypeError: Cannot read property 'click' of null
        at n.<anonymous> (test/my-element.test.ts:16:44)
```

`rush` uses `pnpm` as a package manager by default, but this repository had to switch to `yarn` so that `web-test-runner` finds phantom dependencies.

To replicate the problem of phantom dependencies, edit the file `<monorepo_root>/rush.json`, uncomment `"pnpmVersion"` and comment out `"yarnVersion"`.

```
"pnpmVersion":"4.14.4",
//"npmVersion":"4.5.0",
//"yarnVersion":"1.9.4",
```

Now purge the monorepo and rebuild

```
$ rm <monorepo_root>/common/config/rush/yarn.lock
$ rush update --full --purge
$ rush build
```

[...]

PR #1083, opened 2020-12-12
feat(dev-server-core): handle modules resolved outside root dir

LarsDenBakker:
## What I did

This fixes a long-standing issue with es-dev-server, web dev server and web test runner where it can't serve modules outside of the root directory of the web server.

When we encounter a file resolved to a path outside of the root dir we rewrite it with the following pattern:
```
/__wds-outside-root__/${depth}/${originalPath}
``` 

When we see an import with this path server side, we look at the depth walk up the file system with the specified depth:

```js
const newRootDir = path.join(rootDir, '../'.repeat(depth));
```

This avoids dumping the whole file path in the browser URL.

Resolving modules becomes a more plug-and-play experience. Users won't need to expand the root directory anymore in monorepo projects (`--root-dir ../../`) and for demo pages the URL doesn't need to include the package name anymore (`/packages/package-foo/demo/`).

Fixes https://github.com/modernweb-dev/web/issues/1039

### DIFF
--- a/.changeset/four-seals-refuse.md
+++ b/.changeset/four-seals-refuse.md
@@ -1,0 +1,9 @@
+---
+'@web/dev-server': minor
+'@web/dev-server-core': minor
+'@web/dev-server-rollup': minor
+'@web/test-runner': minor
+'@web/test-runner-core': minor
+---
+
+handle modules resolved outside root dir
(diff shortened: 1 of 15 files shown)


LarsDenBakker:
Note: do not merge yet, I want to combine it with some other breaking changes.

PR #1108, opened 2020-12-21
fix(dev-server-esbuild): handle calling without parameters

DrSensor:
## Motivation
Kinda surprised when I call `hmrPlugin()` and found this error
```
@web/dev-server-esbuild/dist/esbuildPluginFactory.js:7
    const target = (_a = args.target) !== null && _a !== void 0 ? _a : 'auto';
                              ^

TypeError: Cannot read property 'target' of undefined
```
Although calling `hmrPlugin({})` works fine.

## What I did

1. add `{}` as default args since all keys are optional


LarsDenBakker:
Thanks!

PR #1158, opened 2020-12-31
feat(rollup-plugin-html): inject service worker if defined by other plugin

daKmoR:
## What I did

1. check if another plugin created a service worker and define it as `serviceWorkerDestination` if so we inject a service worker registration code

I'm not sure if I'm currently using the best hooks for it... 🤔

If overall this sounds good I will make sure that `rollup-plugin-workbox` sets the custom shared option `serviceWorkerDestination`

what do you think?

### DIFF
--- a/packages/rollup-plugin-html/src/rollupPluginHTML.ts
+++ b/packages/rollup-plugin-html/src/rollupPluginHTML.ts
@@ -29,6 +29,7 @@ export function rollupPluginHTML(pluginOptions: RollupPluginHTMLOptions = {}): R
   let generatedBundles: GeneratedBundle[] = [];
   let externalTransformHtmlFns: TransformHtmlFunction[] = [];
   let defaultInjectDisabled = false;
+  let serviceWorkerDestination = '';
 
   function reset() {
     inputs = [];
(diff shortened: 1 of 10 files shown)


packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -846,4 +846,56 @@ describe('rollup-plugin-html', () => {
       '<html><head></head><body><img src="./image-c.png"><link rel="stylesheet" href="./styles.css"><img src="./image-b.svg"></body></html>',
     );
   });
+
+  it.only('will inject service worker registration code if a service worker path is defined by another plugin', async () => {
```
thepassle, line 850: it.only ^

packages/rollup-plugin-html/src/rollupPluginHTML.ts
```diff
@@ -110,6 +115,8 @@ export function rollupPluginHTML(pluginOptions: RollupPluginHTMLOptions = {}): R
      * @param {OutputBundle} bundle
      */
     async generateBundle(options, bundle) {
+      // @ts-ignore
+      // console.log({ serviceWorkerDestination });
```
LarsDenBakker, line 119: console.log

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -846,4 +846,56 @@ describe('rollup-plugin-html', () => {
...
+    function createServiceWorker() {
+      return {
+        name: 'create-service-worker',
+        // @ts-ignore
+        async resolveId(source, importer) {
```
LarsDenBakker, line 855: Can you explain what's going on in this function? Shouldn't you check the actual import being resolved is the service worker's? It seems like this would match all entrypoints?

LarsDenBakker:
Closing this since we took another approach.

PR #1159, opened 2021-01-01
feat(rollup-plugin-html): option to inject a service worker registration

daKmoR:
## What I did

1. inject a service worker registration script if enabled and a swDest filepath is provided

### DIFF
--- a/packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
+++ b/packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
@@ -28,6 +28,10 @@ export interface RollupPluginHTMLOptions {
   absoluteBaseUrl?: string;
   /** Whether to set full absolute urls for ['meta[property=og:image]', 'link[rel=canonical]', 'meta[property=og:url]'] or not. Requires a absoluteBaseUrl to be set. Default to true. */
   absoluteSocialMediaUrls?: boolean;
+  /** Should a service worker registration script be injected. Defaults to false. */
+  injectServiceWorker?: boolean;
+  /** File system path to the generated service worker file */
+  serviceWorkerPath?: string;
 }
 
 export interface GeneratedBundle {
--- a/packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
+++ b/packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
@@ -1,0 +1,48 @@
+import path from 'path';
+import { Document } from 'parse5';
+import {
+  findElement,
+  getTagName,
+  appendChild,
+  createScript,
+  setAttribute,
+} from '@web/parse5-utils';
+
+export interface injectServiceWorkerRegistrationArgs {
+  document: Document;
+  serviceWorkerPath: string;
+  outputDir: string;
+  htmlFileName: string;
+}
+
+export function injectServiceWorkerRegistration(args: injectServiceWorkerRegistrationArgs) {
+  const { document, serviceWorkerPath, outputDir, htmlFileName } = args;
+  const body = findElement(document, e => getTagName(e) === 'body');
+  if (!body) {
+    throw new Error('Missing body in HTML document.');
+  }
+
+  let serviceWorkerUrl = path.relative(outputDir, serviceWorkerPath);
+  serviceWorkerUrl = path.relative(path.dirname(htmlFileName), serviceWorkerUrl);
+  serviceWorkerUrl = serviceWorkerUrl.split(path.sep).join('/');
+
+  const code = `
+    if ('serviceWorker' in navigator) {
+      window.addEventListener('load', function() {
+        navigator.serviceWorker
+          .register('${serviceWorkerUrl}')
+          .then(function() {
+            console.log('ServiceWorker registered from "${serviceWorkerUrl}".');
+          })
+          .catch(function(err) {
+            console.log('ServiceWorker registration failed: ', err);
+          });
+      });
+    }
+  `;
+
+  const script = createScript({}, code);
+  setAttribute(script, 'inject-service-worker', '');
+
+  appendChild(body, script);
+}
--- a/packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
+++ b/packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
@@ -846,4 +846,42 @@ describe('rollup-plugin-html', () => {
       '<html><head></head><body><img src="./image-c.png"><link rel="stylesheet" href="./styles.css"><img src="./image-b.svg"></body></html>',
     );
   });
+
+  it('can inject a service worker registration script if injectServiceWorker and serviceWorkerPath are provided', async () => {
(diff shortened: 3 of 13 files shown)


packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
```diff
@@ -28,6 +28,10 @@ export interface RollupPluginHTMLOptions {
   absoluteBaseUrl?: string;
   /** Whether to set full absolute urls for ['meta[property=og:image]', 'link[rel=canonical]', 'meta[property=og:url]'] or not. Requires a absoluteBaseUrl to be set. Default to true. */
   absoluteSocialMediaUrls?: boolean;
+  /** Should a service worker registration script be injected */
+  injectServiceWorker?: boolean;
+  /** File system path to the generated service worker file */
+  swDest?: string;
```
LarsDenBakker, line 34:
```suggestion
  servicWorkerPath?: string;
```

Seems like a clearer name

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -846,4 +846,42 @@ describe('rollup-plugin-html', () => {
       '<html><head></head><body><img src="./image-c.png"><link rel="stylesheet" href="./styles.css"><img src="./image-b.svg"></body></html>',
     );
   });
+
+  it.only('can inject a service worker registration script if injectServiceWorker and swDest are provided', async () => {
```
LarsDenBakker, line 850:
```suggestion
  it('can inject a service worker registration script if injectServiceWorker and swDest are provided', async () => {
```

daKmoR: still testing windows 🙈

packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
daKmoR: that's from workbox... it only gets passed on so it's probably good to use the same name here and in the workbox plugin? 🤔

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function injectServiceWorkerRegistration(args: injectServiceWorkerRegistrationArgs) {
+  const { document, swDest, outputDir, htmlFileName } = args;
+  const body = findElement(document, e => getTagName(e) === 'body');
+  if (!body) {
+    throw new Error('Missing body in HTML document.');
+  }
+
+  let swPath = swDest.replace(`${outputDir}${path.sep}`, '');
+  swPath = path.relative(path.dirname(htmlFileName), swPath);
```
LarsDenBakker, line 26: This doesn't take into account windows file paths. We also need to take into account the `publicPath` option here. 

The logic for generating the right path is already in the `injectedUpdatedAssetPaths` function. So perhaps we can pull that to a separate util?

packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
LarsDenBakker: True.. through in the workbox plugin I guess there is a bit more context. I guess changing it now in the workbox plugin would be too much

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
LarsDenBakker: no problem just marking so we don't forget :D

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
daKmoR: hmmm windows test now succeed?

I tried to apply the same logic but there is no relativeOutputPath (e.g. that is what we want to get... but that logic needs it as an input? 🤔)

packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
daKmoR: rename to `serviceWorkerPath`

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
daKmoR: removed and all green

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
LarsDenBakker: What the user puts in the option is a file system path, right? so on windows it could be something like `C:\\foo\bar` which needs to be transformed to a browser path

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function injectServiceWorkerRegistration(args: injectServiceWorkerRegistrationArgs) {
+  const { document, serviceWorkerPath, outputDir, htmlFileName } = args;
+  const body = findElement(document, e => getTagName(e) === 'body');
+  if (!body) {
+    throw new Error('Missing body in HTML document.');
+  }
+
+  let serviceWorkerUrl = serviceWorkerPath.replace(`${outputDir}${path.sep}`, '');
```
LarsDenBakker, line 25:
```suggestion
  let serviceWorkerUrl = path.relative(outputDir, serviceWorkerUrl);
```

This would fail when `outputDir` has a trailing slash. I *think* this has the same but leaves handling edge cases to the path library.

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
```diff
@@ -0,0 +1,47 @@
...
+  serviceWorkerUrl = path.relative(path.dirname(htmlFileName), serviceWorkerUrl);
+
```
LarsDenBakker, line 27:
```suggestion
serviceWorkerUrl = serviceWorkerUrl.split(path.sep).join('/);
```

for windows

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
daKmoR: indeed - applied 👍 
thx 🙇‍♂️

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
LarsDenBakker: see below

packages/rollup-plugin-html/src/output/injectServiceWorkerRegistration.ts
daKmoR: applied and coved in test 👍

LarsDenBakker (approved):
Looks good to me!

PR #1161, opened 2021-01-01
fix(dev-server-esbuild): pass `esbuildconfig.define` to esbuild transform function

DrSensor:
Just found out that `define` option is not being used by dev-server-esbuild despite it's declared in the `EsBuildPluginArgs` and `EsbuildConfig` interface

## What I did

1. pass `esbuildconfig.define` to esbuild transform function


LarsDenBakker:
Thanks!

PR #1167, opened 2021-01-03
Fix commonjs include paths

abdonrd:
## What I did

1. Fix commonjs include paths

### DIFF
--- a/docs/guides/dev-server/using-plugins.md
+++ b/docs/guides/dev-server/using-plugins.md
@@ -93,8 +93,8 @@ export default {
     commonjs({
       include: [
         // the commonjs plugin is slow, list the required packages explicitly:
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Are you certain this is necessary? It should work with `**/*`?

abdonrd:
That syntax didn't work for me in a project with `apollo-client`. I had to leave it as:

```
      commonjs({
        include: [
          '**/node_modules/fast-json-stable-stringify/**/*',
          '**/node_modules/graphql-tag/**/*',
          '**/node_modules/zen-observable/**/*'
        ]
      }),
```

LarsDenBakker:
hmm ok, thanks for the fix!

PR #1200, opened 2021-01-13
feat(test-runner): add coverageConfig.reporters configuration option

MathieuPuech:
## What I did

1. add a reporters option for coverageConfig allowing to use [alternative istanbul reporters](https://istanbul.js.org/docs/advanced/alternative-reporters/) for code coverage

### DIFF
--- a/.changeset/rude-countries-pretend.md
+++ b/.changeset/rude-countries-pretend.md
@@ -1,0 +1,6 @@
+---
+'@web/test-runner': minor
+'@web/test-runner-core': minor
+---
+
+add configuration option reporters in coverageConfig to use various istanbul reporters
--- a/.changeset/rude-countries-pretend.md
+++ b/.changeset/rude-countries-pretend.md
@@ -1,6 +1,6 @@
 ---
-'@web/test-runner': minor
-'@web/test-runner-core': minor
+'@web/test-runner': patch
+'@web/test-runner-core': patch
 ---
 
 add configuration option reporters in coverageConfig to use various istanbul reporters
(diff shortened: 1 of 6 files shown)


.changeset/rude-countries-pretend.md
```diff
@@ -0,0 +1,6 @@
+---
+'@web/test-runner': minor
+'@web/test-runner-core': minor
```
LarsDenBakker, line 3: this should be a patch since we're still > 0.x.y

LarsDenBakker:
Looks good, thanks! Is it possible to add reporters other than the ones built-in from istanbul?

MathieuPuech:
I don’t find any documentation about that. I’ve found this custom reporter:
https://github.com/Elgolfin/istanbul-reporter-shield-badge#readme

It seems to be possible to register a reporter like that, I can add some code to enable this in the PR

MathieuPuech:
this package is a wrong example since it's based on the deprecated [istanbul](https://www.npmjs.com/package/istanbul) package.

A package using the good version could be coveradge but it seems to need some configuration when created:

https://github.com/brettz9/coveradge/blob/master/src/reporter.js

I will check about theses missing options

MathieuPuech:
it seems that istanbul reporters could need to options when using the `create` function.

See https://github.com/istanbuljs/nyc/blob/91ae8b8764cef361fe6add31ae0802dc221fb577/index.js#L457

I've added the options needed for `coveradge`

You can use `coveradge` installing in dev dependencies and add in the `reporters` array of `coverageConfig` the string `coveradge`.

The coveradge package ignore the context and generate the badge in the project root but it seems to be an issue on this package.

dwilson6:
Thanks for implementing this! I was just looking to see if passing custom reporters was possible.

PR #1202, opened 2021-01-14
chore: fix timeout docs for wtr

jorenbroekema:
Update docs to reflect the right defaults

See:
https://github.com/modernweb-dev/web/blob/master/packages/test-runner/src/config/parseConfig.ts#L35


daKmoR:
sweet - thx 🤗

PR #1232, opened 2021-01-26
fix(test-runner-webdriver) Wait until the page unloads before reusing frame

### LINKED ISSUE #1216: [test-runner] test-runner-saucelabs intermittently times out on Safari with html tests
We are seeing intermittent/flaky behavior running html tests on Safari using test-runner-saucelabs, leading to this error:
```
❌  Browser tests did not start after 120000ms You can increase this timeout with the testsStartTimeout option. Check the browser logs or open the browser in debug mode for more information.
```
After a bunch of debugging, I've found that on occasion it seems like the iframe fails to load or else errors before `@web/test-runner-mocha/dist/standalone.js` is able to open the WebSocket and send the `wtr-session-started` command (instrumenting the server, I never see those occur when it fails). Given the nature of Safari's WebDriver implementation, I can't figure out how to see what's actually going on inside the iframe (and we can't rely on wtr's console.log exfiltration due to there being no ws connection to send them out), so sort of stuck on finding the error.

### Reproduction steps
(unfortunately I don't have a reduced repro yet, but it is pretty easily reproducible using these steps on the `lit-next` branch of `lit-html`)
```
git clone git@github.com:polymer/lit-html -b lit-next
cd lit-html
npm install
npm run bootstrap
npm run build
cd packages/test
SAUCE_USERNAME=polymer-devs SAUCE_ACCESS_KEY=dddd22bf-ee83-48f0-832c-e182e53be09d BROWSERS='sauce:macOS 10.15/safari@13' npm run test:dev -- --watch ../{lit-html,lit-element,reactive-element}/development/test/**/*_test.html
```
That will filter the tests to run just the `.html` tests (~15 files; the `.js` tests don't seem to experience the timeout). If they run to completion (15/15) with no errors, keep re-running by pressing Enter until one times out; the failure case will typically hang at `(14/15)` followed by the "testsStartTimeout" error above; it will generally be a random one of the 15 files that times out, no one in particular.

For an example `.html` test file to see how we're setting up manual testing, see https://github.com/Polymer/lit-html/blob/lit-next/packages/reactive-element/src/test/polyfill-support/reactive-element_test.html. Note that in order to test different polyfill configurations, we opt the page out of `@web/dev-server-legacy`'s `polyfills` injection and include them on the page ourselves.

Although the repro fails on the head of the `lit-next` branch, it still fails after trying these:
* Upgrading versions of the `@web/test-runner(-*)` packages
* Changing the test setup boilerplate to just do `runTests(async () => await import('...test file...'))` rather than the more manual setup shown
* Running on Safari 14 rather than 13

Original issue at https://github.com/Polymer/lit-html/issues/1550.

kevinpschaaf:
## What I did

1. Use `executeAsync` when setting the iframe src to an empty page, and wait until the load (or error) event fires before returning the frame to `inactiveFrames`

This appears to resolve the intermittent `testsStartTimeout` timeouts we've seen on Safari when running on Sauce, reported in #1216.

Fixes #1216

### DIFF
--- a/.changeset/mighty-clocks-relax.md
+++ b/.changeset/mighty-clocks-relax.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-webdriver': patch
+---
+
+Fix intermittent testsStartTimeout on Safari on Sauce
--- a/packages/test-runner-webdriver/src/IFrameManager.ts
+++ b/packages/test-runner-webdriver/src/IFrameManager.ts
@@ -115,24 +115,26 @@ export class IFrameManager {
   async stopSession(id: string) {
     const frameId = this.getFrameId(id);
 
-    // Retrieve test results from iframe. Note: IIFE is used to prevent
-    // WebdriverIO from crashing failure with Puppeteer (default mode):
-    // Error: Evaluation failed: SyntaxError: Illegal return statement
-    // See https://github.com/webdriverio/webdriverio/pull/4829
-    const returnValue = await this.driver.execute(`
-      return (function() {
-        var iframe = document.getElementById("${frameId}");
-        var testCoverage;
-        try {
-          testCoverage = iframe.contentWindow.__coverage__;
-        } catch (error) {
-          // iframe can throw a cross-origin error if the test navigated
-        }
-
-        // set src after retrieving values to avoid the iframe from navigating away
-        iframe.src = "data:,";
-        return { testCoverage: testCoverage };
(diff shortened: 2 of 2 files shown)


LarsDenBakker:
Ah, excellent. I dug into this over the weekend, and tried a similar solution by returning a promise but that didn't work for me. Looks like `executeAsync` does the trick 👍

Issue #1234, opened 2021-01-27
"0 failed" should not be red


hmans:
This is mostly about ergonomics. It'd be a simple change, but I wanted to check in on how the maintainers feel about this before submitting a PR.

Currently, when there are zero test failures, test-runner will report it like this:

![image](https://user-images.githubusercontent.com/1061/105971935-5c49e780-608b-11eb-85fd-6ccf92a4ea06.png)

Red is a signal color that indicates failure. When there are 0 failures, two things happen:

- The text being red will startle users like myself who understand "red" as a color signaling failure (seriously, it can't be just me?)
- I now have to check if the number is "0". Yes, I know that if there had been actual failures, the output would look very different, but they're called signal colors for a reason. :)~

This is why I suggest to change the test output code so it only uses the red color when there have actually been failures. Without failures, the "0 failures" string should simply use the default color.

The change would be made [here](https://github.com/modernweb-dev/web/blob/d07c4c20ce58ceca43e39db0867843ceaa883f25/packages/test-runner/src/reporter/getTestProgress.ts#L39-L41), and I'm happy to submit a PR if this change is wanted.

LarsDenBakker:
Makes sense, if you can make the change that would be great :)

hmans:
Alright, I'm on it!

LarsDenBakker:
Added in https://github.com/modernweb-dev/web/pull/1235

PR #1235, opened 2021-01-27
Don't use red text color when there are 0 failures

hmans:
As discussed in #1234. Some notes:

- The contributing documentation didn't include any specific notes on how to run tests for the repository, but I think I got them to work via `yarn build && yarn test`. At least one test complained about Selenium not being able to start up; maybe it would be nice to have some documentation on how to get things up and running (or provide an install script that installs required system dependencies.)
- It didn't look like there was a specific test for this functionality, and considering the simplicity of the change, I did not bother adding one.

### DIFF
--- a/.changeset/heavy-rabbits-sin.md
+++ b/.changeset/heavy-rabbits-sin.md
@@ -1,0 +1,5 @@
+---
+"@web/test-runner": patch
+---
+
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
Thanks!

PR #1258, opened 2021-02-02
feat(dev-server-core): share websocket with iframe parent

LarsDenBakker:
## What I did

This shares the websocket instance with the iframe parent. This improves testing stability in webdriver where we use iframes to isolate test suites, especially on IE11 where the total number of concurrent connections is limited.

CC @kevinpschaaf

### DIFF
--- a/.changeset/poor-llamas-sin.md
+++ b/.changeset/poor-llamas-sin.md
@@ -1,0 +1,10 @@
+---
+'@web/test-runner': patch
+'@web/test-runner-core': patch
+'@web/dev-server-core': patch
+'@web/test-runner-browserstack': patch
+'@web/test-runner-saucelabs': patch
+'@web/test-runner-webdriver': patch
+---
+
+feat(dev-server-core): share websocket instances with iframe parent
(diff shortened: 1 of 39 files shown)


kevinpschaaf:
Cool, thanks for running with this!

The one thing I wasn't sure about this approach was whether there are plugins that rely on a 1:1 mapping between sockets and frames (since an e.g. `webSocket.send('hi')` sent from a server plugin would be received by the parent frame plus all iframes after this change, since all of them add a `message` listener to the same WebSocket). It seems like a plugin could pretty easily add addressing to messages to work around the fact that some frames may share a socket, but if that wasn't a requirement then before then the wouldn't have been written that way.

It looks like the watcher & hmr plugins would probably be fine since they broadcast messages to all receivers (?). Not sure if there are any that e.g. send messages to specific sockets based on information sent from the frame.

LarsDenBakker:
Yeah right now it's fine, and I don't foresee any functionalities where we need this at the moment. If it becomes a problem we can add a test runner specific socket connection.. or perhaps IE11 will have died out by then :)

PR #1284, opened 2021-02-09
fix(parse5-utils): ignore comments in isHtmlFragment

SBylemans:
## What I did

1. Replaced comments passed to `isHtmlFragment` with empty string.
```
function isHtmlFragment(html) {
  let htmlWithoutComments = html.replace(/<!--.*?-->/g, '');
  return !REGEXP_IS_HTML_DOCUMENT.test(htmlWithoutComments);
}
```
2. Add test which checks `isHtmlFragment('<!-- COMMENT --><!DOCTYPE><my-element></my-element>')` is false.

### DIFF
--- a/.changeset/nine-pears-invent.md
+++ b/.changeset/nine-pears-invent.md
@@ -1,0 +1,5 @@
+---
+'@web/parse5-utils': minor
+---
+
+Ignore comments when checking isHtmlFragment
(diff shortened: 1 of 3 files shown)


abdonrd:
We update the package, but this doesn't work for us:

https://github.com/IBM/pwa-lit-template/blob/main/index.html#L1-L8

PR #1288, opened 2021-02-11
fix(parse5-utils): allow break lines when checking isHtmlFragment

abdonrd:
## What I did

1. Allow break lines when checking isHtmlFragment

Reference: https://github.com/modernweb-dev/web/pull/1284#issuecomment-777316386


abdonrd:
@SBylemans @LarsDenBakker because https://github.com/modernweb-dev/web/pull/1284#issuecomment-777316386

PR #1291, opened 2021-02-11
fix(rollup-plugin-polyfills-loader): fix relative paths to polyfills with flattenOutput: false

### LINKED ISSUE #1290: [rollup-plugin-polyfills-loader] When flattenOutput:false, path to polyfills can break
`@web/rollup-plugin-polyfills-loader` doesn't work well with the `flattenOutput: false` option on `@web/rollup-plugin-html`  because it looks like the loader always creates polyfill scripts relative to the current page (whereas polyfills are generated to a fixed dir).

This line in the code assumes the `./polyfills` will always reach the `${rootDir}/polyfills` folder, which will fail with `html` files in folders non-flattened folders off the `rootDir`: the https://github.com/modernweb-dev/web/blob/5cb31bb2268fddc2482f12d7efa1c8c3fb635893/packages/polyfills-loader/src/createPolyfillsLoader.ts#L150

kevinpschaaf:
## What I did

1. Added optional `relativePathToPolyfills` config option to `@web/polyfills-loader`
2. Changed `@web/rollup-plugin-polyfills-loader` to set `relativePathToPolyfills` based on the relative path between `polyfillsDir` and the input html file.

Fixes #1290

### DIFF
--- a/.changeset/mighty-pandas-brake.md
+++ b/.changeset/mighty-pandas-brake.md
@@ -1,0 +1,6 @@
+---
+'@web/polyfills-loader': patch
+'@web/rollup-plugin-polyfills-loader': patch
+---
+
+fix(rollup-plugin-polyfills-loader): fix relative paths to polyfills with `flattenOutput: false`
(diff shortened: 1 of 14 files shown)


LarsDenBakker:
Thanks for this! There is a test failure on windows, probably due to windows paths.

LarsDenBakker:
Thanks!

PR #1292, opened 2021-02-12
attempt to fix reading configName when there is no CLI args

TomzBench:
## What I did

1. When calling `startDevServer()` programatically with readCliArgsFlag as false, the `startDevServer()` routine attempts to read the cli args anyway.  This patch will instead attempt to guess the directory from the configName option.

### DIFF
--- a/packages/config-loader/src/index.js
+++ b/packages/config-loader/src/index.js
@@ -11,6 +11,10 @@ const EXTENSIONS = ['.mjs', '.cjs', '.js'];
  * @param {string} [basedir]
  */
 async function readConfig(name, customPath, basedir = process.cwd()) {
+  if (fileExists(name)) {
+    return importOrRequireConfig(name);
+  }
+
   const resolvedCustomPath = customPath ? path.resolve(basedir, customPath) : undefined;
   if (resolvedCustomPath && !(await fileExists(resolvedCustomPath))) {
     throw new ConfigLoaderError(`Could not find a config file at ${resolvedCustomPath}`);
--- a/packages/dev-server/src/startDevServer.ts
+++ b/packages/dev-server/src/startDevServer.ts
@@ -7,7 +7,6 @@ import { readFileConfig } from './config/readFileConfig';
 import { DevServerStartError } from './DevServerStartError';
 import { createLogger } from './logger/createLogger';
 import { openBrowser } from './openBrowser';
-import { dirname } from 'path';
 
 export interface StartDevServerParams {
   /**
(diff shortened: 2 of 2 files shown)


TomzBench:
Sorry, this doesn't fix issue at all.

However, if I pass the fully qualified path to my web-dev-server.config.mjs file, the readConfig seems to manipulate it and guess and ends up resolving the wrong place

TomzBench:
I removed the first commit and the second commit corrects the issue for me.  

For recap the problem is, i am passing the fully resolved path of my web-dev-server.config.mjs file to `startDevServer()` under the configName configuration property. I intuit that this configuration parameter asking for my config should accept the full path of my config. However, the config parsing infrastructure seems to require a base dir and attempts to resolve an already resolved path. But this commit will first see if the file name exists with out any guessing work, and if it does, accept that as the config.

importOrRequireConfig requires 2 parameters (the base dir as the second parameter). However, it works with out this parameter and the error is silently ignored. which is a very sane thing to do here since the file is resolved already and you don't need to parse the package.json file since you have the extension already.

LarsDenBakker:
The option is intended for passing the config file name, not the fully resolved path. It's always resolved relative to the current working directory.

There is an option for setting the config path, but it's only possible to pass it as a CLI arg currently. We can add a separate option for that, but you could also just pass the config as an object which is also supported.

TomzBench:
@LarsDenBakker I tried passing the config directly but I had trouble importing my config because it is mjs file.  So I wanted to take advantage of the importing scheme used by this package and just pass the file path.

Since my path is fully resolved already I would have to basically make my path relative, or use the argv option. Intuitively though I assumed the resolved path should work. So it is confusing part of the API.

```javascript
 async function readConfig(name, customPath, basedir = process.cwd()) {
  if (fileExists(name)) {
    return importOrRequireConfig(name);
  }
  // ...
  ```
  
This change here seems to make the `configName` work with the path already resolved and makes this API much more versatile. I don't think there are any negative consequences - all current usages of this routine would fail the fileExists check and continue as before. But adding this check allows the configName option behave as expected in that I think it is sane to expect a fully resolve path to work. Though that is my opinion about the API.

LarsDenBakker:
A name is not a path, they're different things. We can add a check and throw if you're passing an absolute path or something with a file extension if that makes it better :)

In the same function you adjusted there is a `customPath` parameter, that's what intended to be used. It's just not exposed right now.

TomzBench:
@LarsDenBakker 

Yes the name is not a path and that should have clued me in better.  But no worries! I'm using the argv[] to shovel in my config so this works for me.

PR #1329, opened 2021-03-02
Fix/rollup csp

jorenbroekema:
## What I did

1. Fix multi-output docs error
2. Add missing prepend util export
3. Add strictCSPInlineScripts option

Works in my app now, but if other plugins insert scripts (e.g. polyfills-loader), these inline scripts are not caught by rollup-plugin-html where I am scanning for inline scripts and injecting the meta tag, because I am doing it too early.

The above is fixed, should add some more logic though to scan for existing meta CSP tags, and instead of overriding or adding a new meta CSP tag, reuse the existing one and add to it.

Above is fixed too now, ready for review

### DIFF
--- a/.changeset/large-glasses-attack.md
+++ b/.changeset/large-glasses-attack.md
@@ -1,0 +1,5 @@
+---
+'@web/parse5-utils': patch
+---
+
+Add missing export for prepend utility.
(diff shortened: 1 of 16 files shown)


LarsDenBakker:
Looks good :)

PR #1339, opened 2021-03-06
[RFC] feat(test-runner-visual-regression): make "update" strict and add "buildCache" for more repeatable tests

Westbrook:
## What I did

1. Make the `update` property in the VRT config "strict". It will no _only_ update if set to true, allowing newly added regressions (tests without a baseline) to fail the test pass.
2. Add the `buildCache` property to the VRT config. This has the test save all of the current test output in one place so that it is easier to create a new cache from updated tests listings, including when tests have been intentionally removed.

## Questions
1. Using `update` strictly means that a user with new screenshots can't add those shots while testing if existing tests fail. Is this a common use case? Should we add an `update: 'auto'` option? Or possibly a `strict` flag so that the current functionality can be accessed without breaks, but users can alter their workflow to the new version if desired?
2. `buildCache` currently leverages the "failed" workflow. In that way should we update the naming conventions throughout to the `saveCurrent` or similar to clarify the usage of the code throughout?
3. Placing "current" next to "failed" may make it more difficult for a user to parse the output of their tests. Should these two outputs be treated differently?


LarsDenBakker:
Looks good to me!

LarsDenBakker (approved):

PR #1369, opened 2021-03-28
Send keys

### LINKED ISSUE #1289: [RFC] [Test Runner] "sendKeys" command for Playwright
A `sendKeys` plugin for `@web/test-runner` _could_ be so simple as:
```
import type { Page } from 'playwright';

export function sendKeysPlugin() {
    return {
        name: 'send-keys-command',
        async executeCommand({
            command,
            payload = {},
            session,
        }: {
            payload: { type?: string; press?: string };
            command: string;
            session: {
                id: string;
                browser: { type: string; getPage: (id: string) => Page };
            };
        }) {
            if (command === 'send-keys') {
                // handle specific behavior for playwright
                if (session.browser.type === 'playwright') {
                    if (payload.type) {
                        const page = session.browser.getPage(session.id);
                        await page.keyboard.type(payload.type);
                        return true;
                    } else if (payload.press) {
                        const page = session.browser.getPage(session.id);
                        await page.keyboard.press(payload.press);
                        return true;
                    }
                }
                // you might not be able to support all browser launchers
                throw new Error(
                    `Sending keys is not supported for browser type ${session.browser.type}.`
                );
            }
            return false;
        },
    };
}
```

Here a single approximates the [`Keyboard` class](https://playwright.dev/docs/api/class-keyboard) and suggests the ability to more fully cover it the remaining `down`, `insertText`, and `up` interfaces by expanding the `payload` to optionally accept properties describing the values to send those methods. A "complete" rendering of this interface would also allow for an `options` bag as accepted by the `press` and `type` methods.

Initial questions regarding that are:
1. If we cover all of the `Keyboard` interface, should we rename the plugin `keyboardPlugin` to make the relationship between it and the Playwright API more clear?
2. Does the `payload` relationship to the interface on `Keyboard` make sense, or if there are better path to supporting the various capabilities there?
3. Is there context where individual `keyboardPressPlugin`, `keyboardInsertTextPlugin`, etc. would be more productive than this full `Keyboard` class implementation?
4. Is Playwright coverage alone enough? Or is it better to have Puppeteer in the same plugin? Should they each have a purpose built plugin of their own?

**Puppeteer**
The Puppeteer API also surfaces a [`Keyboard` class](https://pptr.dev/#?product=Puppeteer&version=v5.1.0&show=api-class-keyboard) with a _similar_ API:
- keyboard.down(key[, options]) // matched with addition of options bag
- keyboard.press(key[, options]) // matched
- keyboard.sendCharacter(char) // send `keypress` and `input` events that I'm not even sure what the real use parallel to which would be and is one more event (`keypress`) than the `insertText` API for Playwright.
- keyboard.type(text[, options]) // matched
- keyboard.up(key) // matched 

## Code format
We structure a lot of our tools in CJS + esm, would we want to persist that here or has the community (Node@12) gotten to a place where we could go esm only on this? 

## Code location
After we get closer on the structure of the plugin, it'll be useful to spend a little thought on where the code should live. If this is the _only_ plugin we ever vend, it makes sense to live in this repo. If we end up vending a LOT of plugin, then it's possible that these should live elsewhere, which may be a problem for another time? 

**dir**
Possible directories for the plugin:

`packages/test-runner-plugin-keyboard`
`packages/test-runner-plugin-keyboard-playwright`
`packages/test-runner-plugin-keyboard-puppeteer`
`packages/something-else`

In the case we want to separate the plugins in somewhat but don't want another repo, would could expand into an additional workspace beyond `packages`, e.g.:

`plugins/test-runner-plugin-keyboard`
`plugins/test-runner-plugin-keyboard-playwright`
`plugins/test-runner-plugin-keyboard-puppeteer`
`plugins/something-else`

**Package name**
How will we want to vend this (and other) plugin(s)?

`@web/test-runner-plugin-keyboard`
`@web/test-runner-plugin-keyboard-playwright`
`@web/test-runner-plugin-keyboard-puppeteer`

Does it matter much that the name is quite long?

Westbrook:
## What I did

1. Adds the `sendKeysPlugin` to the commands package for use in WTR sessions.
  - simplifies this first iteration to use only the `press` and `type` commands which are currently available in both Playwright and Puppeteer
  - future work (or possibly clean up to make this not a Draft) should expand the coverage to more of the keyboard API and the options bags that these command accept
2. Adds generic typing to `TestRunnerPlugin` so that commands can define their payload.
  - `isViewportObject` is used to runtime test the shape of the `setViewportPlugin` payload, but I'm not sure it makes sense to do such a complicated test on the `sendKeysPlugin` payload at this time.
3. Adds documentation to the "Docs" section
  - should additional "Guides" information be added?

fixes #1289


packages/test-runner-commands/test/send-keys/sendKeysPlugin.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  // playwright doesn't work on windows VM right now
+  if (platform() !== 'win32') {
+    it('can set the viewport on playwright', async () => {
+      await runTests({
+        files: [path.join(__dirname, 'browser-test.js')],
+        browsers: [
+          playwrightLauncher({ product: 'chromium' }),
+          playwrightLauncher({ product: 'firefox' }),
+          // TODO: make webkit work in the CI
+          // playwrightLauncher({ product: 'webkit' }),
```
LarsDenBakker, line 29: Is this TODO copied from another test? If so can we check if works now?

LarsDenBakker:
Looks good to me!

packages/test-runner-commands/test/send-keys/sendKeysPlugin.test.ts
Westbrook: It is indeed. I'll switch it on and see what happens...

Westbrook: That seems to have run...should I switch that on in other tests?

packages/test-runner-commands/src/sendKeysPlugin.ts
```diff
@@ -0,0 +1,41 @@
...
+            await page.keyboard.type(payload.type);
+            return true;
+          } else if (payload.press) {
+            await page.keyboard.press(payload.press);
+            return true;
+          }
+        }
+
+        // you might not be able to support all browser launchers
+        throw new Error(`Sending keys is not supported for browser type ${session.browser.type}.`);
```
justinfagnani, line 37: What happens for Web Driver tests? It'd be great to be able to do this on Sauce/BrowserStack too.

packages/test-runner-commands/src/sendKeysPlugin.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export function sendKeysPlugin(): TestRunnerPlugin<SendKeysPayload> {
+  return {
+    name: 'send-keys-command',
+    async executeCommand({ command, payload = {}, session }): Promise<any> {
+      if (command === 'send-keys') {
+        // handle specific behavior for playwright
+        if (session.browser.type === 'playwright') {
+          const page = (session.browser as PlaywrightLauncher).getPage(session.id);
+          if (payload.type) {
```
justinfagnani, line 15: warn on unknown options or including both `type` and `press`?

packages/test-runner-commands/browser/commands.mjs
```diff
@@ -48,3 +48,7 @@ export function emulateMedia(media) {
 export function setUserAgent(options) {
   return executeServerCommand('set-user-agent', options);
 }
+
+export function sendKeys(options) {
```
justinfagnani, line 52: Is there a way to get a type for options here so our IDEs will tell us how to use the API? Including a link to the `type` vs `press` distinction in the Puppeteer API?

packages/test-runner-commands/test/send-keys/sendKeysPlugin.test.ts
LarsDenBakker: Yeah!

packages/test-runner-commands/src/sendKeysPlugin.ts
LarsDenBakker: I think we can support webdriver with this as well

packages/test-runner-commands/browser/commands.mjs
LarsDenBakker: There is a handwritten type definition file for this, to avoid having to run a second build for the browser code.

https://github.com/modernweb-dev/web/blob/master/packages/test-runner-commands/browser/commands.d.ts

packages/test-runner-commands/src/sendKeysPlugin.ts
Westbrook: Not sure whether Web Driver has a similar API here or not, but I can check. Last I heard at least one of Sauce/BrowserStack was adding support for Playwright/Puppeteer, but I've not been keeping a close eye on that either. 

There are also a handful of places where there can/should be more features/options/API that I figured we'd do feature releases on if this worked enough to get learning on the scope and flexibility people would want form this sort of feature.

packages/test-runner-commands/src/sendKeysPlugin.ts
Westbrook: Makes sense. Would you rather thing in the warnings or in the type that you asked for in `commands.mjs`? If we went `{keys:string} | {press:string}` for a type, we could manage this is TS primarily?

packages/test-runner-commands/browser/commands.mjs
Westbrook: I'll update this. Hopefully it's easy enough to share the types between this and the actual plugin so that they don't have live in multiple places.

Westbrook: @justinfagnani you wouldn't happen to have an example of a comment with links in it that I could replicate here? Not sure if you just want the raw URLs or if something else would be more useful here.

packages/test-runner-commands/src/sendKeysPlugin.ts
Westbrook: For reference: https://webdriver.io/docs/api/browser/keys It's got a little more wiring up needed, but I'll dive a little deeper on this.

LarsDenBakker: We can make this a separate PR as well if it's a lot of effort. And like you said there are more places where we need to add more support.

packages/test-runner-commands/browser/commands.mjs
Westbrook: Added expansive comments to the the `*.d.ts` file, but I'm not sure if that means it'll 100% pass into local files or not...thoughts?

packages/test-runner-commands/src/sendKeysPlugin.ts
Westbrook: Added.

packages/test-runner-commands/src/sendKeysPlugin.ts
Westbrook: I'm not sure I can make time to follow the rabbit on this right now. Would love to get this in without if we all OK with that for now.

PR #1382, opened 2021-04-08
Do not resolve multiple times outside root files

edoardocavazza:
Right now, files outside the root dir are not correctly served when using a rollup plugin in combination with the node resolve option. Multiple `resolveImport` passes result in `/__wds-outside-root__/<n>/__wds-outside-root__/<n>/<module>/<file>`.

## What I did

1. Added a check for already resolved files outside the `rootDir`


LarsDenBakker:
Thanks!

PR #1387, opened 2021-04-08
Use unescaped import specifier

edoardocavazza:
Since version 0.4.0, `es-module-lexer` exposes a `n` property for each import specifier with its unescaped source. This is useful (and more accurate) for virtual modules like `\0commonHelpers`.

## What I did

1. try to use the `n` property of the specifier, fallback to the previous version if missing.

### DIFF
--- a/packages/dev-server-core/src/plugins/transformModuleImportsPlugin.ts
+++ b/packages/dev-server-core/src/plugins/transformModuleImportsPlugin.ts
@@ -25,6 +25,7 @@ interface ParsedImport {
   ss: number;
   se: number;
   d: number;
+  n?: string;
 }
 
 const CONCAT_NO_PACKAGE_ERROR =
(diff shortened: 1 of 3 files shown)


edoardocavazza:
Hello @LarsDenBakker, I am sorry for the mention :( Could you please review this PR? It fixes a bug with the import of the commonjs helper. Thank you!

LarsDenBakker:
This is interesting.. but in the `@web/dev-server-rollup` we already handle these null character. Could you elaborate which use case this fixes? Also can JS handle the unescaped import characters?

edoardocavazza:
The problem occurs when using both esbuild and commonjs plugins. I think that esbuild output "reverts" the null character transformation 🤔

LarsDenBakker:
I'm not sure that's the case, we transform the import to a whole different import. Do you see that this change fixes a specific problem? What I saw in the past was that the null character would be sent to the browser but not sent back in the request URL.. making it useless.

edoardocavazza:
I'll try to setup a unit test for reproduction.

edoardocavazza:
Hello @LarsDenBakker, I added an unit test with the use case. If you revert the proposed fix, you can see that null character is not handled after esbuild transformation.

LarsDenBakker:
Thanks, should be good to go then.

PR #1388, opened 2021-04-09
Detect websocket url from server

edoardocavazza:
WebSocket should be created using the server url instead of the runtime location (script may be imported from a differente origin).

## What I did

1. Stored the server url using the `serverStart` hook
2. Replaced the url argument of the `WebSocket` constructor


LarsDenBakker:
Thanks!

PR #1391, opened 2021-04-14
feat(rollup-plugin-html): retain attributes of script tags

daKmoR:
## What I did

1) Preserve attributes on html script tags.

Input:

```html
<script type="module" src="..." some="attribute">
```

Output before:

```html
<script type="module" src="...">
```

Output now:

```html
<script type="module" src="..." some="attribute">
```

2) Update `@web/polyfills-loader` to preserve attributes if given
3) Update `@web/rollup-plugin-polyfills-loader` to preserve attributes if given by `rollup-plugin-html`


packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
```diff
@@ -44,6 +44,17 @@ export interface GeneratedBundle {
   bundle: OutputBundle;
 }
 
+export interface Attribute {
```
bennypowers, line 47: maybe we can use https://developer.mozilla.org/en-US/docs/Web/API/Attr here?

bennypowers (approved):
In principle I think a list of attr objects makes sense.

daKmoR:
I'm almost there 🤞

but there are 4 failing tests I can not figure out and the error is really not helpful 😭

```
  1) rollup-plugin-html
       can build with inline modules:
     Error: Invalid regular expression flag
      at error (node_modules/rollup/dist/shared/rollup.js:5265:30)
      at Module.error (node_modules/rollup/dist/shared/rollup.js:9858:16)
      at tryParse (node_modules/rollup/dist/shared/rollup.js:9739:23)
      at Module.setSource (node_modules/rollup/dist/shared/rollup.js:10165:19)
      at ModuleLoader.addModuleSource (node_modules/rollup/dist/shared/rollup.js:18335:20)
      at ModuleLoader.fetchModule (node_modules/rollup/dist/shared/rollup.js:18391:9)
      at async Promise.all (index 0)
 ```
 
 I did not change any regex 🤔
 
does anyone has an idea?

File: packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts (hunk 1 above)
Comment (daKmoR) [reply packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts:47]: there is also `import { Attribute } from '@web/parse5-utils';` with

```js
/**
 * Element attribute.
 */
export interface Attribute {
    /**
     * The name of the attribute.
     */
    name: string;

    /**
     * The value of the attribute.
     */
    value: string;

    /**
     * The namespace of the attribute.
     */
    namespace?: string;

    /**
     * The namespace-related prefix of the attribute.
     */
    prefix?: string;
}
```

will use that for now

LarsDenBakker:
Looks good to me

PR #1419, opened 2021-04-26
Update Lit & Friends slack links

simi:
- previously called Polymer slack

PS: similar change to https://github.com/open-wc/open-wc/pull/2146


daKmoR:
thx 🤗

PR #1426, opened 2021-04-28
Deserialize bound function from browser logs

edoardocavazza:
Bound function were not correctly handled by the browser logs module.

## What I did

1. Convert `bound fn` to `fn` in order to avoid syntax errors

### DIFF
--- a/packages/browser-logs/src/deserialize.ts
+++ b/packages/browser-logs/src/deserialize.ts
@@ -41,7 +41,12 @@ function createReviver(promises: Promise<unknown>[], options?: DeserializeOption
           return;
         case 'Function':
           // Create a fake function with the same name. We don't log the function implementation.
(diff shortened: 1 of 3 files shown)


packages/browser-logs/src/deserialize.ts
```diff
@@ -41,7 +41,12 @@ function createReviver(promises: Promise<unknown>[], options?: DeserializeOption
           return;
         case 'Function':
           // Create a fake function with the same name. We don't log the function implementation.
-          return new Function(`return function ${value.name}() { /* implementation hidden */ }`)();
+          return new Function(
+            `return function ${value.name.replace(
+              /^bound\s+/,
```
LarsDenBakker, line 46: Can we make the regexp a constant? That avoids recreating on each call, since this code is run a lot it does make a difference.

edoardocavazza: I pushed the change

Issue #689, opened 2020-10-05
[test-runner] Cannot read property 'decl' of undefined in coverage UI


web-padawan:
Two files in the [PR for `vaadin-date-picker`](https://github.com/vaadin/vaadin-date-picker/pull/733) have this problem:

```
Cannot read property 'decl' of undefined
TypeError: Cannot read property 'decl' of undefined
    at /Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-reports/lib/html/annotator.js:89:31
    at Array.forEach ()
    at annotateFunctions (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-reports/lib/html/annotator.js:86:29)
    at annotateSourceCode (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-reports/lib/html/annotator.js:234:9)
    at HtmlReport.onDetail (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-reports/lib/html/index.js:409:33)
    at LcovReport. [as onDetail] (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-reports/lib/lcov/index.js:28:23)
    at Visitor.value (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-lib-report/lib/tree.js:38:38)
    at ReportNode.visit (/Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-lib-report/lib/tree.js:88:21)
    at /Users/sergey/vaadin/vaadin-date-picker-new/node_modules/istanbul-lib-report/lib/tree.js:92:19
    at Array.forEach ()
```

As you can see the numbers are there:

![Screenshot 2020-10-05 at 13 30 19](https://user-images.githubusercontent.com/10589913/95069169-fb2e8a80-070e-11eb-898d-68beb86a8c18.png)

But when opening the files, I get this:

![Screenshot 2020-10-05 at 13 30 26](https://user-images.githubusercontent.com/10589913/95069172-fc5fb780-070e-11eb-88f7-40d004b4aa56.png)

![Screenshot 2020-10-05 at 13 30 33](https://user-images.githubusercontent.com/10589913/95069180-fd90e480-070e-11eb-976d-81bdce50574e.png)

LarsDenBakker:
Seems related to https://github.com/istanbuljs/v8-to-istanbul/issues/121

[...]

PR #1436, opened 2021-05-04
fix(test-runner): fix error when function metadata varies between tests

Stuk:
Fixes https://github.com/modernweb-dev/web/issues/689

See also https://github.com/istanbuljs/v8-to-istanbul/issues/121

## What I did

1. This PR updates `addingMissingCoverageItems` to also add missing functions. This can occur when a function is defined in another function, and the outer function is not called in every test run. For example methods defined on a class using arrow functions `=>` are only created when the constructor runs. Specifically, I have moved the code inside the loops out into their own functions, so they can be called with the branch coverage data and the function coverage data.

### DIFF
--- a/packages/test-runner-core/src/coverage/getTestCoverage.ts
+++ b/packages/test-runner-core/src/coverage/getTestCoverage.ts
@@ -4,6 +4,7 @@ import {
   CoverageMap,
   CoverageMapData,
   BranchMapping,
+  FunctionMapping,
   Location,
   Range,
 } from 'istanbul-lib-coverage';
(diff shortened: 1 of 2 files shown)


Stuk:
Let me know what changeset I should add. Maybe a minor bump?

LarsDenBakker:
@Stuk Thanks a lot! We can do a patch, you can run `yarn changeset` or add one by clicking the PR comment. I can also do it for you.

@benjamind @Westbrook  could you check if this fixes it for you as well?

Westbrook:
Looking!

benjamind:
Confirmed this fixes the reproducible test case found [here](https://github.com/Stuk/wtr-coverage-repro)

Stuk:
@LarsDenBakker actually, I don't want to mess anything up, so would you ind adding the changeset?

Westbrook:
The results certainly look nice here!

Before:
image

After:
image

packages/test-runner-core/src/coverage/getTestCoverage.ts
```diff
@@ -26,61 +27,105 @@ export interface TestCoverage {
...
+
 /**
- * Cross references coverage mapping data, looking for missing code branches
- * and adding empty entries for them if found. This is necessary because istanbul
- * expects code branch data to be equal for all coverage entries, while v8 only
- * outputs actual covered code branches.
+ * Cross references coverage mapping data, looking for missing code branches and
+ * functions and adding empty entries for them if found. This is necessary
+ * because istanbul expects code branch and function data to be equal for all
+ * coverage entries. V8 only outputs actual covered code branchesm and functions
```
Westbrook, line 86:
```suggestion
 * coverage entries. V8 only outputs actual covered code branches and functions
```

Stuk:
Added changeset with some help from Westbrook. Thanks!

Westbrook (approved):
I think this is good to go! It's already a big win for my usage in Spectrum Web Components as outlined in the screenshots above and I know a number of other have bumped into this as well. Thanks for digging into this for us @Stuk 🎉

LarsDenBakker:
Thanks, much appreciated!

Issue #1443, opened 2021-05-10
[web-test-runner] plugin_node_resolve_1.nodeResolve on node 14 + .mjs config


btopro:
running node 14 and using `rollup.config.mjs` I was getting the following error: `plugin_node_resolve_1.nodeResolve`

In searching I was able to find that versions of https://www.npmjs.com/package/@rollup/plugin-node-resolve newer versions should resolve this config loading issue. In adding 13.0.0 to my package.json and running again it now correctly accepts the command. This seems an interrelated issues potentially as 

https://github.com/modernweb-dev/web/blob/master/packages/dev-server-storybook/package.json#L62
dev-server-storybook is pointed to 11.x.x.

Making this 13 (assuming that still works) should resolve this issue for those of us starting to use `.mjs` based configuration routines.

Westbrook:
Very interesting. Could we interest you in making a PR with this change to `dev-server-storybook`?

btopro:
to up this version? As long as it doesn't break anything else, absolutely. It was a very odd series of events so wasn't sure if this was really a bug or a series of unfortunate configurations.

PR #1444, opened 2021-05-10
#1443 up to 13 so web-test-runner happy in mjs configs

btopro:
> copied from the issue in question for the full explanation

running node 14 and using rollup.config.mjs I was getting the following error: plugin_node_resolve_1.nodeResolve

In searching I was able to find that versions of https://www.npmjs.com/package/@rollup/plugin-node-resolve newer versions should resolve this config loading issue. In adding 13.0.0 to my package.json and running again it now correctly accepts the command. This seems an interrelated issues potentially as

https://github.com/modernweb-dev/web/blob/master/packages/dev-server-storybook/package.json#L62
dev-server-storybook is pointed to 11.x.x.

Making this 13 (assuming that still works) should resolve this issue for those of us starting to use .mjs based configuration routines.

### DIFF
--- a/packages/dev-server-storybook/package.json
+++ b/packages/dev-server-storybook/package.json
@@ -59,7 +59,7 @@
     "@babel/preset-env": "^7.12.11",
     "@mdx-js/mdx": "^1.6.22",
     "@rollup/plugin-babel": "^5.2.2",
-    "@rollup/plugin-node-resolve": "^11.0.1",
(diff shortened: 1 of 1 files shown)


web-padawan:
There is a breaking change that caused build to fail. See https://github.com/modernweb-dev/web/issues/1575 for more details. I'm closing this PR for now.

PR #1460, opened 2021-05-19
chore: update lockfile

LarsDenBakker:
## What I did

Updates the lockfile and fixes any issues that cropped up. Most important one is an issue with the latest playwright and retrieving test coverage.


packages/test-runner-playwright/src/PlaywrightLauncherPage.ts
```diff
@@ -71,7 +79,13 @@ export class PlaywrightLauncherPage {
     }
 
     // get native coverage from playwright
-    const coverage = ((await this.playwrightPage.coverage?.stopJSCoverage()) ?? []) as V8Coverage[];
+    let coverage: V8Coverage[];
+    if (this.product === 'chromium') {
+      coverage = await this.playwrightPage?.coverage?.stopJSCoverage();
```
Westbrook, line 84: Does the need for this change imply that Playwright is moving towards coverage in other browsers? Or it is just not protecting the call anymore?

Might be interesting to support multi-browser coverage (if they are moving that direction) in the future, but definitely not worth getting into at this point.

Westbrook (approved):
LGTM.

One comment for conversation, but ready to move this forward!

packages/test-runner-playwright/src/PlaywrightLauncherPage.ts
LarsDenBakker: Those functions exist now on all browsers, but they throw on non-chromium. So we can't rely on feature detection anymore, and have to hardcode known browsers.

PR #1464, opened 2021-05-20
fix(test-runner-junit-reporter): strip invalid XML characters in stdout

### LINKED ISSUE #1463: [test-runner-junit-reporter] Logging ANSI colors to stdout produces malformed XML reports
JUnit XML reports get corrupted when non UTF-8 characters (like ANSI escape characters) get logged to STDOUT.

Lines from stdout are currently being wrapped in CDATA without further escaping:
https://github.com/modernweb-dev/web/blob/10a188b9ed90879cbb16729ef5b8abdf06ae8be3/packages/test-runner-junit-reporter/src/junitReporter.ts#L100-L101

Any invalid characters should be stripped from the CDATA using…
https://github.com/modernweb-dev/web/blob/10a188b9ed90879cbb16729ef5b8abdf06ae8be3/packages/test-runner-junit-reporter/src/junitReporter.ts#L120-L128

Repro:
1. add the following log line to a test case…
    ```js
    console.log('Hello \u001b[41mWorld\u001b[0m')
    ```
2. run `wtr --reporter=junit`
3. open generated `test-report.xml` in a browser, which should display an error like:
   ```
   error on line 14 at column 26: Input is not proper UTF-8, indicate encoding !
   Bytes: 0x1B 0x5B 0x34 0x31
   ```

muhqu:
Fixes #1463

### DIFF
--- a/.changeset/two-mangos-allow.md
+++ b/.changeset/two-mangos-allow.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-junit-reporter': patch
+---
+
+fix logging non UTF-8 characters (e.g. ANSI escape seq) produces malformed junit xml report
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
Thanks!

PR #1465, opened 2021-05-21
Add support for holding and releasing keys to the sendKeys command

sissbruecker:
Currently it is not possible to send a key combination using the sendKeys command in all test runners that the plugin supports. The original PR states in the [changeset](https://github.com/modernweb-dev/web/commit/ce90c7c36dc53b03ba98fe47f83eadf427009137#diff-ae1d797ecb266266783436e2ef62e5ed325be942cf34b476e82c72490b651bdaR10) that combinations such as `Shift-Tab` are supported by the `press` command, which is not true from my testing (also there are no tests to verify this). My understanding is that Playwright supports it, but Puppeteer does not.

To solve this, this change introduces two additonal actions / properties to the sendKeys command: `down` and `up`. These allow to hold down a key, and then release a key. This effectively allows pressing key combinations or trigger other interactions that require holding down a key, such as text selection. 

Both new commands are supported by Playwright and Puppeteer and use the same API.

### DIFF
--- a/packages/test-runner-commands/src/sendKeysPlugin.ts
+++ b/packages/test-runner-commands/src/sendKeysPlugin.ts
@@ -2,29 +2,57 @@ import { TestRunnerPlugin } from '@web/test-runner-core';
 import type { ChromeLauncher, puppeteerCore } from '@web/test-runner-chrome';
 import type { PlaywrightLauncher } from '@web/test-runner-playwright';
 
+type TypePayload = { type: string; };
+type PressPayload = { press: string; };
+type DownPayload = { down: string; };
+type UpPayload = { up: string; };
+
 export type SendKeysPayload =
-  | { type: string; press?: undefined }
-  | { type?: undefined; press: string };
+  | TypePayload
+  | PressPayload
+  | DownPayload
+  | UpPayload;
 
 function isObject(payload: unknown): payload is Record<string, unknown> {
   return payload != null && typeof payload === 'object';
 }
 
 function isSendKeysPayload(payload: unknown): boolean {
+  const validOptions = ['type', 'press', 'down', 'up'];
+
   if (!isObject(payload)) throw new Error('You must provide a `SendKeysPayload` object');
-  if (!!payload.type && !!payload.press)
+
+  const numberOfValidOptions = Object.keys(payload).filter(key => validOptions.includes(key)).length;
+  const unknownOptions = Object.keys(payload).filter(key => !validOptions.includes(key));
+
+  if (numberOfValidOptions > 1)
     throw new Error(
-      'You must provide ONLY one of the `type` or `press` properties for pass to the browser runner.',
+      `You must provide ONLY one of the following properties to pass to the browser runner: ${validOptions.join(', ')}.`,
     );
-  if (!payload.type && !payload.press)
+  if (numberOfValidOptions === 0)
     throw new Error(
-      'You mist provide one of the `type` or `press` properties for pass to the browser runner.',
+      `You must provide one of the following properties to pass to the browser runner: ${validOptions.join(', ')}.`,
     );
-  const payloadKeys = Object.keys(payload).filter(key => key !== 'press' && key !== 'type');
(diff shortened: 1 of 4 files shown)


packages/test-runner-commands/src/sendKeysPlugin.ts
```diff
@@ -2,31 +2,59 @@ import { TestRunnerPlugin } from '@web/test-runner-core';
...
+  | PressPayload
+  | DownPayload
+  | UpPayload;
 
 function isObject(payload: unknown): payload is Record<string, unknown> {
   return payload != null && typeof payload === 'object';
 }
 
 function isSendKeysPayload(payload: unknown): boolean {
+  const validOptions = ['type', 'press', 'down', 'up'];
```
sissbruecker, line 21: Had to change the validation logic quite a bit here. Unfortunately there are no existing tests for it, so I did not add tests either. I can add them if necessary.

packages/test-runner-commands/src/sendKeysPlugin.ts
```diff
@@ -2,31 +2,59 @@ import { TestRunnerPlugin } from '@web/test-runner-core';
 import type { ChromeLauncher, puppeteerCore } from '@web/test-runner-chrome';
 import type { PlaywrightLauncher } from '@web/test-runner-playwright';
 
+type TypePayload = { type: string; };
+type PressPayload = { press: string; };
+type DownPayload = { down: string; };
+type UpPayload = { up: string; };
```
sissbruecker, line 8: Decided to keep the API simple here instead of adding something like a `modifier` prop. While it is a bit more verbose for the user to send multiple sendKeys commands to trigger a key combination, it matches the APIs that Playwright and Puppeteer expose, which allows to defer to their documentation for the most part. Plus it keeps the logic in this plugin slim.

LarsDenBakker:
Looks good to me. @Westbrook what do you think?

Westbrook (approved):
This looks like a good first step.

I really feel Puppeteer is making a mistake not supporting this API and  I wonder if we should close over it in this (or more likely follow up work) when leveraged via the `sendKeys` API so that it looks the same in both contexts... 🤔 

If we don't close over it, we may alternatively think about this sort of `steps` approach so we don't force users to make so many round trips to the browser runner: https://github.com/modernweb-dev/web/discussions/1397#discussioncomment-653803 That may also be fast enough not to worry, but I've never bench marked it in any form or fashion.

Unless one of the ideas lights a fire under you, I've no reason for them to be blockers on this change. Thanks so much for covering over the x-runner capabilities here!

sissbruecker:
Thanks for the quick feedback 👍

> via the sendKeys API so that it looks the same in both contexts... 🤔

Just to make sure this is not misunderstood: Both runners support the "lowest common denominator" which is first sending a `down` and then sending a `press` - the API here is exactly the same for both, down to the strings used for the keys to press (as far as I can see). That's why I documented it as the recommended approach in the docs, instead of going into details about the differences between test runners, and that Playwright actually has a simpler API available.

It would also be possible to implement support for key kombinations for Puppeteers `press` in WTR, but that would be a lot more effort (detecting modifier key strings, holding modifiers first, adding error handling to ensure keys are always released). OTOH adding support for the "low-level" APIs like is done here enables you to do the same with a bit more effort and also gives more flexibility for other use-cases that we did not anticipate.

> If we don't close over it, we may alternatively think about this sort of steps approach so we don't force users to make so many round trips to the browser runner: #1397 (comment) That may also be fast enough not to worry, but I've never bench marked it in any form or fashion.

> Unless one of the ideas lights a fire under you, I've no reason for them to be blockers on this change. Thanks so much for covering over the x-runner capabilities here!

OK, I see the issue here. Each command goes from browser -> Node -> test runner, etc. The steps idea is interesting, but I feel that could easily be added as a separate enhancement. I might look into that at some point if there is interest.

Westbrook:
I think we’re 100% in the same place on this. Thanks so much for the great work here!

PR #1488, opened 2021-06-02
feat(rollup-plugin-html): support picture tags with `<source srset="...">`

daKmoR:
## What I did

1. Support `picture` tags by handling `source` tags with `srcset` attributes.

Example of supported html

```html
<picture>
  <source type="image/avif" srcset="small.avif 30w, big.avif 60w" sizes="30px" />
  <source type="image/jpeg" srcset="small.jpeg 30w, big.jpeg 60w" sizes="30px" />
  <img alt="..." src="small.jpeg" width="30" height="15" loading="lazy" decoding="async" />
</picture>
```

PS: there will be a follow-up to support `src` and `srcset` on both `img` and `source`


packages/rollup-plugin-html/src/assets/utils.ts
```diff
@@ -7,12 +7,35 @@ import { serialize } from 'v8';
 const hashedLinkRels = ['stylesheet'];
 const linkRels = [...hashedLinkRels, 'icon', 'manifest', 'apple-touch-icon', 'mask-icon'];
 
+function getSrsSetUrls(srcset: string) {
```
LarsDenBakker, line 10:
```suggestion
function getSrcSetUrls(srcset: string) {
```

packages/rollup-plugin-html/src/assets/utils.ts
```diff
@@ -92,13 +120,20 @@ export function getSourceAttribute(node: Node) {
...
+
   const src = getAttribute(node, key);
   if (typeof key !== 'string' || src === '') {
     throw createError(`Missing attribute ${key} in element ${serialize(node)}`);
   }
-  return src as string;
+
+  let paths: string[] = [];
+  if (src) {
+    paths = key !== 'srcset' ? [src] : getSrsSetUrls(src);
```
LarsDenBakker, line 133:
```suggestion
    paths = key !== 'srcset' ? [src] : getSrcSetUrls(src);
```

Issue #1480, opened 2021-05-27
Issues resolving Storybook imports when building for prod


sgammon:
Hey there modernweb-devs,

We're having issues building using the `@web/dev-server-storybook` package. It works in development just fine, but can't resolve relative imports at build time when we run `build-storybook`:

**`package.json`** (full dependencies available [here](https://gist.github.com/sgammon/b4115c2a22f053dca61e1791296568ce)):
```json
{
  "scripts": {
    "catalog": "wds --config tools/devserver.config.mjs",
    "storybook:build": "build-storybook --config-dir=storybook/.storybook --output-dir=storybook/dist",
  },
  "devDependencies": {
    "...": "lots of stuff here",
    "@web/dev-server": "^0.1.17",
    "@web/dev-server-esbuild": "^0.2.12",
    "@web/dev-server-storybook": "^0.3.5",
    "@web/storybook-prebuilt": "^0.1.20",
    "typescript": "3.9.7"
  }
}
```

**`storybook/.storybook/main.js`**:
```js
const css = require('rollup-plugin-import-css');

module.exports = {
  stories: ['../stories/**/*.stories.@(js|jsx|ts|tsx)'],
  rollupConfig(config) {
    // allow CSS imports
    config.plugins.push(css());
    return config;
  }
};
```

**`storybook/stories/Button.stories.ts`**:
```
import '../../elements/standard/button/button';

// Storybook code here
```

**`elements/standard/button/button.ts`**:
```
// valid button code here
```

`@web/dev-server` has no trouble finding this import when we run `wds` via the package scripts above. However, when we run `build-storybook`, also via the package scripts, it results in:

```
(node:20818) UnhandledPromiseRejectionWarning: Error: Could not resolve '../../elements/standard/button/button' from storybook/stories/Button.stories.ts
    at error (/Volumes/VANTAGE/lit-bazel/node_modules/@web/dev-server-storybook/node_modules/rollup/dist/shared/rollup.js:7426:30)
    at ModuleLoader.handleResolveId (/Volumes/VANTAGE/lit-bazel/node_modules/@web/dev-server-storybook/node_modules/rollup/dist/shared/rollup.js:19141:24)
    at /Volumes/VANTAGE/lit-bazel/node_modules/@web/dev-server-storybook/node_modules/rollup/dist/shared/rollup.js:19097:22
    at async Promise.all (index 0)
[...]

PR #1490, opened 2021-06-03
fix(dev-server-esbuild): null checks in transform function, allowing …

Marshal27:
…the use of this plugin for .storybook/main.js with @rollup/plugin-typescript

```javascript
config.plugins.unshift(
      esbuildPlugin({ 
        ts: true, 
        json: true,
        target: 'auto' }),
      typescript()
    );
```

## What I did
null checked context.response and context.path
```typescript
if (context.response && context.response.is('html')) {
    // we are transforming inline scripts
    loader = 'js';
}
else if(context.path) {
    const fileExtension = path_1.default.posix.extname(context.path);
    loader = this.esbuildConfig.loaders[fileExtension];
  }
```

### DIFF
--- a/packages/dev-server-esbuild/src/EsbuildPlugin.ts
+++ b/packages/dev-server-esbuild/src/EsbuildPlugin.ts
@@ -97,10 +97,10 @@ export class EsbuildPlugin implements Plugin {
   async transform(context: Context) {
     let loader: Loader;
 
-    if (context.response.is('html')) {
+    if (context && context.response.is('html')) {
       // we are transforming inline scripts
       loader = 'js';
-    } else {
+    } else if (context.path) {
       const fileExtension = path.posix.extname(context.path);
       loader = this.esbuildConfig.loaders[fileExtension];
     }


LarsDenBakker:
When adding those guards, what's the point of adding the plugin. Does it do any work?

This is not intended as a rollup plugin. There is quite some specific logic for the dev server.I think there is already an esbuild plugin for rollup.

Marshal27:
The typescript plugin, unfortunately, does not work yet for the dev server pipeline; the official recommendation is to use the esbuildPlugin for wds.

This creates a significant gap for import resolution, etc, for the local version of storybook and the static build for production.

https://github.com/modernweb-dev/web/issues/1480

This simple modification allowed me to standardize the import resolution/bundling between both instances of storybook local dev and production. I understand you line of thinking, wondering, "what is the plugin actually doing in the main.js pipeline, if anything"... But I would argue that, it is solving a problem.

I also understand there is an official rollup esbuild plugin, but that in my opinion, is just one more variable to manage. If I can drop the same esbuild plugin in both main.js and wds and call it a day... why not.

Having these guards to allow this to be leveraged in the main.js pipeline and quickly standardize the imports between inner loop and production build, sof far, has been working well for me.

Marshal27:
I apologize for the confusion on this item... this was a result of my lack of understanding regarding the differences between the `@web/dev-server` build pipeline static storybook main.js rollup pipeline... this was a complete config issue on my part and the details provided in this issue are not applicable.

PR #1492, opened 2021-06-04
feat(rollup-plugin-import-meta-assets): allow transform to skip processing asset

TimvdLippe:
This allows users who pre-process a subset of assets separately to
skip processing the assets via this plugin.

As part of https://crrev.com/c/2939994 I am working on optimizing
our SVG assets dynamically. However, we also have other image
formats in our source tree, including `.png` and `.avif`. These images
are pre-optimized and we aren't sure yet whether we want to
dynamically optimize this during build time (as that tends to be
computationally expensive work).

However, when trying to integrate the plugin, I discovered that our
build system (GN + Ninja) started complaining about files being
out-of-date. I then discovered that the `.png` and `.avif` assets
were double copied: once by us in GN and once in the plugin.

I then attempted to include only the SVG files using `include` and was
confused as to why that wouldn't work. While reading the source
code of this plugin, I realized that the `include` pattern only works
for the actual JavaScript files and not the assets.

To allow users to conditionally define transformations for a subset
of assets, I would propose to update the plugin to ignore `null`
as transformation result. Alternatively, you can introduce separate
filters called `assetInclude` and `assetExclude`, but I think this
solution is more maintainable, as it provides users full control
without adding a lot more plugin options.

When I update our transform to the snippet below, everything
would be working as expected with GN + Rollup:

```js
importMetaAssets({
  async transform(assetBuffer, assetPath) {
    if (assetPath.endsWith('.svg')) {
      const {data} = await optimize(assetBuffer.toString());
      return data;
    }
    return null;
  }
}),
```

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/test/integration.test.js
+++ b/packages/rollup-plugin-import-meta-assets/test/integration.test.js
@@ -87,6 +87,34 @@ describe('rollup-plugin-import-meta-assets', () => {
     ]);
   });
 
+  it('simple bundle with ignored assets', async () => {
+    const config = {
+      input: { 'transform-bundle-ignored': require.resolve('./fixtures/transform-entrypoint.js') },
+      plugins: [
+        importMetaAssets({
+          transform: async (assetBuffer, assetPath) => {
+            // Only minify SVG files
+            return assetPath.endsWith('.svg')
+              ? // Fake minification with an XML comment
+                assetBuffer.toString() + '<!-- minified -->\n'
+              : null;
+          },
+        }),
+      ],
+    };
+
+    const bundle = await rollup.rollup(config);
+    const { output } = await bundle.generate(outputConfig);
+
+    expect(output.length).to.equal(5);
+    expectChunk(output, 'snapshots/transform-bundle-ignored.js', 'transform-bundle-ignored.js', [
+      expectAsset(output, 'snapshots/one.min.svg', 'one.svg', 'assets/one-d81655b9.svg'),
+      expectAsset(output, 'snapshots/two.min.svg', 'two.svg', 'assets/two-00516e7a.svg'),
+      expectAsset(output, 'snapshots/three.min.svg', 'three.svg', 'assets/three-0ba6692d.svg'),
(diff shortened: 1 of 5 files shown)


packages/rollup-plugin-import-meta-assets/test/integration.test.js
```diff
@@ -87,6 +87,34 @@ describe('rollup-plugin-import-meta-assets', () => {
...
+    const config = {
+      input: { 'transform-bundle-ignored': require.resolve('./fixtures/transform-entrypoint.js') },
+      plugins: [
+        importMetaAssets({
+          transform: async (assetBuffer, assetPath) => {
+            // Only minify SVG files
+            return assetPath.endsWith('.svg')
+              ? // Fake minification with an XML comment
+                assetBuffer.toString() + '<!-- minified -->\n'
+              : null;
```
daKmoR, line 100: what would be the benefit of supporting `null`? is that significantly faster than returning the buffer? 🤔

e.g. how does it compare with
```js
return assetPath.endsWith('.svg') ? assetBuffer.toString() + '<!-- minified -->\n' : assetBuffer;
```

daKmoR: ah I now understand your use case - makes sense 👍

TimvdLippe: Sorry hadn't had the chance to reply yet. This would mostly be useful for denoting you don't want the buffer to be processed at all. E.g. if you are gradually migrating to this plugin, or if you still have some images left that are pre-processed differently. For example, compressing with AVIF can take quite a bit of time, depending on your compression settings. Therefore it doesn't always make sense to compress every single image during build time and instead do that as a pre-processing step.

TimvdLippe:
@daKmoR Is there anything more I can do here or feedback to process? I would like to integrate this plugin in the DevTools build system and remove some techdebt from our image integration 😄

daKmoR:
looks good and the CI agrees as well 🤗

thank you 🙇‍♂️

PR #1504, opened 2021-06-09
fix: allow stripXMLInvalidChars when replace it not available

Westbrook:
## What I did

1. gated the replace behind the presence of replace

I know there has to be something underlying this issue, but I can't seem to find it... this change safely allows the use of replace without seeming to break anything else, so I'd like your thoughts.


Westbrook:
Now with linting and changeset!

PR #1527, opened 2021-06-20
Fix coverage branch fusion

### LINKED ISSUE #1468: Uncovered branches when files are loaded multiple times
I've added on the lit-element example-project the reproduction of the issue
https://github.com/MathieuPuech/example-projects/tree/bug-coverage-fusion/lit-element

The issue is here:
image

This branch is covered by the tests in my-other-element.test.js but uncovered because of my-element.test.js

It seems that there is some fusion of coverage done in WTR removing coverage of covered branches.

jvdsande:
## What I did

1. Add a `findEncompassingKey` function to `getTestCoverage.ts`
2. Use this instead of `findKey` in `patchCoverageItems`

Those two changes allow `patchCoverageItems` to fallback to an encompassing branch coverage status when adding an unreported branch to a session report.

The reasoning for this change is described here: https://github.com/modernweb-dev/web/issues/1468#issuecomment-864518266.

This Pull Request fixes #1468.

## Additional notes

I came up with this solution through trial & error, analysing the output of `coverage` maps received from the various sessions. I'm not sure whether this fix is proper or only a patchwork on a patchwork.

I tested the solution on a large codebase and only saw positive consequences, but I cannot guarantee that this does not bring unintended side-effects. I'd really appreciate if someone knowledgeable about the `getTestCoverage.ts` file and overall V8-to-Istanbul conversion could take a look at what I did!

### DIFF
--- a/.changeset/shy-frogs-hug.md
+++ b/.changeset/shy-frogs-hug.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-core': patch
+---
+
+fix(test-runner): fix error when merging multiple test-suites falsely reporting covered branches as uncovered
--- a/.changeset/ten-pens-tease.md
+++ b/.changeset/ten-pens-tease.md
@@ -1,0 +1,5 @@
+---
+"@web/test-runner": patch
+---
+
+Fix coverage branch fusion
--- a/packages/test-runner-core/src/coverage/getTestCoverage.ts
+++ b/packages/test-runner-core/src/coverage/getTestCoverage.ts
@@ -25,19 +25,28 @@ export interface TestCoverage {
 }
 
 const locEquals = (a: Location, b: Location) => a.column === b.column && a.line === b.line;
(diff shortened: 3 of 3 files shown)


LarsDenBakker:
Thanks a lot for digging into is. It seems to be correct as far as I can determine. So let's merge it, and see how it works out.

jvdsande:
Ah, I forgot to run prettier before committing it seems. Should be OK now!

jvdsande:
Added a commit to pass the `findOriginalKey` function to the `patchCoverageItems` instead of directly using `findEncompassingKey` inside.

Use `findEncompassingKey` for branches, but keep using `findKey` for _functions_, as nested functions don't seem to be impacted by this behaviour. I think it's cleaner that way and should avoid potential edge cases.

LarsDenBakker:
Fyi @Westbrook @web-padawan

PR #1529, opened 2021-06-21
feat: handle preload and template assets

bennypowers:
## What I did

1. add assets in `<link rel="preload">` tags
2. traverse the AST for assets hidden in `<template>` tags


bennypowers:
Tests fail here, I think I maybe am making a boolean error somewhere, so PTAL

PR #1533, opened 2021-06-29
feat(rollup-plugin-html): exclude files

bennypowers:
## What I did

1. add `exclude` option to rollup-plugin-html

### Case
User wishes to exclude assets (for example, 11ty html file _includes_, `playground-elements/playground-service-worker-proxy.html`) from the build.

#### Example Usage

in`rocket.config.js`
```js
setupBuildPlugins: [
  adjustPluginOptions('html', {
    ignore: ['**/_assets/**/*.html', '**/_merged_assets/**/*.html'],
  }),
],
```

### DIFF
--- a/packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
+++ b/packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
@@ -421,6 +421,22 @@ describe('rollup-plugin-html', () => {
     expect(pageC).to.include('<script type="module" src="./shared.js"></script>');
   });
 
+  it('can exclude globs', async () => {
+    const config = {
+      plugins: [
+        rollupPluginHTML({
+          input: 'exclude/**/*.html',
+          exclude: '**/partial.html',
+          rootDir,
+        }),
+      ],
+    };
+
+    const bundle = await rollup(config);
+    const { output } = await bundle.generate(outputConfig);
+    expect(output.length).to.equal(2);
+  });
+
   it('creates unique inline script names', async () => {
     const config = {
       plugins: [
--- a/.changeset/little-fans-shake.md
+++ b/.changeset/little-fans-shake.md
@@ -1,0 +1,5 @@
+---
+'@web/rollup-plugin-html': minor
+---
+
+Add `exclude` option. Used e.g. to ignore html file assets when bundling.
--- a/docs/docs/building/rollup-plugin-html.md
+++ b/docs/docs/building/rollup-plugin-html.md
@@ -307,6 +307,8 @@ export interface InputHTMLOptions {
 export interface RollupPluginHTMLOptions {
   /** HTML file(s) to use as input. If not set, uses rollup input option. */
   input?: string | InputHTMLOptions | (string | InputHTMLOptions)[];
+  /** HTML file glob patterns or patterns to ignore */
+  exclude?: string | string[];
   /** Whether to minify the output HTML. */
   minify?: boolean;
   /** Whether to preserve or flatten the directory structure of the HTML file. */
--- a/packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
+++ b/packages/rollup-plugin-html/src/RollupPluginHTMLOptions.ts
@@ -13,6 +13,8 @@ export interface InputHTMLOptions {
 export interface RollupPluginHTMLOptions {
   /** HTML file(s) to use as input. If not set, uses rollup input option. */
   input?: string | InputHTMLOptions | (string | InputHTMLOptions)[];
+  /** HTML file glob pattern or patterns to ignore */
+  exclude?: string | string[];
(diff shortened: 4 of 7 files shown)


bennypowers:
patch for anyone in urgent need:

`patches/@web+rollup-plugin-html+1.8.0.patch`
```patch
diff --git a/node_modules/@web/rollup-plugin-html/dist/input/getInputData.js b/node_modules/@web/rollup-plugin-html/dist/input/getInputData.js
index 5580f6e..a55de83 100644
--- a/node_modules/@web/rollup-plugin-html/dist/input/getInputData.js
+++ b/node_modules/@web/rollup-plugin-html/dist/input/getInputData.js
@@ -10,8 +10,8 @@ const glob_1 = __importDefault(require("glob"));
 const utils_1 = require("../utils");
 const normalizeInputOptions_1 = require("./normalizeInputOptions");
 const extractModulesAndAssets_1 = require("./extract/extractModulesAndAssets");
-function resolveGlob(fromGlob, rootDir) {
-    const files = glob_1.default.sync(fromGlob, { cwd: rootDir, absolute: true });
+function resolveGlob(fromGlob, opts) {
+    const files = glob_1.default.sync(fromGlob, { ...opts, absolute: true });
     return (files
         // filter out directories
         .filter(filePath => !fs_1.default.lstatSync(filePath).isDirectory()));
@@ -43,7 +43,7 @@ function createInputData(params) {
 }
 function getInputData(pluginOptions, rollupInput) {
     var _a, _b;
-    const { rootDir = process.cwd(), flattenOutput, extractAssets = true, absolutePathPrefix, } = pluginOptions;
+    const { rootDir = process.cwd(), flattenOutput, extractAssets = true, absolutePathPrefix, ignore } = pluginOptions;
     const allInputs = normalizeInputOptions_1.normalizeInputOptions(pluginOptions, rollupInput);
     const result = [];
     for (const input of allInputs) {
@@ -59,7 +59,7 @@ function getInputData(pluginOptions, rollupInput) {
             result.push(data);
         }
         else if (typeof input.path === 'string') {
-            const filePaths = resolveGlob(input.path, rootDir);
+            const filePaths = resolveGlob(input.path, { cwd: rootDir, ignore });
             if (filePaths.length === 0) {
                 throw new Error(`Could not find any HTML files for pattern: ${input.path}, resolved relative to ${rootDir}`);
             }
```

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -421,6 +421,22 @@ describe('rollup-plugin-html', () => {
...
+          input: 'exclude/**/*.html',
+          exclude: '**/partial.html',
+          rootDir,
+        }),
+      ],
+    };
+
+    const bundle = await rollup(config);
+    const { output } = await bundle.generate(outputConfig);
+    expect(output.length).to.equal(2);
```
daKmoR, line 437: why is this not 1? 
only `packages/rollup-plugin-html/test/fixtures/rollup-plugin-html/exclude/index.html` should be considered right? 🤔

bennypowers:
```js
[
  {
    exports: [ 'default' ],
    facadeModuleId: '@web/rollup-plugin-html-noop',
    isDynamicEntry: false,
    isEntry: true,
    isImplicitEntry: false,
    modules: [Object: null prototype] {
      '@web/rollup-plugin-html-noop': [Object]
    },
    name: [Getter],
    type: 'chunk',
    code: 'var rollupPluginHtmlNoop = "noop";\n' +
      '\n' +
      'export default rollupPluginHtmlNoop;\n',
    dynamicImports: [],
    fileName: 'rollup-plugin-html-noop.js',
    implicitlyLoadedBefore: [],
    importedBindings: {},
    imports: [],
    map: null,
    referencedFiles: []
  },
  {
    fileName: 'index.html',
    name: 'index.html',
    isAsset: [Getter],
    source: '<html><head></head><body><a href="assets/docs/partial.html"></a>\n' +
      '</body></html>',
    type: 'asset'
  }
]
```

bennypowers: does that explanation satisfy, @daKmoR ? should I make it explicit in the test code?

peschee:
This would be an awesome addition to the plugin!

We're using `rollup-plugin-html` to bundle different microfrontends (in a microservice architecture) and reference some assets provided by another microfrontend at build-time. 

My use-case is a bit different from yours though @bennypowers since I need to exclude assets referenced in an `index.html` file, example:

```html
<link rel="icon" type="image/png" sizes="32x32" href="/microservice1/favicon-32x32.png" />
```

The plugin is looking for this path `/microservice1/favicon-32x32.png` as an asset, and I'd need that ignored.

PR #1537, opened 2021-07-01
fix(dev-server): fixes #1536, correctly handle outside-root paths

### LINKED ISSUE #1536: [dev-server] commonjs plugin does not work with monorepo
It appears when a file is served from outside the root the virtual file import that is designed to pass through to the rollup plugin does not get resolved properly in the plugin.

In the non-monorepo case we see a file requested of this form:
```
http://localhost:8001/__web-dev-server__/rollup/react.production.min.js?web-dev-server-rollup-null-byte=%00%2Fhome%2Fben%2Fcode%2Fexperiments%2Freact-wtr%2Fnode_modules%2Freact%2Fcjs%2Freact.production.min.js%3Fcommonjs-proxy
```

This is served by the commonjs plugin and returns this content:
```
import { __moduleExports } from "./../../node_modules/react/cjs/react.production.min.js"; export default __moduleExports;
```

Under the monorepo environment however we get a url of this form:
```
http://localhost:8000/__web-dev-server__/rollup/react.production.min.js?web-dev-server-rollup-null-byte=%00%2F__wds-outside-root__%2F2%2Fcommon%2Ftemp%2Fnode_modules%2F.pnpm%2Freact%4017.0.1%2Fnode_modules%2Freact%2Fcjs%2Freact.production.min.js%3Fcommonjs-proxy
```

This request remains in pending state forever. I believe its due to the file not being found by the rollup plugin and perhaps not having any good error handling for the plugin interface. If you modify the url in the non-monorepo case to make it point to an invalid filesystem path you get the same exact behavior.

I have found that if you resolve the outside-of-root path to the absolute filesystem path then pass this onto the rollup plugin it fixes things. Will file a PR to fix shortly.

benjamind:
## What I did

1. Fully resolve `resolvedImportPath` for files outside of the root before passing to plugins.

## Case

Using rollup plugins like `@rollup/plugin-commonjs` in a monorepo environment. In these cases we can be serving dependencies from outside of the 'root' (where we're executing the test runner or WDS). In these cases we were previously passing the `/__wds-outside-root__/` prefix to the plugins which was causing them to not resolve the files being processed properly. Doing full path resolution seems to fix it.

### DIFF
--- a/packages/dev-server-rollup/src/rollupAdapter.ts
+++ b/packages/dev-server-rollup/src/rollupAdapter.ts
@@ -28,6 +28,7 @@ import { createRollupPluginContexts, RollupPluginContexts } from './createRollup
 const NULL_BYTE_PARAM = 'web-dev-server-rollup-null-byte';
 const VIRTUAL_FILE_PREFIX = '/__web-dev-server__/rollup';
 const WDS_FILE_PREFIX = '/__web-dev-server__';
+const OUTSIDE_ROOT_REGEXP = /\/__wds-outside-root__\/([0-9]+)\/(.*)/;
 
 /**
  * Wraps rollup error in a custom error for web dev server.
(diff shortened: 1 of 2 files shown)


packages/dev-server-rollup/src/rollupAdapter.ts
```diff
@@ -192,6 +192,12 @@ export function rollupAdapter(
           const filename = path.basename(
             resolvedImportPath.replace(/\0*/g, '').split('?')[0].split('#')[0],
           );
+          // if the resolve import path is outside our normal root, fully resolve the file path for rollup
+          const matches = resolvedImportPath.match(/\/__wds-outside-root__\/([0-9]+)\/(.*)/);
```
LarsDenBakker, line 196: Can we make the regexp a constant? that avoids recreation the regexp each time, and this is quite a hot path

benjamind: Good catch, done in f02a638

Westbrook (approved):

PR #1543, opened 2021-07-07
feat: bump webdriverio to 7.7.4

web-padawan:
## What I did

1. Updated `webdriver` and `webdriverio` dependencies to latest versions.


bennypowers:
does this change any of our user-facing apis?

web-padawan:
I don't think this affects user facing APIs, but according to semver, a minor dependency bump should require a minor bump.

PR #1546, opened 2021-07-08
fix(test-runner): use block comments in snapshots to make them work in all browsers

manolakis:
## What I did

1. I changed the way markers are done in snapshots because Firefox and Webkit have some problems with line comments, and using block comments seems to work properly.


manolakis:
@LarsDenBakker

PR #1560, opened 2021-07-23
fix(test-runner): escape snapshots to make them work in firefox and webkit

manolakis:
## What I did

1. First, I modified the test to be able to reproduce the issue. In order to do that I had to add an option to avoid the use of the snapshots memory cache, and secondly, add a random number in the file content before import it to avoid browser cache.
2. Then, I escaped the carriage returns in the snapshot. Otherwise, Firefox and Webkit seem to have problems.

### DIFF
--- a/packages/test-runner-commands/browser/commands.mjs
+++ b/packages/test-runner-commands/browser/commands.mjs
@@ -97,7 +97,8 @@ export function findAccessibilityNode(node, test) {
 }
 
 let snapshotConfig;
-let snapshots;
+let cachedSnapshots;
+
 export async function getSnapshotConfig() {
   if (!snapshotConfig) {
     snapshotConfig = await executeServerCommand(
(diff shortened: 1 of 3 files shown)


packages/test-runner-commands/browser/commands.mjs
```diff
@@ -110,9 +111,19 @@ export async function getSnapshotConfig() {
   return snapshotConfig;
 }
 
-export async function getSnapshots() {
-  if (snapshots) {
-    return snapshots;
+const ESCAPE_REGEX = /snapshots\[[^\]]+] = (\n)?(?<content>`[^`]+`)/gm;
```
Westbrook, line 114: A little comment of what this regex achieves and I'm good here!

manolakis: done!

PR #1599, opened 2021-07-30
test: unskip test-runner-webdriver tests

### LINKED ISSUE #1223: [test-runner-webdriver] unskip tests
We had to skip the webdriver tests because the chrome version in the CI does not match the version we download. I tried to downgrade, but that did not work. We should unskip when the versions are in sync again.

web-padawan:
## What I did

1. Unskipped the tests to see if they pass now.

Closes #1223


web-padawan:
The tests are failing, will continue investigating this issue.

PR #1611, opened 2021-08-11
feat: add sendKeys support for webdriver

web-padawan:
## What I did

1. Added basic support for `sendKeys` plugin using Webdriver launcher [`browser.keys`](https://webdriver.io/docs/api/browser/keys) API.

For now only `press` and `type` are implemented. I could not quickly figure out whether `down` and `up` are supported.

### DIFF
--- a/.changeset/unlucky-rockets-mate.md
+++ b/.changeset/unlucky-rockets-mate.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-commands': patch
+---
+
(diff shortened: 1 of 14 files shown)


bennypowers:
Does it use the same key strings (same as KeyboardEvent.key)?

Westbrook (approved):
Thanks @web-padawan this looks great!

@bennypowers if you've not any more questions, let's merge this?

PR #1613, opened 2021-08-11
docs: add esbuild plugin to test-runner docs

bennypowers:
This question has come up frequently in the last few days, so it makes sense to highlight it in the docs

## What I did

1. Add a short 'guides' page detailing how to test ts sources using `esbuildPlugin`

### DIFF
--- a/docs/guides/test-runner/typescript.md
+++ b/docs/guides/test-runner/typescript.md
@@ -1,0 +1,36 @@
+# Test Runner >> Testing TypeScript || 70
+
+If you write your source files in TypeScript, you can test directly from sources without
+compiling using `tsc`. Add `esbuildPlugin({ ts: true })` to your `web-test-runner.config.js`
+file.
+This uses esbuild to [transform TS sources on-the-fly](https://esbuild.github.io/api/#transform-api).
+[There are some caveats to using esbuild with TypeScript](https://esbuild.github.io/content-types/#typescript-caveats).
+For example, if you use TypeScript paths to alias imports, You may need to build first.
+
+```js
+import { esbuildPlugin } from '@web/dev-server-esbuild';
+
+export default {
+  files: ['src/**/*.test.ts', 'src/**/*.spec.ts'],
+  plugins: [esbuildPlugin({ ts: true })],
+};
+```
+
+Keep in mind that esbuild merely removes TypeScript syntax and transforms decorators, etc;
+It does not provide any type checking, and it's [not intended to](https://esbuild.github.io/faq/#upcoming-roadmap). If you'd like to run `tsc` in parallel, you can use `concurrently` or `npm-run-all`
+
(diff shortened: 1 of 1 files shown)


Westbrook (commented):
Thought on including a section on how you might do something like:
```
concurrently --kill-others --names tsc,wtr \"npm run tsc:watch\" \"wtr --watch\"
```
Which aligns with the `start` script in the generator (or at least a previous version of it)?

Also, any other concepts (beyond aliases) that might _require_ `tsc`?

bennypowers:
Added `concurrently` example and expounded on caveats @Westbrook

Westbrook:
The build is not happy...I wonder what's up there. I wanted to check out the preview site to see how that `<figure>` content displayed. Do you know of another example I could checkout in the interim?

docs/guides/test-runner/typescript.md
```diff
@@ -0,0 +1,39 @@
+# Test Runner >> Testing TypeScript || 70
+
+If you write your source files in TypeScript, you can test directly from sources without 
+compiling using `tsc`. Add `esbuildPlugin({ ts: true })` to your `web-test-runner.config.js`
+file.
+This uses esbuild to [transform TS sources on-the-fly](https://esbuild.github.io/api/#transform-api).
+Esbuild is not TypeScript, so
+[there are some caveats to using esbuild for testing](https://esbuild.github.io/content-types/#typescript-caveats).
```
web-padawan, line 8: nit: let's update the link to include `caveats when using TypeScript with esbuild` as formulated in its docs.

docs/guides/test-runner/typescript.md
```diff
@@ -0,0 +1,39 @@
+# Test Runner >> Testing TypeScript || 70
+
+If you write your source files in TypeScript, you can test directly from sources without 
+compiling using `tsc`. Add `esbuildPlugin({ ts: true })` to your `web-test-runner.config.js`
+file.
+This uses esbuild to [transform TS sources on-the-fly](https://esbuild.github.io/api/#transform-api).
+Esbuild is not TypeScript, so
```
web-padawan, line 7:
```suggestion
Esbuid has built-in support for parsing TypeScript syntax and discarding the type annotations, but it's not a replacement for the TypeScript compiler, so
```

web-padawan (changes requested):
Overall LGTM, added some comments. Also, please run Prettier to fix the lint.

docs/guides/test-runner/typescript.md
bennypowers: covered that one in the next sentence

bennypowers:
Thanks for the notes everyone.

bennypowers:
merged. @Westbrook if the figure is an issue we can remove it with another PR. should just be default - some inline margin

PR #1618, opened 2021-08-13
Snapshot releases instead of pre-releases

Westbrook:
## What I did

1. Use next as a snapshot release branch

Future work could include chronjobing this to a nightly release... 🤔 

Added @LarsDenBakker for review for the `as Plugin` work that seemed required to catch up to `"@web/rollup-plugin-html": "^1.9.1",`.


web-padawan (approved):
LGTM. I'm wondering about modifying husky folder, is it related to some changes in version 7?

Westbrook:
It is. My other projects got them automatically somewhere, but the file there previously was Husky output that got added to the project incorrectly.

PR #1619, opened 2021-08-13
fix: align @web/test-runner-core version

web-padawan:
## What I did

1. Updated all the packages to use the latest version of `@web/test-runner-core` for consistency.


Westbrook (commented):
I think we've technically not manages these versions that _should_ resolve to a single version since we left the Lerna crowd, though I've definitely felt the issue of it not being there from time to time. If changeset is moving in the direction of managing this, I'm quite interested in where that process goes. This is all well and good if you're comfortable with "version bump only" releases.

Westbrook (approved):

PR #1623, opened 2021-08-14
fix: update Rollup dependency to 2.56.2

web-padawan:
## What I did

1. Updated Rollup dependency to 2.56.2
2. Fixed tests to align with the change landed in 2.55 - https://github.com/rollup/rollup/pull/4182 (see [this article](https://jakearchibald.com/2021/export-default-thing-vs-thing-as-default/) for an overview of how the mutable live bindings work).


Westbrook (approved):
Lgtm!!

PR #1628, opened 2021-08-16
feat: update workbox-build dependency to 6.2.4

web-padawan:
## What I did

1. Updated `workbox-build` to the latest version which comes with types: https://github.com/GoogleChrome/workbox/pull/2867
2. Removed no longer needed `@types/workbox-build`
3. Updated the `generateSw` and `injectManifest` types. Note, `RollupPartial` is used to align with [`WebpackPartial`](https://github.com/GoogleChrome/workbox/blob/0074b20b17b8047a6d9ecd7d9082778a61ad0468/packages/workbox-build/src/types.ts#L169-L179).

### DIFF
--- a/packages/rollup-plugin-workbox/src/index.ts
+++ b/packages/rollup-plugin-workbox/src/index.ts
@@ -1,14 +1,16 @@
-import {
-  generateSW as _generateSw,
-  injectManifest as _injectManifest,
-  GenerateSWConfig,
-  InjectManifestConfig,
-} from 'workbox-build';
+import { generateSW as _generateSw, injectManifest as _injectManifest } from 'workbox-build';
+import type { GenerateSWOptions, InjectManifestOptions } from 'workbox-build';
 import * as prettyBytes from 'pretty-bytes';
 import * as rollup from 'rollup';
 import replace from '@rollup/plugin-replace';
 import { terser } from 'rollup-plugin-terser';
 import resolve from '@rollup/plugin-node-resolve';
+
+interface RollupPartial {
+  mode?: string | null;
+}
+
+type RollupInjectManifestOptions = InjectManifestOptions & RollupPartial;
 
 const name = 'workbox';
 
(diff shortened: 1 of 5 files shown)


packages/rollup-plugin-workbox/src/index.ts
```diff
@@ -1,15 +1,23 @@
...
 import resolve from '@rollup/plugin-node-resolve';
 
+interface RollupPartial {
+  mode?: string | null;
+}
+
+type RollupGenerateSWOptions = GenerateSWOptions & RollupPartial;
+
+type RollupInjectManifestOptions = InjectManifestOptions & RollupPartial;
+
```
Westbrook, line 20: It feels weird to need to construct types here. Are they not included in the new version or an updated types packages?

web-padawan: The new version includes following types:

- `GenerateSWOptions`
- `WebpackGenerateSWOptions`
- `InjectManifestOptions`
- `WebpackInjectManifestOptions`

The problem comes from `InjectManifestOptions` type which does not support `mode` property.
At the same time, `WebpackInjectManifestOptions` does support `mode`, but not `globDirectory`.

So I decided to create types for both, even though I could use `GenerateSWOptions` as is.
This is done for consistency and to ensure we can rely on `mode` also in `generateSW`.

If you think types need to be in a separate file `types.ts`, let's move them there.

packages/rollup-plugin-workbox/src/index.ts
```diff
@@ -1,15 +1,23 @@
 import {
   generateSW as _generateSw,
   injectManifest as _injectManifest,
-  GenerateSWConfig,
-  InjectManifestConfig,
+  GenerateSWOptions,
```
abdonrd, line 4: We can use the `import type {}`?

packages/rollup-plugin-workbox/src/index.ts
thepassle: We support the `mode` property for `injectManifest`, which workbox itself doesnt support, so to me there it makes sense. It shouldnt be needed for `generateSw` though, because workbox should support the `mode` property there already

packages/rollup-plugin-workbox/src/index.ts
web-padawan: Sure. I will update the PR once we agree about how to proceed regarding the above comment.

packages/rollup-plugin-workbox/src/index.ts
web-padawan: Thanks, indeed `GenerateSWOptions` contains this property. Updated to only add one type.

packages/rollup-plugin-workbox/src/index.ts
web-padawan: Done.

bennypowers (approved):

PR #1630, opened 2021-08-17
grammar tweaks

davidmaxwaterman:
Just some grammar tweaks.

## What I did

1. Main idea is to make it clear that both the skills and code remain relevant, rather than just the skills; and that they do so because they track the standards/platform (rather than some 'here today, gone tomorrow' library).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,11 +9,11 @@ reasons:
   [
     {
       header: Built on web standards,
-      text: "Work with and extend what's available in modern browsers, learning skills, and writing code that stays relevant.",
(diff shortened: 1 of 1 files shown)


davidmaxwaterman:
> Hey @davidmaxwaterman! Sorry for the long time this has taken to get reviewed sob
> 
> I wanted to merge this, but I think it needs to be updated. Can you update/rebase `master` and push those changes? Then, if the tests pass, I'll merge this.

Done. Thanks.

koddsson:
Thanks 🙌🏻

PR #1633, opened 2021-08-19
fix: bump @rollup/plugin-node-resolve to v13

### LINKED ISSUE #1376: [dev-server-esbuild] Unable to resolve packages with correct extensions when imported in an .ts file using node's `exports`
When files being imported in an `.ts` file, the resolver automatically assumes all the imports must be ended with `.ts` when some could be `.js`. It seems like the dev server does not support node's `exports` quite well.

## Actual behavior

Packages that use `exports` in node resolves with `.ts` when it gets imported in a `.ts` file.

## Expected behavior

Packages that use `exports` in node must be resolved with the correct extensions when it gets imported in a `.ts` file

## Steps to reproduce

Clone this [repo](https://github.com/motss/wds-import-bug) and run the project to reproduce the stated problem.

### LINKED ISSUE #1443: [web-test-runner] plugin_node_resolve_1.nodeResolve on node 14 + .mjs config

### LINKED ISSUE #1568: [dev-server-rollup] Failing test: node modules resolved outside root directory are rewritten with commonjs
There is one test failing after upgrading `@rollup/plugin-node-resolve` to 13.0.0 (#1433).

```sh
  1) @rollup/plugin-node-resolve
       node modules resolved outside root directory are rewritten with commonjs:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/web/web/packages/dev-server-rollup/test/node/plugins/node-resolve.test.ts)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

See https://github.com/modernweb-dev/web/pull/1567/checks?check_run_id=3159207368

Downgrading back to `11.0.0` makes this test pass. The reason is most likely this PR: https://github.com/rollup/plugins/pull/799

### LINKED ISSUE #1575: [dev-server-rollup] Update @rollup/plugin-node-resolve to v13
The version has been downgraded by #1569 to make the failing test pass.
See #1568 for the example of a test failing with v13.

We need to investigate the failing test and then upgrade the plugin back. This should address these issues:
- https://github.com/modernweb-dev/web/issues/1376#issuecomment-886153882
- https://github.com/modernweb-dev/web/issues/1443

web-padawan:
## What I did

1. Updated `@rollup/plugin-node-resolve` to v13 (again)

Fixes #1376
Fixes #1443 
Fixes #1575 
Fixes #1568

### DIFF
--- a/.changeset/yellow-rockets-argue.md
+++ b/.changeset/yellow-rockets-argue.md
@@ -1,0 +1,7 @@
+---
+'@web/dev-server': patch
+'@web/dev-server-rollup': patch
+'rollup-plugin-workbox': patch
+---
+
+Update dependency @rollup/plugin-node-resolve to v13
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "devDependencies": {
     "@changesets/cli": "^2.16.0",
-    "@rollup/plugin-node-resolve": "^11.0.1",
+    "@rollup/plugin-node-resolve": "^13.0.4",
     "@rollup/plugin-typescript": "^8.1.0",
     "@types/chai": "^4.2.14",
     "@types/mocha": "^8.2.0",
--- a/packages/dev-server-rollup/package.json
+++ b/packages/dev-server-rollup/package.json
@@ -58,7 +58,7 @@
     "@rollup/plugin-alias": "^3.1.1",
     "@rollup/plugin-babel": "^5.2.2",
     "@rollup/plugin-commonjs": "^18.0.0",
-    "@rollup/plugin-node-resolve": "^11.0.1",
+    "@rollup/plugin-node-resolve": "^13.0.4",
     "@rollup/plugin-replace": "^3.0.0",
     "@types/parse5": "^6.0.1",
     "@types/whatwg-url": "^8.0.0",
--- a/packages/dev-server-rollup/package.json
+++ b/packages/dev-server-rollup/package.json
@@ -57,7 +57,7 @@
     "@babel/plugin-transform-template-literals": "^7.12.1",
     "@rollup/plugin-alias": "^3.1.1",
     "@rollup/plugin-babel": "^5.2.2",
-    "@rollup/plugin-commonjs": "^18.0.0",
+    "@rollup/plugin-commonjs": "^20.0.0",
     "@rollup/plugin-node-resolve": "^13.0.4",
     "@rollup/plugin-replace": "^3.0.0",
     "@types/parse5": "^6.0.1",
--- a/packages/dev-server-rollup/src/createRollupPluginContextAdapter.ts
+++ b/packages/dev-server-rollup/src/createRollupPluginContextAdapter.ts
@@ -69,10 +69,7 @@ export function createRollupPluginContextAdapter<
       if (!context) throw new Error('Context is required.');
 
       for (const pl of config.plugins ?? []) {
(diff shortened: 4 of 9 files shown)


bennypowers:
does the failing test mean users won't be able to use both plugins? 🤔

web-padawan:
Here is one issue which could potentially affect other plugins: https://github.com/rollup/plugins/issues/973

See also [related code](https://github.com/rollup/plugins/blob/6463b9635e20b8425599f14005b9efdd8f94ecd6/packages/node-resolve/src/index.js#L267-L270) which returns `false` for external modules. This could be the problem.

bennypowers:
Trying to figure out what's going on here, I noticed the test fails because it gets caught in an infinite loop where `resolveId` is called over and over with the same parameters:
```
resolveId {
  importee: '/Users/bennyp/Developer/web/packages/dev-server-rollup/test/node/fixtures/resolve-outside-dir/node_modules/module-a/index.js',
  importer: '/Users/bennyp/Developer/web/packages/dev-server-rollup/test/node/fixtures/resolve-outside-dir/src/app.js'
}
```

I'm not sure yet what is causing that loop

bennypowers:
@tjenkinson would you mind taking a look if you have a moment? I suspect our loop is caused by the changes to rollup/plugin-node-resolve's use of rollup api (`this.resolve`), and the changes you made in https://github.com/rollup/plugins/commit/3a543dfd9142f936b0157893580c25f4cb729b0f

bennypowers:
Thanks @tjenkinson that was very helpful.

@LarsDenBakker PTAL, thanks.

LarsDenBakker:
@bennypowers your change will always skip over all plugins if skipSelf is true. We need to mimick what's being done inside rollup, where when one of the other plugins calls this.resolve again from its resolveId hook with the exact same source and importer, the initial plugin will be skipped as well.

Marshal27:
Ended up here due to not being able to import side-effects for selector registrations in the customElements registry... thank you all for the work on this one, I will be eagerly watching this in hopes that it lands soon.

gloriarodrife:
Any news? Thanks!

Tansito:
I was following this PR and really thank you for the work! I assume there was no advances here, was it? 🤞

daKmoR:
quite the investigation happened here... seems to be related to a change in the commonjs plugin 🤔

this "mixed" approach got superseded by https://github.com/modernweb-dev/web/pull/1942 which only updates the plugin-node-resolve

updating the dev dependencies plugin-commonjs and/or rollup may be done in a separate step? Is anyone up for it? (beware there may be dragons 😅 as with everything related to esm <-> cjs 🙈)

PR #1651, opened 2021-08-30
feat(test-runner-browserstack): migrate `BrowserstackLauncher` to extend from `WebdriverLauncher` instead of `SeleniumLauncher`

melink14:
## What I did

Replaced `SeleniumLauncher` with `WebdriverLauncher` and updated the way options were passed to the parent class.


packages/test-runner-visual-regression/src/visualRegressionPlugin.ts
```diff
@@ -109,6 +110,24 @@ export function visualRegressionPlugin(
             return visualDiffCommand(mergedOptions, screenshot, session.browser.name, payload.name);
           }
 
+          if (session.browser.type === 'selenium') {
```
melink14, line 113: Since this branch is identical to the `webdriver` branch except for type it might be good to combine the branches and use a union type when defining `browser`...

melink14:
I was thinking about testing and realized that even if we can't run small selenium based tests we could test this functionality as part of the BrowserStack CI since that's an easy stable way to get access to a Selenium server.

For example, as part of the browserstack verify workflow we could just run the tests from the test-runner-selenium package as well as the visual regression package with a flag that makes the tests run only on selenium...

Not sure the complexity is worth it but wanted to share the idea since my own testing was via BrowseStack...

web-padawan (commented):
I have tried this approach previously and for me it was quite slow when using in Sauce Labs.
The sequence "switch to frame, take screenshot, switch back" was up to 10 seconds.

melink14:
> I have tried this approach previously and for me it was quite slow when using in Sauce Labs.
> The sequence "switch to frame, take screenshot, switch back" was up to 10 seconds.

Oh, interesting. I wasn't sensitive to the latency due to the setup time being so long on browserstack.

That being said, I'm not sure there's another option given the use of iframes... (Webdriver launcher has the same limitation of course)

Maybe we can also add a warning to the docs; iirc the docs already mention that selenium/browserstack are slower than puppeteer/playwright.

web-padawan:
@melink14 Sorry about the delay with reviewing. Have you tested the latency? Do you think it's good enough?

Alternative option could be to update Browserstack launcher to use `@web/test-runner-webdriver` and use it.

melink14:
@web-padawan No worries and thanks for the comment.

When testing via Browserstack, the latency increase wasn't noticeable to me given the long base time. I'll see about doing some more scientific timing later.

That being said, it's an interesting idea to have Browserstack launcher depend on `@web/test-runner-webdriver` instead. I had assumed there was some key reason it needed to inherit from Selenium but I'll have to investigate a bit and refresh my understanding. :)

web-padawan:
In fact, when I implemented `@web/test-runner-wevdriver` and then updated `@web/test-runner-saucelabs` to use it, I didn't include the Browserstack launcher purely because of my lack of experience with it. So I would appreciate if you give some pointers on the necessary changes to get the launcher updated.

melink14:
It seems well supported so I'll attempt it and if it works will submit a new pull request in lieu of this.

ref:
https://www.browserstack.com/docs/automate/selenium/getting-started/nodejs/webdriverio#integrating-your-tests-with-browserstack
https://webdriver.io/docs/cloudservices#browserstack

melink14:
@web-padawan I migrated `BrowserstackLauncher` to extend `WebdriverLauncher` and it was fairly straight forward. Tests passing locally.

I might make an issue for a follow up to migrate to using the new W3C spec based capabilities in the future since there's nice types for those APIs.

web-padawan:
Thanks for the PR. It looks good to me as long as you tested that it works with a real Browsertack account 👍

web-padawan:
Changed the base branch, will merge it there and then create a separate PR to trigger Browserstack tests.

web-padawan (approved):

PR #1655, opened 2021-09-01
feat(test-runner-visual-regression): output diff file path on failure

WickyNilliams:
this makes it easy and quick to open the diff from the CLI, inspect the results, and work out what went wrong

## What I did

1. Append file path for diff image to existing error message

### DIFF
--- a/packages/test-runner-visual-regression/src/visualDiffCommand.ts
+++ b/packages/test-runner-visual-regression/src/visualDiffCommand.ts
@@ -44,6 +44,7 @@ export async function visualDiffCommand(
 
   const diffName = options.getDiffName({ browser, name });
   const failedName = options.getFailedName({ browser, name });
+  const diffFilePath = resolveImagePath(baseDir, diffName);
 
   const saveFailed = async () => {
     await options.saveFailed({
@@ -56,7 +57,7 @@ export async function visualDiffCommand(
 
   const saveDiff = async () => {
     await options.saveDiff({
-      filePath: resolveImagePath(baseDir, diffName),
+      filePath: diffFilePath,
       baseDir,
       name: diffName,
       content: diffImage,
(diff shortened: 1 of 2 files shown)


packages/test-runner-visual-regression/src/visualDiffCommand.ts
```diff
@@ -100,7 +100,9 @@ export async function visualDiffCommand(
 
   return {
     errorMessage: !passed
-      ? `Visual diff failed. New screenshot is ${diffPercentage.toFixed(2)} % different.`
+      ? `Visual diff failed. New screenshot is ${diffPercentage.toFixed(
+          2,
+        )}% different.\nSee diff for details: ${resolveImagePath(baseDir, diffName)}`
```
web-padawan, line 105: This looks like doing the same work twice. How about changing the method above?

```js
const saveDiff = async () => {
  const diffPath = resolveImagePath(baseDir, diffName);
  await options.saveDiff({
    filePath: diffPath,
    baseDir,
    name: diffName,
    content: diffImage,
  });
  return diffPath;
};
```

Then we can declare the variable above:

```js
let diffPath
```

And update both calls accordingly:

```js
diffPath = await saveDiff();
```

So that we can later include this into the error message.

web-padawan (commented):
Please check my comment, and if you update this PR, I will approve and merge 👍

packages/test-runner-visual-regression/src/visualDiffCommand.ts
WickyNilliams: Sounds good to me. Will update tomorrow

web-padawan: Thanks! And then please aldo add xhangeset using `npx changeset` as described in the above comment.

WickyNilliams: Done! Not exactly as you described, but it has same result (I calculate it outside of `saveDiff` to avoid changing function signatures). I also added a changeset and rebased/squashed

web-padawan: Thanks! Your approach does make sense and looks better than what I suggested 👍

web-padawan (approved):
LGTM 💯

Issue #1642, opened 2021-08-24
Make web-test-runner and web-dev-server pure ESM packages


web-padawan:
### Motivation

Once Node.js 16 becomes LTS, it will become a third LTS version with native ES modules support.

Some maintainers have already started migrating their packages to be pure ES modules.
See [the gist](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) for an explanation of what it a pure ESM package.

In practice, it means that we **drop support for CommonJS** and drop support for older node versions (10.x).

### Dependencies

Making `@web` packages pure ES modules would enable us to use dependencies that are pure ES modules too:

- `cli-cursor@4` - #1641
- `estree-walker@3` - #1241
- `globby@12` - #1583

There might be other dependencies that get major bumps with dropping support for CommonJS in future.

### Proposed solution

1. Update `"exports"` field to remove `"require"` entrypoints for all affected packages
2. Add `"type": "module"` to `package.json` to indicate packages are pure ES modules
3. Maybe also a few other tweaks, see [How can I move my CommonJS project to ESM?](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm)

Westbrook:
I like it! Would we also be in a place to use this change as he switch to 1.0?

web-padawan:
Some PRs that also not directly related but need to be merged before 1.0:

- `mocha@9` https://github.com/modernweb-dev/web/pull/1498
-  `ts-node@10` https://github.com/modernweb-dev/web/pull/1469

PR #1661, opened 2021-09-05
chore: update node-fetch to 3.0.0 stable

web-padawan:
## What I did

1. Updated `node-fetch` to 3.0.0 stable
2. Updated `@types/node-fetch` accordingly

Closes #1660 

Note: no changeset is added, however `dev-server-core` imports `node-fetch` in `src/test-helpers.ts` 
These helpers are only used by tests for other packages in this monorepo so presumably it's ok.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "@types/chai": "^4.2.14",
     "@types/mocha": "^8.2.0",
     "@types/node": "*",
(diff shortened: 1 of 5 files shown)


web-padawan:
Looks like `ndoe-fetch` 3.0.0 stable is now ESM-only package, so I'm closing this in favor of #1642

Issue #1557, opened 2021-07-20
[browser-logs] Error while deserializing browser logs for functions inside objects


jorenbroekema:
In any test, you can write:

```js
const foo = { 'my-element': locale => 'bar' };
console.log(foo);
```

Which will result in `null` as output and the following error thrown by `@web/browser-logs`

![image](https://user-images.githubusercontent.com/36734656/126286446-137ce580-a8ea-44ee-8649-693d0acc0dc2.png)

```js
const foo = { 'my-element': 'bar' };
console.log(foo);
```

The above code works like expected, so I think it has something to do with functions inside objects.

WickyNilliams:
I just got the same error when logging:

```ts
console.log(e.composedPath())
```

Testing a little more, logging a function on its own is fine, as this _does not_ cause errors:

```ts
console.log(() => 1)
```

So it does seem to be related specifically to functions inside objects as @jorenbroekema pointed out

PR #1668, opened 2021-09-09
fix: support method names in objects containing a dash

thepassle:
fixes https://github.com/modernweb-dev/web/issues/1557

## What I did

1. The serializer currently crashes on function names containing a hyphen (in objects, e.g. `{'a-b': ()=>{}}`, because the deserializer tries to revive them like so:

```js
new Function(`return function my-function() {}`);
```
which is invalid

### DIFF
--- a/.changeset/cuddly-snails-beg.md
+++ b/.changeset/cuddly-snails-beg.md
@@ -1,0 +1,5 @@
+---
+"@web/browser-logs": patch
+---
+
+fix: support method names in objects containing a dash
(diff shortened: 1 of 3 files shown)


bennypowers (approved):

LarsDenBakker:
Doesn't this lose the function body? Its always empty? Maybe only put the function name if it doesnt have hyphen?

thepassle:
> Doesn't this lose the function body? Its always empty? Maybe only put the function name if it doesnt have hyphen?

For 'regular' functions we also don't log the function body:

https://github.com/modernweb-dev/web/blob/bc72c6c67e5cdb4d0146462e103b9252950e4703/packages/browser-logs/src/deserialize.ts#L51-L56

LarsDenBakker:
Oh, right.

PR #1685, opened 2021-09-22
refactor: replace chalk with nanocolors

web-padawan:
## What I did

1. Replaced all `chalk` usage with the new `nanocolors` library.

See their README for some benefits: https://github.com/ai/nanocolors/tree/main#nano-colors


Westbrook (approved):
LGTM!

web-padawan:
Thanks, for some reason the git hook didn't work 🤔

packages/dev-server-rollup/src/rollupAdapter.ts
```diff
@@ -172,7 +172,7 @@ export function rollupAdapter(
             !['/', './', '../'].some(prefix => resolvableImport.startsWith(prefix)) &&
             adapterOptions.throwOnUnresolvedImport
           ) {
-            const errorMessage = red(`Could not resolve import ${cyanBright(`"${source}"`)}.`);
+            const errorMessage = red(`Could not resolve import ${cyan(`"${source}"`)}.`);
```
daKmoR, line 175: cyan !== cyanBright ?

web-padawan: Yes it's not quite the same. See available colors [here](https://github.com/ai/nanocolors/blob/44908c1348e3137f6ede3b80bac74766b6a58244/example.png). Any better ideas?

daKmoR: Screenshot 2021-09-22 at 18 45 22

ok the difference is not that big...


once thing I could not find is... can I disable the colors without a system variable? 🤔

```js
import { options } from 'colorette';
describe('cli', () => {
  before(() => {
    // ignore colors in tests as most CIs won't support it
    options.enabled = false;
  });
```

web-padawan: Functions will use colors only if Nano Colors auto-detect that current environment supports colors.
There is an option to override color support auto detection, [example](https://github.com/ai/nanocolors/tree/44908c1348e3137f6ede3b80bac74766b6a58244#conditional-support).

daKmoR: sorry for being picky 😅
change looks good.

@LarsDenBakker you ok with the color change? you picked those initially I think

LarsDenBakker: Fine with me

PR #1687, opened 2021-09-22
refactor: replace uuid with nanoid

web-padawan:
## What I did

1. Replaced `uuid` dependency with `nanoid`

See the README for this package: https://github.com/ai/nanoid#nano-id

### DIFF
--- a/.changeset/khaki-boxes-pretend.md
+++ b/.changeset/khaki-boxes-pretend.md
@@ -1,0 +1,8 @@
+---
+'@web/test-runner-browserstack': patch
+'@web/test-runner-core': patch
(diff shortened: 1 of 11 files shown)


Westbrook (commented):
Is `crypto.randomUUID( )` not yet available in these contexts?

web-padawan:
Based on the [docs](https://nodejs.org/api/crypto.html#crypto_crypto_randomuuid_options), `crypto.randomUUID` was added in Node 15.6.0 and we still support 12.x and 14.x.

LarsDenBakker (approved):

PR #1695, opened 2021-09-27
fix: use encodeURIComponent to escape snapshot content

### LINKED ISSUE #1690: @web/test-runner fails to load snapshot
The test runner can successfully save snapshots but when using `compareSnapshot` and sending the snapshot to the browser, if the content contains a character like `#` it fails [in this line](https://github.com/modernweb-dev/web/blob/adeab906c0ac9a4ec2291e8a7088f5bf63394f03/packages/test-runner-commands/browser/commands.mjs#L148) with the following message: `SyntaxError: unexpected end of input`.

I can make the error go away by using `encodeURIComponent` before saving the snapshot, but if I use `compareSnapshot`, even with a successful match the snapshot will be updated without encoding so it fails in the next run. Because of this, I have to add the logic to load, compare and save the snapshot myself.

alfonsogarciacaro:
Fix #1690 by using `encodeURIComponent` to escape the snapshot content before building the data url instead of just escaping the line breaks. Please let me know if there's a better way to do this.

### DIFF
--- a/.changeset/pink-lobsters-develop.md
+++ b/.changeset/pink-lobsters-develop.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-commands': patch
+---
+
+Fix snapshots with # character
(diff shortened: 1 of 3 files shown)


web-padawan (approved):
LGTM. @LarsDenBakker could you take a look when you have time?

LarsDenBakker:
Looks like a good solution for this

LarsDenBakker:
Looks like a good solution for this,  thanks!

alfonsogarciacaro:
Thanks for merging and for the test runner, really enjoying it! :clap:

PR #1696, opened 2021-09-27
fix(docs): example plugin should return undefined, not false

WickyNilliams:
The docs specifically say that you should return `undefined`/`null `to signal a command _hasn't_ been handled. However, the example plugin returns `false` in this case. 

In my testing, it seems that `false` is treated as having handled the command, and so no other commands ever get executed. I just spent time debugging why one of my commands wouldn't execute, and eventually tracked it down to having based my plugins on this example :)


WickyNilliams:
Aside: the changeset requirement makes it hard to make quick PRs via github's web UI. I wonder if it's possible to automate this?

PR #1763, opened 2021-11-05
feat: add hover command

### LINKED ISSUE #1758: [test-runner-commands] Add a command for triggering the `:hover` state
## Motivation

The standard Web API doesn't make it possible to enable the `:hover` state for an element programmatically. Although you can synthesize mouse events on the element with JavaScript, it still won't cause the element to get the `:hover` state. It would be useful if there was a command provided by the test runner for such a purpose based on the browser's internal API.

### Puppeteer

The `page.hover(selector)`  method can be used to hover over an element by a CSS selector:

https://pptr.dev/#?product=Puppeteer&version=v10.4.0&show=api-pagehoverselector

### Playwright

The `page.hover(selector[, options])` method can be used to hover over an element by a CSS selector:

https://playwright.dev/docs/api/class-page#page-hover

### WebDriver

The `browser.elementHover(elementId)` method can be used to hover over an element in Chromium with the caveat that the element can be only selected by id:

https://webdriver.io/docs/api/chromium/#elementhover

## Command API proposal

Introduce an asynchronous `hover` command that would trigger the `:hover` state for an element by id, or by a CSS selector if there is a way to make that work for WebDriver.

```js
import { hover } from '@web/test-runner-commands';

describe('element', () => {
  it('should be blue by default', () => {
    expect(element.getComputedStyle().color).to.equal('blue');
  });
  
  it('should become red on hover', async () => {
    await hover('#element');
    
    expect(element.getComputedStyle().color).to.equal('red');
  });
});
```

vursen:
Fixed #1758


packages/test-runner-commands/browser/commands.d.ts
```diff
@@ -30,6 +30,16 @@ export function setViewport(viewport: Viewport): Promise<void>;
...
+ * ```ts
+ *    await emulateHover('.wrapper > .item');
+ * ```
+ */
+export function emulateHover(selector: string): Promise<void>;
+
+ /**
  * Sets the user agent of the browser. This is a puppeteer-only API.
  */
 export function setUserAgent(userAgent: string): Promise<void>;
```
web-padawan, line 45:
```suggestion
/**
 * Sets the user agent of the browser. This is a puppeteer-only API.
 */
export function setUserAgent(userAgent: string): Promise<void>;
```

web-padawan (changes requested):
Please run `npm run format` to make lint pass, then I will take a closer look.

packages/test-runner-commands/browser/commands.mjs
```diff
@@ -61,6 +61,10 @@ export function emulateMedia(media) {
   return executeServerCommand('emulate-media', media);
 }
 
+export function emulateHover(options) {
```
web-padawan, line 64: I would suggest this to `hover` because we don't **emulate** it (unlike media query command) but trigger the browser API to actually move cursor over the element so it gets `:hover` state.

packages/test-runner-commands/test/emulate-hover/emulateHoverPlugin.test.ts
```diff
@@ -0,0 +1,122 @@
...
+import { TestRunnerPlugin } from '@web/test-runner-core';
+import { runTests } from '@web/test-runner-core/test-helpers';
+import { chromeLauncher } from '@web/test-runner-chrome';
+import { webdriverLauncher } from '@web/test-runner-webdriver';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+import { visualRegressionPlugin } from '@web/test-runner-visual-regression/plugin';
+
+import { emulateHoverPlugin } from '../../src/emulateHoverPlugin';
+
+import { startSeleniumServer } from '../selenium-server.js';
```
web-padawan, line 13: Let's convert the file to `selenium-server.ts` to mitigate "Could not find a declaration file" error.

packages/test-runner-commands/src/emulateHoverPlugin.ts
```diff
@@ -0,0 +1,39 @@
...
+          return true;
+        }
+
+        if (session.browser.type === 'webdriver') {
+          const browser = session.browser as WebdriverLauncher;
+          await browser.hover(session.id, payload);
+          return true;
+        }
+
+        throw new Error('emulating hover is only supported on puppeteer.');
```
web-padawan, line 34:
```suggestion
        throw new Error(
          `Browser launcher ${session.browser.type} does not support hover command at the moment.`,
        );
```

web-padawan (changes requested):
A few more comments, please address them so that build actually runs tests in CI.
Also, please rebase to pick up the latest changes (see also #1765 for Firefox tests).

packages/test-runner-commands/browser/commands.mjs
vursen: Good point, I'll rename it.

Westbrook:
How do you feel about `hover` being a helper around something like https://github.com/modernweb-dev/web/discussions/1397 rather than being its own API? Maintaining this sort of hyper specific API directly into the commands architecture seems like it opens the door for a lot of tedious APIs that require a good deal of maintenance over time...

web-padawan:
@Westbrook Good point, we can agree about a more generic API. How about something along these lines:

```js
await sendPointer({ type: 'hover', selector: '.wrapper > .item' });
await sendPointer({ type: 'click' });
```

This seems more in line with `sendKeys()` API command, but more generic than `sendMouse` you proposed.

packages/test-runner-commands/test/emulate-hover/emulateHoverPlugin.test.ts
vursen: Done.

packages/test-runner-commands/src/emulateHoverPlugin.ts
vursen: Done.

packages/test-runner-webdriver/src/SessionManager.ts
```diff
@@ -84,6 +84,27 @@ export class SessionManager {
...
+        type: 'pointer',
+        id: 'finger1',
+        parameters: {
+          pointerType: 'mouse',
+        },
+        actions: [
+          {
+            type: 'pointerMove',
+            duration: 0,
+            x: x + 1,
```
Westbrook, line 100: Other runner _generally_ click/hover/etc in the middle of an element, does that matter?

Westbrook:
I've worked with a [version of this](https://github.com/adobe/spectrum-web-components/blob/main/test/plugins/send-mouse-plugin.ts) that could be the basis of this. I like the idea of surfacing both the `position` and the `selector` API.

I wonder if having the `selector` API implied the presence of a `host` property to the options.

I'd like to hear if you gave a good reason for preferring `hover` over `move` or `moveTo` as the "type" in a "hover" action. Not all contexts support "hover" and having the testing API support shaping people's understanding of their code seems useful.

I do think you need all of the "types": `type: 'move' | 'down' | 'up' | 'click';` and I would greatly prefer an API that accepted an array of "actions" or "steps" as you'll see in the API above.

packages/test-runner-webdriver/src/SessionManager.ts
web-padawan: I agree it makes more sense to use something like [`middleOfNode`](https://github.com/vaadin/testing-helpers/blob/6e66938563adff75158374801b9af1948539f9ba/src/pointer.ts#L13) here.

vursen:
I'd like to note that moving a pointer to an element by coordinates or by a selector may actually not cause that element to get the hover state if it is positioned under another element, while the hover command (it doesn't matter if it is shaped as a command or it is one of the possible actions of a generic command like `sendMouse`) would cover this case, at least in Puppeteer and Playwright.

Now about what follows from above for the current implementation of the hover command. It appears that it might happen that for the same selector the command would lead to different elements getting hovered in WebDriver and in Puppeteer / Playwright as WebDriver doesn't **hover** over an element but rather **moves** the pointer to the element's coordinates. The only work-around I can think of is to use the `hoverElement` API in WebDriver. Another way could be to completely refuse the hover command's support for WebDriver.

Westbrook:
I’d say that the test hovering something that isn’t the top of the element stack is a big false positive for testing. I feel pretty strongly about not supporting a test case for “hover” unless it could actually happen for a user with a pointing device at that time.

vursen:
Well, I agree, it is not what the user is expected to do. Although, to me, the above also sounds like an argument for not supporting selectors in the `sendMouse`-like API.

What would you expect calling `sendMouse({ type: 'down', selector: '.element' })` if the element is overlapped? Perhaps, the `coordinates` option is the most obvious API from that perspective.

web-padawan:
> Another way could be to completely refuse the support for WebDriver for the command.

Let's try to provide some minimal functionality to cover as much browser launchers as possible.

Note that WebDriver implements an extension of [W3C protocol](https://www.w3.org/TR/webdriver/) which we should use as a guidance.
By the way, I'd prefer to provide API that would work in `@web/test-runner-selenium` too.

Westbrook:
We should test, but at least in Playwright the "hover" functionality reads like "move mouse to center of the element":

> Use page.mouse to hover over the center of the element, or the specified position.

Which would only `:hover` if the element was appropriately related to the top of the stack. But, I read this from the idea that it manages a "real mouse" as opposed to "synthetic events", which may be incorrect.

web-padawan (commented):
Let's not try to come up with universal approach right away. This PR is only a first step.

The way I see it, we could add the API that might look roughly like this:

```js
await sendPointer({ type: 'click' });
await sendPointer({ type: 'move' });
```

Speaking of testing, we can use [`MouseMoveController`](https://lit.dev/docs/composition/controllers/#use-cases) from Lit docs in visual tests.

web-padawan:
Closing in favor of #1783. @Westbrook please add your comments to the new PR if you have some.

PR #1767, opened 2021-11-06
fix: do not process scripts with non-JS type attribute

smartrejames:
## What I did

1. dev-server-esbuild: fix "importmap" or other non-js type in html file


web-padawan (approved):
Thanks for the PR, the solution looks good to me 👍

PR #1779, opened 2021-11-11
fix(test-runner-visual-regression): improve output when very small diff percentage

WickyNilliams:
## What I did

1. in cases where diff percentage was very small the output would be like `New screenshot is 0.00% different` which is confusing
2. i kept formatting to 2 decimal places when the value is large enough, otherwise output the raw value
3. this avoid cases as in (1)

### DIFF
--- a/.changeset/curly-cameras-deliver.md
+++ b/.changeset/curly-cameras-deliver.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-visual-regression': patch
+---
+
+improve output in visual regression tests when very small differences from baseline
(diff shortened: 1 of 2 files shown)


web-padawan (approved):
Thanks for the PR, we faced this problem in some of our tests and it's very annoying.

WickyNilliams:
Yeah it was confusing. I thought it was a false negative at first, but the diff image showed some pixel mismatches. So I realised it was being reported incorrectly. I also saw someone on slack get confused by this, which finally spurred me to fix. 

Thanks for merging!

PR #1783, opened 2021-11-13
feat: add sendMouse command

vursen:
## What I did

- [x] Introduced a new command `sendMouse`.
- [x] Added basic unit tests.
- [x] Added support for Playwright.
- [x] Added support for Puppeteer.
- [x] Added support for WebDriver.
- [x] Make unit tests work in Playwright Webkit.
- [x] Make unit tests work in Windows Chrome.

### Move
```ts
function sendMouse(payload: {
  type: 'move',
  position: [number, number]
}): Promise<void>;
```

### Click
```ts
function sendMouse(payload: {
  type: 'click',
  position: [number, number],
  button?: 'left' | 'middle' | 'right'
}): Promise<void>;
```

### Down
```ts
function sendMouse(payload: {
  type: 'down',
  button?: 'left' | 'middle' | 'right'
}): Promise<void>;
```

### Up
```ts
function sendMouse(payload: {
  type: 'up',
  button?: 'left' | 'middle' | 'right'
}): Promise<void>;
```

## Future improvements

It is possible to extend the command's API in the future to also support sending a sequence of mouse actions, e.g.:

```js
await sendMouse([
  { type: 'move', position: [100, 100] },
  { type: 'move', position: [200, 200] }
]);
```

### DIFF
--- a/packages/test-runner-commands/src/sendMousePlugin.ts
+++ b/packages/test-runner-commands/src/sendMousePlugin.ts
@@ -1,6 +1,7 @@
 import { TestRunnerPlugin } from '@web/test-runner-core';
 import type { ChromeLauncher } from '@web/test-runner-chrome';
 import type { PlaywrightLauncher } from '@web/test-runner-playwright';
+import type { WebdriverLauncher } from '@web/test-runner-webdriver';
 
 type MousePosition = [number, number];
 type MouseButton = 'left' | 'middle' | 'right';
@@ -102,6 +103,25 @@ export function sendMousePlugin(): TestRunnerPlugin<SendMousePayload> {
               return true;
             case 'up':
               await page.mouse.up({ button: payload.button });
+              return true;
+          }
+        }
+
+        // handle specific behavior for webdriver
+        if (session.browser.type === 'webdriver') {
+          const page = session.browser as WebdriverLauncher;
+          switch(payload.type) {
+            case 'move':
+              await page.sendMouseMove(session.id, payload.position[0], payload.position[1]);
+              return true;
+            case 'click':
+              await page.sendMouseClick(session.id, payload.position[0], payload.position[1], payload.button);
+              return true;
+            case 'down':
+              await page.sendMouseDown(session.id, payload.button);
+              return true;
+            case 'up':
+              await page.sendMouseUp(session.id, payload.button);
               return true;
           }
         }
--- a/packages/test-runner-commands/src/sendMousePlugin.ts
+++ b/packages/test-runner-commands/src/sendMousePlugin.ts
@@ -52,7 +52,8 @@ function isSendMousePayload(payload: unknown): payload is SendMousePayload {
       !Array.isArray(payload.position) ||
       typeof payload.position[0] !== 'number' ||
       typeof payload.position[1] !== 'number' ||
-      !/^\d+$/.test(payload.position.join(''))
+      !Number.isInteger(payload.position[0]) ||
+      !Number.isInteger(payload.position[1])
     ) {
       throw new Error(
         'You must provide a position option as a [x, y] tuple where x and y are integers.',
--- a/packages/test-runner-commands/src/sendMousePlugin.ts
+++ b/packages/test-runner-commands/src/sendMousePlugin.ts
@@ -1,0 +1,104 @@
+import { TestRunnerPlugin } from '@web/test-runner-core';
+import type { ChromeLauncher } from '@web/test-runner-chrome';
+import type { PlaywrightLauncher } from '@web/test-runner-playwright';
+
+export type SendMousePayload = MovePayload | ClickPayload | DownPayload | UpPayload;
+
+export type MovePayload = {
+  type: 'move',
+  position: [number, number]
+};
+
+export type ClickPayload = {
+  type: 'click',
+  position: [number, number],
+  // TODO: button: 'left' | 'right' | 'middle'
+}
+
+export type DownPayload = {
+  type: 'down',
+  // TODO: button: 'left' | 'right' | 'middle'
+}
+
+export type UpPayload = {
+  type: 'up',
+  // TODO: button: 'left' | 'right' | 'middle'
+}
+
+function isObject(payload: unknown): payload is Record<string, unknown> {
+  return payload != null && typeof payload === 'object';
+}
+
+function isSendMousePayload(payload: unknown): payload is SendMousePayload {
+  const validTypes = ['move', 'click', 'down', 'up'];
+
+  if (!isObject(payload)) {
+    throw new Error('You must provide a `SendMousePayload` object');
+  }
+
+  if (typeof payload.type !== 'string' || !validTypes.includes(payload.type)) {
+    throw new Error(`You must provide a type option with one of the following values: ${validTypes.join(', ')}.`);
+  }
+
+  if (['click', 'move'].includes(payload.type)) {
+    if (!Array.isArray(payload.position) || typeof payload.position[0] !== 'number' || typeof payload.position[1] !== 'number') {
+      throw new Error('You must provide a position option as a numerical [x, y] tuple.');
+    }
+  }
+
+  return true;
+}
+
+export function sendMousePlugin(): TestRunnerPlugin<SendMousePayload> {
+  return {
+    name: 'send-mouse-command',
+    async executeCommand({ command, payload, session }): Promise<any> {
+      if (command === 'send-mouse') {
+        if (!payload || !isSendMousePayload(payload)) {
+          throw new Error('You must provide a `SendMousePayload` object');
+        }
+
+        // handle specific behavior for playwright
+        if (session.browser.type === 'playwright') {
+          const page = (session.browser as PlaywrightLauncher).getPage(session.id);
+          switch(payload.type) {
+            case 'move':
+              await page.mouse.move(payload.position[0], payload.position[1]);
+              return true;
+            case 'click':
+              await page.mouse.click(payload.position[0], payload.position[1]);
+              return true;
+            case 'down':
+              await page.mouse.down();
+              return true;
+            case 'up':
+              await page.mouse.up();
+              return true;
+          }
+        }
+
+        // handle specific behavior for puppeteer
+        if (session.browser.type === 'puppeteer') {
+          const page = (session.browser as ChromeLauncher).getPage(session.id);
+          switch(payload.type) {
+            case 'move':
+              await page.mouse.move(payload.position[0], payload.position[1]);
+              return true;
+            case 'click':
+              await page.mouse.click(payload.position[0], payload.position[1]);
+              return true;
+            case 'down':
+              await page.mouse.down();
+              return true;
+            case 'up':
+              await page.mouse.up();
+              return true;
+          }
+        }
+
+        // you might not be able to support all browser launchers
+        throw new Error(`Sending mouse is not supported for browser type ${session.browser.type}.`);
+      }
+    },
+  };
+}
--- a/packages/test-runner-commands/src/sendMousePlugin.ts
+++ b/packages/test-runner-commands/src/sendMousePlugin.ts
@@ -2,27 +2,30 @@ import { TestRunnerPlugin } from '@web/test-runner-core';
 import type { ChromeLauncher } from '@web/test-runner-chrome';
 import type { PlaywrightLauncher } from '@web/test-runner-playwright';
 
+type MousePosition = [number, number];
+type MouseButton = 'left' | 'middle' | 'right';
+
 export type SendMousePayload = MovePayload | ClickPayload | DownPayload | UpPayload;
 
 export type MovePayload = {
   type: 'move',
-  position: [number, number]
+  position: MousePosition
 };
 
 export type ClickPayload = {
   type: 'click',
-  position: [number, number],
-  // TODO: button: 'left' | 'right' | 'middle'
+  position: MousePosition,
+  button?: MouseButton
 }
 
 export type DownPayload = {
   type: 'down',
-  // TODO: button: 'left' | 'right' | 'middle'
+  button?: MouseButton
 }
 
 export type UpPayload = {
   type: 'up',
-  // TODO: button: 'left' | 'right' | 'middle'
+  button?: MouseButton
 }
 
 function isObject(payload: unknown): payload is Record<string, unknown> {
@@ -31,6 +34,7 @@ function isObject(payload: unknown): payload is Record<string, unknown> {
 
 function isSendMousePayload(payload: unknown): payload is SendMousePayload {
   const validTypes = ['move', 'click', 'down', 'up'];
+  const validButtons = ['left', 'middle', 'right'];
 
   if (!isObject(payload)) {
     throw new Error('You must provide a `SendMousePayload` object');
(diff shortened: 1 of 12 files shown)


packages/test-runner-commands/test/send-mouse/sendMousePlugin.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  it('can send mouse on puppeteer', async () => {
+    await runTests({
+      files: [path.join(__dirname, 'browser-test.js')],
+      browsers: [chromeLauncher()],
+      plugins: [sendMousePlugin()],
+    });
+  });
+
+  // playwright doesn't work on windows VM right now
+  if (platform() !== 'win32') {
```
web-padawan, line 36: You can remove this check, most of Playwright tests have been passing lately.

packages/test-runner-commands/test/send-mouse/sendMousePlugin.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import { chromeLauncher } from '@web/test-runner-chrome';
+import { webdriverLauncher } from '@web/test-runner-webdriver';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+import { sendMousePlugin } from '../../src/sendMousePlugin';
+import { startSeleniumServer } from '../selenium-server';
+
+let seleniumServer: selenium.ChildProcess;
+
+describe('sendMousePlugin', function test() {
+  this.timeout(20000);
```
web-padawan, line 14: Let's try the same timeout as `webdriver` launcher (and `selenium` launcher) tests.

```suggestion
  this.timeout(50000);
```

packages/test-runner-webdriver/src/IFrameManager.ts
```diff
@@ -155,6 +155,11 @@ export class IFrameManager {
     return this.driver.keys(keys);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async performActions(_: string, _actions: object[]) {
+    console.warn('Device actions are not supported in the concurrent mode.');
```
web-padawan, line 160: I would rather explicitly mention the necessary config option, e.g. something like this:

```
Unsupported operation. In order to use device actions in Webdriver, set "concurrency" to 1.
```

packages/test-runner-commands/test/send-mouse/sendMousePlugin.test.ts
vursen: Done.

packages/test-runner-commands/test/send-mouse/sendMousePlugin.test.ts
vursen: Done.

packages/test-runner-webdriver/src/IFrameManager.ts
vursen: Good point, done.

packages/test-runner-commands/src/sendMousePlugin.ts
```diff
@@ -0,0 +1,156 @@
...
+      `You must provide a type option with one of the following values: ${validTypes.join(', ')}.`,
+    );
+  }
+
+  if (['click', 'move'].includes(payload.type)) {
+    if (
+      !Array.isArray(payload.position) ||
+      typeof payload.position[0] !== 'number' ||
+      typeof payload.position[1] !== 'number' ||
+      !/^\d+$/.test(payload.position.join(''))
```
web-padawan, line 55: Consider using [`Number.isInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) for better readability.

packages/test-runner-webdriver/src/IFrameManager.ts
```diff
@@ -155,6 +155,13 @@ export class IFrameManager {
     return this.driver.keys(keys);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async performActions(_: string, _actions: object[]) {
+    console.warn(
+      'Unsupported operation. In order to use device actions in Webdriver, set "concurrency" to 1.',
+    );
```
web-padawan, line 162: I think `console.warn` isn't quite right here. Should we throw an error instead?

web-padawan (changes requested):
For some reason tests fail on Windows, need to investigate what goes wrong there 🤔 

```
  1) sendMousePlugin
       can send mouse on puppeteer:
     Error: Test run did not pass
      at EventEmitter.<anonymous> (packages\test-runner-core\src\test-helpers.ts:143:16)
      at EventEmitter.emit (domain.js:483:12)
      at TestRunner.emit (packages\test-runner-core\src\utils\EventEmitter.ts:24:20)
      at TestRunner.stop (packages\test-runner-core\src\runner\TestRunner.ts:178:10)

  2) sendMousePlugin
       can send mouse on playwright:
     Error: Test run did not pass
      at EventEmitter.<anonymous> (packages\test-runner-core\src\test-helpers.ts:143:16)
      at EventEmitter.emit (domain.js:483:12)
      at TestRunner.emit (packages\test-runner-core\src\utils\EventEmitter.ts:24:20)
      at TestRunner.stop (packages\test-runner-core\src\runner\TestRunner.ts:178:10)

  3) sendMousePlugin
       can send mouse on webdriver:
     Error: Test run did not pass
      at EventEmitter.<anonymous> (packages\test-runner-core\src\test-helpers.ts:143:16)
      at EventEmitter.emit (domain.js:483:12)
      at TestRunner.emit (packages\test-runner-core\src\utils\EventEmitter.ts:24:20)
      at TestRunner.stop (packages\test-runner-core\src\runner\TestRunner.ts:178:10)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Westbrook:
One thing we caught with this shape of API recently in Spectrum Web Components is that if you have multiple tests in the same file and you pointer down in one test the pointer won't be reset for the following test. To support that in the equivalent of `commands.mjs` we needed to add this: https://github.com/adobe/spectrum-web-components/blob/main/test/plugins/browser.ts We don't test the plugin, but it caused a failure in [this test](https://github.com/adobe/spectrum-web-components/blob/main/packages/slider/test/slider.test.ts#L488) because this [earlier test](https://github.com/adobe/spectrum-web-components/blob/d362050c2fd2c8b105bb32929923cfe69ddbb1b3/packages/slider/test/slider.test.ts#L334-L372) wasn't previously releasing the pointer.

packages/test-runner-commands/src/sendMousePlugin.ts
vursen: Done.

packages/test-runner-webdriver/src/IFrameManager.ts
vursen: Done.

vursen:
Yeah, that's not an obvious thing. I ran into that once while working on the plugin unit tests: it was not possible to `down` any mouse button until the previous one is not `up` and for that reason, I test `down` and `up` within one test case rather than within separate ones. Nevertheless, doesn't the current `sendKeys` API have the same problem?

```js
await sendKeys({ down: 'Shift' });
// ...
await sendKeys({ up: 'Shift' }); // <-------- should be explicitly called to release a key
```

I wonder if there is actually a way to get it around or make it somehow work more obvious for the user other than just mention it in the docs e.g:  "remember to release the mouse button after you down it"?

At least, we might consider auto-releasing the previously pressed mouse button once the user sends `down` for another button.

```
await sendMouse({ type: 'down' });
await sendMouse({ type: 'down' }); // <----- will release the left mouse button before making it `down` again.
```

...or maybe having a built-in wrapper around `sendMouse` with a queue like you shared would be even more preferable.

daKmoR:
how about a generic "reset function"?

e.g.  something like this

```js
await sendMouse({ type: 'down' });

// do some expects

await resetMouse(); // move to 0 0 (or another location?) and don't press any button

```

a similar approach could work for the keys as well?

then you can run it as a cleanup/setup as afterEach/beforeEach

web-padawan:
Personally, I like the idea about `resetMouse()` command - and IMO it's better to call it manually in `adterEach`.

Westbrook:
I like a reset method, but I don't like not automatically calling it in the after/teardown phase. Having run into the error, it takes forever to debug the issue there and we shouldn't be placing traps in our API for users to fall into like that.

daKmoR:
where/how would you add it? the `sendMouse` can not automatically add something to the afterEach? I mean we can add it to our docs and to our default setups but I don't think we can do more? at least I don't know of a way 🤔

vursen:
I'm not 100% sure if it is a good idea to reset the mouse (or the keyboard) on `afterEach`. Consider the following examples:

**sendMouse API:**
```js
describe('mouse', () => {
  before(async () => {
    await sendMouse({ type: 'move', position: [100, 100] }); 
  });

  it('should click once', () => {
    await sendMouse({ type: 'up' }); 
    await sendMouse({ type: 'down' }); 
  });

  it('should click once again', () => {
    // The mouse position will be reset to 0, 0 which is not obvious.
    await sendMouse({ type: 'up' }); 
    await sendMouse({ type: 'down' }); 
  });
});
```

**sendKeys API:**
```js
describe('keyboard', () => {
  before(async () => {
    await sendKeys({ down: 'Shift' });
  });

  after(async () => {
     await sendKeys({ up: 'Shift' }); 
  });

  it('should tab once', () => {
    await sendKeys({ press: 'Tab' });
  });

  it('should tab once again', () => {
    // The Shift key is not down anymore which is not obvious.
    await sendKeys({ press: 'Tab' });
  });
});
```

I don't see anything that would stop the user from doing this since it looks like a legal way to use the commands.

web-padawan:
I agree with @vursen and generally I would object enabling side-effects by default. Let's make sure we document the reset method. I would like to proceed with this PR once the build is green.

daKmoR:
I personally do not "like" these "global" setups... personally, I prefer each it as it's own individual unit... 

why I think it is bad

```js
describe('el', () => {
  let obj;
  before(async () => {
    obj = await setupObj(); // returns { foo: 'fooValue' }
  });

  it('first', () => {
    obj.bar = true; // we set this for something else we test
    expect(obj.foo).to.equal('fooValue');
  });
  
  it('second', () => {
    expect(obj.bar).to.be.true;
    // this now succeeds if run after "first" even though it should fail
  });
});
```


that is also how it's currently handled for fixture of  `@open-wc/testing`.

```js
import { fixture } from '@open-wc/testing';

describe('el', () => {
  let el;
  before(async () => {
    el = await fixture(html`<my-el></my-el>`);
  });

  it('first', () => {
    expect(el)...
  });
  // fixture get's clean up in afterEach

  it('second', () => {
    expect(el)... // this will error now
  });
});
```

imho if you set something up then it should be done in a `beforeEach` so every `it()` has a clean slate.

there are probably situations where sharing these "states" across `its` makes sense... but then it's the responsibility of the author to create and account for that.

In short
- we should document how to use `resetMouse` via `afterEach`
- if we add it to `@open-wc/testing` then we should add it to its "global" `afterEach`
- we should not apply any "magic" to automatically clean itself up
- (this recommends a good practice to use beforeEach & afterEach but does not prevent different uses like with shared states)

web-padawan:
@daKmoR Thank you for putting this together! I fully agree with your proposal, so we are on the same line here 💯

Westbrook:
@vursen the decision to write a test as you've supplied is a decision not to be making a unit test, as you are not cleanly separating your units as preparing yourself for any number of unknown issues from the bleed of test data from one test to the next.

What's more, if you were to add a test as follows to the end, it would fail/not test the interaction you were intending for not being able to complete the `click` action (at least in Playwright) as the mouse will already be _down_ and the internal API is to send a `down` then an `up` which means you'll only be able to register the `up` command: 

```
  it('should click once again', () => {
    // The mouse position will be reset to 0, 0 which is not obvious.
    await sendMouse({ type: 'click' }); 
  });
```

I'd very much say that we are similarly letting consumers down with the `sendKeys` API as it stands, but I'm not sure why it was easier to run into this issue with mouse in our actual usage.

> if we add it to @open-wc/testing then we should add it to its "global" afterEach

Why would we want our web component consumers to be better covered in this situation that our "web" consumers?

If at large the team is against cleaning up by default, I'd like to have clean up added optionally to the API signature. If we weren't batching the whole command at the options level, I'd propose:
```
await sendMouse({
   steps: [
        {type: 'click' }
    ],
    cleanup: true/false
); 
```
But seeing as we are, maybe a second argument?
```
const cleanup = true; /// false by default
await sendMouse({ type: 'click' }, cleanup); 
```

daKmoR:
`@web` does not have any testing helpers for specific frameworks so far right? 

so imho sendMouse should not be aware of the testing framework it is used on... that cleanup would fail if you use it with uvu or  any other testing system that is not using afterEach...

if you really need that then we could expose it as a wrapping function like `sendMouseBDDCleanup` or so... but it should only be a small section in the docs... most of the example should be about sendMouse

alternatively, we could introduce a `@web/testing` package and bundle it there... but I'm not sure if that make it even more confusing 🙈

Westbrook:
These are decent options. I do think we need to ship one from the start. 

The `sendMouseBDDCleanup` idea kind of implies a middle ground of a `bindSendMouseWithCleanup()` helper (needing lots of bikeshedding on naming) that you configure at the beginning of your tests that would allow for this to managed more cleanly for you. In that way `uvu` via `asdfg` might look like:

```
import { describe, executeTests } from '@asdgf/core';
import { expect } from '@open-wc/testing';
import { bindSendMouseWithCleanup } from '@web/test-runner-commands';

describe('button', ({it, beforeEach, afterEach }) => {
  const sendMouse = bindSendMouseWithCleanup(afterEach);
  let button;
  let count;

  beforeEach(() => {
    button = document.createElement('button');
    count = 0;
    button.addEventListener('click', () => count += 1);
    button.addEventListener('dragstart', () => count += 1);
  });

  it('drags', () => {
    await sendMouse({type: 'down'});
    await sendMouse({type: 'move', position: [100, 100]);
    expect(count)).to.equal(1);
  });

  it('clicks', () => {
    sendMouse({ type: 'click' });
    expect(count)).to.equal(1);
  });
});
```

This protects you both from the foot gun on not cleaning up at all, while also preventing multiple cleanups via a naive wrapping of the command locally that can overrun the test timeout, which is why you see the use of `mouseCleanupQueued` in the version I currently use: https://github.com/adobe/spectrum-web-components/blob/main/test/plugins/browser.ts#L57

web-padawan:
@Westbrook Please note that with this discussion we are probably setting the bar too high for a first time contributor.

Would it be an option to implement `sendMouse` first, and then add the `reset` / `cleanup` API in a subsequent PR?
Personally, I would name it `releaseMouse()` and then call it manually in `afterEach`. That's simple and clean.

BTW, we can move fast and break things, as this project is still at 0.x, so we can change the API any time 🙂

web-padawan (approved):
LGTM. 

Two things that IMO can be done in separate PRs:
1. Add documentation for the new command [here](https://github.com/modernweb-dev/web/blob/master/docs/docs/test-runner/commands.md),
2. Implement `resetMouse` as we previously discussed.

PR #1801, opened 2021-12-01
Update createPolyfillsLoader.ts

mrrasmussendk:
## What I did

1. I have changed the code from ES6 to a more native javascript approach mainly since these polyfills are used for older browsers who do not support arrow functions and declared parameters


LarsDenBakker:
There were multiple pull requests implementing the same thing. Sorry we were late in picking this up, but it's fixed on master now.

PR #1818, opened 2021-12-17
fix: support rollup plugins in manager.js

castastrophe:
## What I did

I noticed that running `build-storybook` was not reading in the rollup image plugin necessary for theme.js to load a brand image.  No issues with this in the `web-dev-server` command for running the local preview.  This adds processing for the rollup plugins for the manager assets as well.

### DIFF
--- a/packages/dev-server-storybook/src/build/build.ts
+++ b/packages/dev-server-storybook/src/build/build.ts
@@ -50,12 +50,16 @@ interface BuildmanagerParams {
 
 async function buildManager(params: BuildmanagerParams) {
   const managerHtml = createManagerHtml(params.storybookConfig, params.rootDir);
-  const config = createRollupConfig({
+  let config = createRollupConfig({
     type: params.type,
     outputDir: params.outputDir,
     indexFilename: 'index.html',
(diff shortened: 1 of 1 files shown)


packages/dev-server-storybook/src/build/build.ts
```diff
@@ -50,13 +50,17 @@ interface BuildmanagerParams {
...
   const managerHtml = createManagerHtml(params.storybookConfig, params.rootDir);
-  const config = createRollupConfig({
+  let config = createRollupConfig({
     type: params.type,
     outputDir: params.outputDir,
     indexFilename: 'index.html',
     indexHtmlString: managerHtml,
   });
 
+  if (storybookConfig.mainJs.rollupConfig) {
```
Westbrook, line 60:
```suggestion
  if (params.storybookConfig.mainJs.rollupConfig) {
```
?

LarsDenBakker:
I think it's better to introduce a new option for this, as most plugins should be used for the preview (that's where the user's own code is). I didn't know about theme.js, but I think it's a special case.

PR #1819, opened 2021-12-19
feat: add resetMouse command

vursen:
This PR adds a `resetMouse` command intended to be used after mouse manipulations by the `sendMouse` command to reset the mouse state and avoid potential side effects caused by the mouse being at a certain position or a mouse button being held down.

Here is an example of the command usage:

```js
it('does something with the mouse', () => {
  await sendMouse({ type: 'move', position: [150, 150] });
  await sendMouse({ type: 'down', button: 'middle' });
});

afterEach(() => {
  await resetMouse();
});
```

A follow-up on #1783

## What I did

1. Extended `sendMousePlugin` to support a `resetMouse` command.
2. Added unit tests.

### DIFF
--- a/.changeset/shy-singers-love.md
+++ b/.changeset/shy-singers-love.md
@@ -1,0 +1,6 @@
+---
+'@web/test-runner-commands': minor
+'@web/test-runner-webdriver': minor
+---
+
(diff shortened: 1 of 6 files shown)


web-padawan (approved):
LGTM. Once again, thanks for all the effort you put into this 🙏

PR #1821, opened 2021-12-22
fix(test-runner-core): evaluate files globs together

bennypowers:
## What I did

1. evaluate test runner `files` config as a glob array

PR #1824, opened 2021-12-23
[test-runner-junit-reporter] Add filepaths and line numbers to output

bennypowers:
## What I did

1. update the XML format with line numbers
2. add a test case that includes a source file (for stack traces)
3. fix tests to run properly
4. run all windows node tests against the production build

### DIFF
--- a/packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
+++ b/packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
@@ -1,30 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
-  <testsuite name="Chrome_puppeteer_/test/fixtures/multiple/nested-test.js" id="0" tests="1" skipped="0" errors="0" failures="0" time="0.001">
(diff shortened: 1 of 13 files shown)


packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
```diff
@@ -1,30 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
bennypowers, line 1: @mikepenz Would you mind reviewing these `expected.xml` files?

Westbrook (approved):
I’d love to be able to see a pre release of this to ensure it still works aOK with CircleCI, but not that big of a deal to make bug fixes. 

So, LGTM!!

bennypowers:
Maybe a patch would help?

Westbrook:
Any word on that test failure?

bennypowers:
ugh.. it's a windows thing

bennypowers:
ok @Westbrook at this point the test failures are the same as on #1821 , AFAICT unrelated

PR #1828, opened 2021-12-29
Fix MDX docs rendering using Storybook compiler and converting more imports to @web/storybook-prebuilt

Westbrook:
## What I did

Made MDX docs work again.

1. Removed our local compiler version in favor of the version for `cfs-tools` which seems to have removed much of the React dependencies that we previously listed needing our own copy for.
2. Pruned and updates some dependencies. (needs a stable release to `@web/storybook-prebuilt`)
3. Replaced one more version of blocks imports with prebuilt imports.

This still needs a number of changes to `storybook-addon-markdown-docs` to dispel security warnings, but good projects.

cc: @abdonrd


abdonrd (approved):
Looks good! Thanks @Westbrook! 👏🏻

PR #1831, opened 2022-01-03
fix(dev-server-storybook): rename storybook config file

samvimes01:
## Why I did this change
I had a bug and couldn't start storybook in my package with type: module

Steps to reproduce:
1. start a package with packcage.json `"type: module"` (or for example clone or download https://github.com/lit/lit-element-starter-ts)
2. according to the current [guide](https://modern-web.dev/guides/dev-server/storybook/) add the dev-server-storybook
  - create a `.storybook/main.js` with `exports.default = ` (N.B. - it's the source of issue)
  - create a `.storybook/server.config.mjs`
3. run `wds --config .storybook/server.config.mjs`
4. got the error `ERR_REQUIRE_ESM` `require() of ES modules is not supported`

## What I did
So the fastest and easiest fix that works both in commonjs and module packages - rename main.js to main.cjs
1. renamed hardcoded filename in `packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts`
2. changed docs and guide - rename main.js to main.cjs

### DIFF
--- a/.changeset/metal-teachers-lay.md
+++ b/.changeset/metal-teachers-lay.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server-storybook': minor
+---
+
+rename .storybook/main.js to .storybook/main.cjs to make it work in packages with "type": "module"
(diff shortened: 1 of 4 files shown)


LarsDenBakker:
This is a breaking change that will break existing projects. It would be better to check if the .cjs file exists, and otherwise use the .js file.

PR #1834, opened 2022-01-04
feat(test-runner): summaryReporter

bennypowers:
## What I did

1. add `summaryReporter`
2. add `dotReporter`

These reporters reuse code from defaultReporter

Example output:

with `flatten: false` (default)

Screen Shot 2022-01-04 at 19 07 49

with `flatten: true`

Screen Shot 2022-01-04 at 19 07 19

dot reporter:
Screen Shot 2022-01-05 at 14 43 35


bennypowers:
oops

Westbrook (commented):
Does the info in the changeset belong in the docs somewhere?

Otherwise, looking good to me. All net new means "feat"!

PR #1844, opened 2022-01-10
fix: check isTTY to prevent node from crashing/exiting early

rquast:
## Why I did this change

- Husky was running WTR but would exit early with error level 0 before running any tests.
- Found out that process.stdin.isTTY was undefined when running with husky since it runs in a non-tty environment.
- Noticed that after doing anything with process.stdin, node would crash or exit (no error report or notice). 
- Also experienced the same problem when setting up CI in TeamCity.
 
## What I did

- add a check for isTTY to be true so that non-tty environments do not crash or exit the process with error level 0.

### DIFF
--- a/.changeset/check-tty-stop-early-exit.md
+++ b/.changeset/check-tty-stop-early-exit.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-core': patch
+---
+
+Check isTTY to prevent node from crashing/exiting early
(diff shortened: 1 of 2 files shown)


rquast:
We've recently converted our Vite/Jest/RTL tests to Vite/WTR/RTL and everything works great! However, I found a few issues along the way to get things running nicely. The first problem is I noticed that husky wasn't running the tests as WTR would exit with error level 0. After lots of digging around through the code, I found out that process.stdin.isTTY was undefined in husky since it runs in a non-tty environment. I noticed that after doing anythign with process.stdin, node would crash or exit (no error report or notice). We also experienced the same problem when setting up the CI in teamcity - the TC runner has no tty and the tests wouldn't run and the process exits with error level 0.

LarsDenBakker:
Thanks, and sorry for the delay.

PR #1859, opened 2022-01-26
feat: update selenium-webdriver dependency to 4.3.0

web-padawan:
## What I did

1. Updated `selenium-webdriver` dependency to 4.1.0


web-padawan (commented):
Rebased, let's see if this passes after updating the Windows image.

packages/test-runner-selenium/test/seleniumLauncher.test.ts
```diff
@@ -11,7 +11,7 @@ async function startSeleniumServer() {
   try {
     await selenium.install({
       drivers: {
-        chrome: { version: 'latest' },
+        chrome: { version: '103.0.5060.134' },
```
Westbrook, line 14: How will we tend this version going forward? Everything else looks good to go, but this feels trouble inducing.

web-padawan: This is a recurring problem caused by the fact that ChromeDriver version installed on GitHub Actions is not the latest one. See for example #1798 where it was pinned previously.

Westbrook: But, what will _we_ do when the next version comes out? Is there a cronjob approach to maintaining this, or do we just wait for tests to start failing, or do we leave it like this and just let it get further out of date over time?

web-padawan: We probably need to detect which Chrome version is installed and then pass it to `selenium-standalone`.

Westbrook: Is that something we should do here? Something we should make an issue about? Etc. 

I’m all for moving forward with this if it’s the right thing, just wanna make sure we have the side effects gathered clearly for the next time.

web-padawan: @Westbrook I tried to use `chrome-launcher` to detect Chrome binary and then run it with `--version`. This works on MacOS and Linux (tests passed in CI). Unfortunately, it's not supported on Windows, see [Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=158372).

web-padawan: The new images for GitHub Actions have been deployed with Chrome 104 support:

- [Ubuntu 20.04](https://github.com/actions/runner-images/releases/tag/ubuntu20%2F20220807.1)
- [Windows server 2022](https://github.com/actions/runner-images/releases/tag/win22%2F20220808.1)

Rebased the PR, let's see if `latest` works now. If so, we can merge the PR.
I will create a separate issue to evaluate how to detect Chrome version.

web-padawan: Windows build fails, apparently because Windows image is still a pre-release 😕

Issue #1521, opened 2021-06-14
Make snapshot directory configurable


LarsDenBakker:
We currently store snapshots next to the test file itself. This doesn't work out when you are using a compiler and output to a separate directory. We should make the snapshot directory configurable.

alfonsogarciacaro:
This would be very helpful, currently I have to set up exceptions in .gitignore to check only the snapshots folder, which is a bit cumbersome.

JakobVogelsang:
Are there any updates on this issue? We are moving from karma to web-test-runner and this is the still an inconvenience.

lovenm:
I opened a PR about this last week: https://github.com/modernweb-dev/web/pull/1866

I'm not sure if there's some protocol I need to follow in order to get someone to give it a look-see, or if everyone's just busy with other things for now.

lovenm:
@LarsDenBakker any thoughts on the PR? The dom snapshots currently feel a bit awkward on Typescript due to this. I was not entirely sure if the way I implemented it in the PR was an optimal way, but I can also work more on this if there are some better ideas.

PR #1866, opened 2022-01-29
feat(test-runner-commands): make snapshot file path configurable

lovenm:
## What I did

1. Added configuration option to control where snapshot files are saved.

I was not totally sure if the configuration of the plugin happen somewhere other than `TestRunnerConfig`, but  could not figure a better way to configure it since the plugin seems to be added by default when the runner is started. If I understood correctly, if a user would add the `snapshotPlugin(...)` in their own configuration file, that would not override the default one, right?

I've not updated the docs yet, but can do that if the code changes otherwise look ok.

### DIFF
--- a/packages/test-runner-commands/src/snapshotPlugin.ts
+++ b/packages/test-runner-commands/src/snapshotPlugin.ts
@@ -36,21 +36,27 @@ function isSaveSnapshotPayload(payload: unknown): payload is SaveSnapshotPayload
   return true;
 }
 
-function getSnapshotPath(testFile: string) {
-  const testDir = path.dirname(testFile);
-  const testFileName = path.basename(testFile);
-  const ext = path.extname(testFileName);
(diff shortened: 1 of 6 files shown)


ca-d:
@koddsson @bashmish would you please comment on whether the behaviour described in #1521 is intended or, if not, what (apart from a changeset) would be needed to accept this pull request? Is the missing changeset only reason this issue and pull request don't see any activity?

Recap of the situation: We currently have to commit an empty `dist` directory with only a `__snapshots__` subdirectory to our repository just because we cannot configure the snapshot directory ever since moving away from karma.

lovenm:
Honestly it's been such a long time since I submitted this, that I'm a bit out of the loop regarding the codebase and have no idea how relevant the changes are anymore. The conflicts don't look too outlandish I suppose.

I've also since left the project where we were struggling with this issue, but can maybe give this a look at if this is deemed relevant in some way.

PR #1879, opened 2022-02-14
feat: add support for configuring the failure threshold

WickyNilliams:
## What I did

1. add config option for setting a failure threshold by percent or number of pixels
2. check this value when diffing the image, and return appropriate message
3. improve console output when image size mismatch

### DIFF
--- a/packages/test-runner-visual-regression/src/config.ts
+++ b/packages/test-runner-visual-regression/src/config.ts
@@ -67,6 +67,7 @@ export interface VisualRegressionPluginOptions {
    * For `failureThresholdType` of "pixels", this should be a positive integer.
    */
   failureThreshold: number;
+
   /**
    * The type of threshold that would trigger a failure.
    */
--- a/packages/test-runner-visual-regression/src/config.ts
+++ b/packages/test-runner-visual-regression/src/config.ts
@@ -28,6 +28,7 @@ export type OptionalImage = Buffer | undefined | Promise<Buffer | undefined>;
 
 export interface DiffResult {
   diffPercentage: number;
+  diffPixels: number;
   diffImage: Buffer;
   error: string;
 }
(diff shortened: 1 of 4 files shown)


packages/test-runner-visual-regression/src/visualDiffCommand.ts
```diff
@@ -79,22 +103,28 @@ export async function visualDiffCommand(
...
   });
 
+  const { error, diffImage } = result;
+
   if (error) {
     // The diff has failed, be sure to save the new image.
     await saveFailed();
     await saveDiff();
 
-    throw new VisualRegressionError(error);
```
WickyNilliams, line 94: Throwing an error here, rather than relying on the usual returned object, created lots of noise in the console log since you get a stack trace spat out separately from the usual test failure output: 

```
VisualRegressionError: Screenshot is not the same width and height as the baseline. Baseline: { width: 752, height: 152 } Screenshot: { width: 752, height: 155 }
    at Object.visualDiffCommand (/Users/blah/modern-web/packages/test-runner-visual-regression/dist/visualDiffCommand.js:89:15)
    at async TestRunnerApiPlugin._onCommand (/Users/blah/design-system/packages/components/node_modules/@web/test-runner-core/dist/server/plugins/api/testRunnerApiPlugin.js:145:32)
```

So I made a slight tweak to use the regular mechanism for reporting failure. Now the output is cleaner and more consistent.

Hope that's ok? Couldn't see any real reason why an error is thrown here for failure, but every other failure case returns an object. Maybe i'm missing something!

packages/test-runner-visual-regression/src/config.ts
```diff
@@ -60,6 +61,17 @@ export interface VisualRegressionPluginOptions {
    */
   diffOptions: PixelMatchOptions;
 
+  /**
+   * The threshold after which a diff is considered a failure, depending on the failureThresholdType.
+   * For `failureThresholdType` of "percentage", this should be a number between 0-100.
+   * For `failureThresholdType` of "pixels", this should be a positive integer.
+   */
+  failureThreshold: number;
```
web-padawan, line 69:
```suggestion
  failureThreshold: number;

```

web-padawan:
Sorry about leaving this hanging. Rebased the PR to trigger the build, I think we can merge when it's green.

WickyNilliams:
No worries! I'd completely forgotten about it until my colleague asked about one of these features :)

web-padawan (approved):

peschee:
@web-padawan @WickyNilliams This PR seems to be missing the updated config parameters in the main readme: https://github.com/modernweb-dev/web/tree/master/packages/test-runner-visual-regression

web-padawan:
Thanks, I missed to check those while reviewing. Feel free to submit another PR to add those.

PR #1881, opened 2022-02-21
fix: use latest "storybook-addon-markdown-docs"

Westbrook:
## What I did

1. Hopefully this cleans up some of the dependency vulnerabilities, but it's also the first step of a larger (read breaking) change to `storybook-addon-markdown-docs` that will convert it to ESM and upgrade a number of similarly converted/breaking dependencies.

### DIFF
--- a/.changeset/tough-cups-report.md
+++ b/.changeset/tough-cups-report.md
@@ -1,0 +1,6 @@
+---
+"@web/dev-server-storybook": patch
+"@web/dev-server": patch
+---
+
+Use latest "storybook-addon-markdown-docs"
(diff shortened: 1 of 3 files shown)


web-padawan (commented):
Updated `yarn.lock` using the [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate) to keep the diff smaller.

Does this need a changeset? Probably not, as it's a patch release bump.

Westbrook:
Added the change set so that the version gets pushed to current consumers via dependency linters like dependabot. Theres a pretty nice bug fix for `mdjs` users in this, so getting it sooner will be better for everyone.

web-padawan (approved):

abdonrd (approved):
Thanks! 👏

PR #1882, opened 2022-02-21
chore: use latest browser drivers and Windows 2022 image

web-padawan:
## What I did

1. Removed hardcoded version which no longer works on GitHub Actions
2. Updated Windows image to `windows-2022` (the one that works)

### DIFF
--- a/packages/test-runner-selenium/test/seleniumLauncher.test.ts
+++ b/packages/test-runner-selenium/test/seleniumLauncher.test.ts
@@ -11,8 +11,8 @@ async function startSeleniumServer() {
   try {
     await selenium.install({
       drivers: {
-        chrome: { version: '96.0.4664.18' },
-        firefox: { version: '0.30.0' },
+        chrome: { version: 'latest' },
+        firefox: { version: 'latest' },
       },
     });
   } catch (err) {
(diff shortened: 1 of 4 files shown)


packages/test-runner-selenium/test/seleniumLauncher.test.ts
```diff
@@ -11,8 +11,8 @@ async function startSeleniumServer() {
   try {
     await selenium.install({
       drivers: {
-        chrome: { version: '96.0.4664.18' },
-        firefox: { version: '0.30.0' },
+        chrome: { version: 'latest' },
+        firefox: { version: 'latest' },
```
Westbrook, line 15: Would this be crazy instead: 
```
  try {
    await selenium.install({
      drivers: {
        chrome: { version: 'latest' },
        firefox: { version: 'latest' },
      },
    });
  } catch (err) {
    try {
      await selenium.install({
        drivers: {
          chrome: { version: '97.0.4692.99' },
          firefox: { version: 'latest' },
        },
      });
    } catch (err) {
      console.error('Error occurred when installing selenium.');
      throw err;
    }
  }
```
Maybe we could parse the `err` for a more nuanced second try?

web-padawan: I'd like to avoid hardcoding. Build passed now, let's reconsider this if the problem happens again.

Westbrook (approved):
LGTM! Thanks for diving into this so quickly. 🙇🏼

PR #1889, opened 2022-02-24
feat(test-runner-core): allow http2/ssl opts in config

emensch:
## What I did

1. Passed http2/ssl configuration options from WTR into the underlying DevServer instance to allow tests served over HTTPS

### DIFF
--- a/.changeset/unlucky-radios-peel.md
+++ b/.changeset/unlucky-radios-peel.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-core': patch
+---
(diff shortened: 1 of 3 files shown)


Westbrook (approved):
LGTM! Thanks @emensch 🙏🏼

LarsDenBakker:
@emensch can you make the change a patch? We are still in 0.x so a minor would be a breaking change.

emensch:
Updated @LarsDenBakker - thanks!

PR #1891, opened 2022-02-26
Update Web Dev Server documentation suggestions

Matsuuu:
Update suggestions on setting up Web Dev Server by suggesting the use of --watch flag.

I feel like this flag is as important as the currently present suggested flags and therefore should be included


daKmoR:
thxxx 🙇‍♂️

PR #1913, opened 2022-03-31
Relax esbuild semver dependency

peschee:
## What I did

1. Changed the semver from `^0.12.21` to `^0.12 || ^0.13 || ^0.14`.

I'm not sure how to check for all possible incompatibilities, but all tests still run when checking with the latest `0.12.x`, `0.13.x` and `0.14.x` version.

### DIFF
--- a/packages/dev-server-esbuild/package.json
+++ b/packages/dev-server-esbuild/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mdn/browser-compat-data": "^4.0.0",
     "@web/dev-server-core": "^0.3.17",
-    "esbuild": "^0.12.21",
+    "esbuild": "^0.12|^0.13|^0.14",
     "parse5": "^6.0.1",
     "ua-parser-js": "^1.0.2"
   },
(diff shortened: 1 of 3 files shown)


packages/dev-server-esbuild/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mdn/browser-compat-data": "^4.0.0",
     "@web/dev-server-core": "^0.3.17",
-    "esbuild": "^0.12.21",
+    "esbuild": "^0.12|^0.13|^0.14",
```
Westbrook, line 55: Is `0.14` “latest”? Would we benefit from `>0.12` to simplify in the future?

peschee: yes, `0.14` is latest. having the `|` in there with 12, 13 and 14 is more restrictive than `>0.12`. I can live with both.

abdonrd, line 55: I think this should be like this:

```suggestion
    "esbuild": "^0.12 || ^0.13 || ^0.14",
```

abdonrd: But I this works better:

```suggestion
    "esbuild": ">=0.12",
```

(but maybe we should update it in a future breaking changes)

peschee: Yes ofc it should be `||`… I have changed that.

I can also change it to `>=0.12` if everybody is ok with that.

Issue #1921, opened 2022-04-02
[@web/dev-server-esbuild] does not read tsconfig.json


pmcelhaney:
As far as I can tell, the ESBuild plugin does not look at tsconfig.json.

It uses the ESBuild's Transform API, which would require the plugin to explicitly pass the contents of the `tsconfig.json` file via the [tsconfigRaw](https://esbuild.github.io/api/#tsconfig-raw) option. 

So far not reading the tsconfig.json hasn't presented any significant problems for me, but the fact that my changes to that file had no effect on web-dev-server was confusing. I would suggest either documenting the fact that the tsconfig.json file is ignored or adding an option to specify the path to the tsconfig.json file. 

If either of those sounds good, or you have some other idea, I'm happy to create a PR.

LarsDenBakker:
Adding an option to specify the path would be a good idea.

kobleistvan:
I agree. We would need this to work, because we're using this within a monorepo, and specifying path aliases would be very useful. Thanks

pmcelhaney:
@kobleistvan Would it be helpful if there was an option to automatically find the tsconfig file for each .ts file (using the same algorithm as TypeScript itself)? 

That's my ultimate goal. I started with specifying the path to a single tsconfig.json because I didn't want to do too much in one PR (and there are cases when I want to point to a specific config).

kobleistvan:
@pmcelhaney In out application, we have a 'base' tsconfig in the monorepo root, and some packages have their own tsconfig file which extends the one from the root. So it would be great if perhaps the wtr command itself could take as an additional argument the path to a specific tsconfig file (and take into consideration the internal 'extends' key's value).

kobleistvan:
@pmcelhaney FYI, I worked around my issue via:
`import { fromRollup } from '@web/dev-server-rollup';
import tsConfigPaths from 'rollup-plugin-tsconfig-paths';

const tsPaths = fromRollup(tsConfigPaths.default);

// ...

  plugins: [
    tsPaths({}), 
    esbuildPlugin({ ts: true }),
  ],
`

So basically there's a rollup plugin to overcome the path issues, and the fromRollup makes it so that I can use it here. Would be nice if there would be native support though (i.e. for other tsConfig configs)

Cheers

lucaelin:
> So far not reading the tsconfig.json hasn't presented any significant problems for me

[...]

Issue #1800, opened 2021-12-01
Polyfill loader injects modern script


mrrasmussendk:
Looking at [https://modern-web.dev/docs/building/rollup-plugin-polyfills-loader/](https://modern-web.dev/docs/building/rollup-plugin-polyfills-loader/)

the output of the polyfill loader is "modern" javascript and can therefore not be executed on IE11

Example of rollup build:

```
// Import rollup plugins
import html from '@web/rollup-plugin-html';
import polyfillsLoader from '@web/rollup-plugin-polyfills-loader';
import {copy} from '@web/rollup-plugin-copy';
import resolve from '@rollup/plugin-node-resolve';
import {getBabelOutputPlugin} from '@rollup/plugin-babel';
import {terser} from 'rollup-plugin-terser';
import minifyHTML from 'rollup-plugin-minify-html-literals';
import summary from 'rollup-plugin-summary';

// Configure an instance of @web/rollup-plugin-html
const htmlPlugin = html({
  rootDir: './',
  flattenOutput: false,
});

export default {
  // Entry point for application build; can specify a glob to build multiple
  // HTML files for non-SPA app
  input: 'index.html',
  plugins: [
    htmlPlugin,
    // Resolve bare module specifiers to relative paths
    resolve(),
    // Minify HTML template literals
    minifyHTML(),
    // Minify JS
    terser({
      module: true,
      warnings: true,
    }),
    // Inject polyfills into HTML (core-js, regnerator-runtime, webcoponents,
    // lit/polyfill-support) and dynamically loads modern vs. legacy builds
    polyfillsLoader({
      modernOutput: {
        name: 'modern',
      },
      // Feature detection for loading legacy bundles
      legacyOutput: {
        name: 'legacy',
        test: "!('noModule' in HTMLScriptElement.prototype)",
        type: 'systemjs',
      },
      // List of polyfills to inject (each has individual feature detection)
      polyfills: {
        webcomponents: false,
        coreJs: true,
        fetch: true,
        // Custom configuration for loading Lit's polyfill-support module,
        // required for interfacing with the webcomponents polyfills
        custom: [
          {
            name: 'webcomponent-loader',
            path: 'node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js',
            module: false,
          },
          {
            name: 'lit-polyfill-support',
            path: 'node_modules/lit/polyfill-support.js',
            test: "!('attachShadow' in Element.prototype)",
            module: false,
          },
        ],
      },
[...]

PR #1922, opened 2022-04-04
dev-server-esbuild: option to specify path to tsconfig file

pmcelhaney:
## What I did

1. Added an option to specify the path to the tsconfig.json file (#1921)
2. Added a unit test 

## To do

- [x] fix the unit test 
- [ ] add more tests?
- [x] update the docs

### DIFF
--- a/docs/docs/dev-server/plugins/esbuild.md
+++ b/docs/docs/dev-server/plugins/esbuild.md
@@ -130,7 +130,10 @@ Transform all .ts files to javascript using settings from tsconfig.json. (The `t
 ```js
 import { fileURLToPath } from 'url';
 
-esbuildPlugin({ ts: true, tsconfig: fileURLToPath(new URL('./tsconfig.json', import.meta.url)) });
+esbuildPlugin({
+  ts: true,
+  tsconfig: fileURLToPath(new URL('./tsconfig.json', import.meta.url)),
+});
 ```
 
 **JSX:**
--- a/docs/docs/dev-server/plugins/esbuild.md
+++ b/docs/docs/dev-server/plugins/esbuild.md
@@ -45,6 +45,7 @@ type Loader =
 
 interface EsbuildPluginArgs {
   target?: string | string[];
+  js?: boolean;
   ts?: boolean;
   json?: boolean;
   jsx?: boolean;
@@ -52,8 +53,8 @@ interface EsbuildPluginArgs {
   jsxFactory?: string;
   jsxFragment?: string;
   loaders?: Record<string, Loader>;
-  strict?: boolean | Strict[];
   define?: { [key: string]: string };
+  tsconfig?: string;
 }
 ```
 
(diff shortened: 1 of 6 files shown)


packages/dev-server-esbuild/test/ts.test.ts
```diff
@@ -243,4 +243,31 @@ class Bar {
...
+    try {
+      const response = await fetch(`${host}/a/b/foo.ts`);
+      const text = await response.text();
+
+      expect(response.status).to.equal(200);
+      expect(response.headers.get('content-type')).to.equal(
+        'application/javascript; charset=utf-8',
+      );
+
+      expectIncludes(text, '// a comment');
```
pmcelhaney, line 268: Because of `"removeComments": false` in the tsconfig.json I expect that comment remain in the output. The plugin is reading the file and passing the contents to ESBuild, but ESBuild doesn't seem to acknowledge. 

I'm probably missing something silly.

pmcelhaney:
Updated the test to work based on`useDefineForClassFields` instead of `removeComments` and now it's passing. It looks like ESBuild respects only a subset of compiler options when passed to `tsconfigRaw`.

packages/dev-server-esbuild/test/ts.test.ts
pmcelhaney: Fixed https://github.com/modernweb-dev/web/pull/1922#issuecomment-1087835781

packages/dev-server-esbuild/src/EsbuildPlugin.ts
```diff
@@ -164,17 +166,23 @@ export class EsbuildPlugin implements Plugin {
     target: string | string[],
   ): Promise<string> {
     try {
-      const { code: transformedCode, warnings } = await transform(code, {
+
+      const tsconfigRaw = this.esbuildConfig.tsconfig ? await promisify(fs.readFile)(this.esbuildConfig.tsconfig, 'utf8') : undefined;
```
LarsDenBakker, line 170: This will read the config each transform. Let's:
- Do promisify(fs.readFile) only once on top
- Read the ts config in server start

docs/docs/dev-server/plugins/esbuild.md
```diff
@@ -45,15 +45,16 @@ type Loader =
...
   ts?: boolean;
   json?: boolean;
   jsx?: boolean;
   tsx?: boolean;
   jsxFactory?: string;
   jsxFragment?: string;
   loaders?: Record<string, Loader>;
-  strict?: boolean | Strict[];
   define?: { [key: string]: string };
+  tsconfig?: string;
```
pmcelhaney, line 57: I copy and pasted from the source code. I guess `strict` was removed and `js` was added since this doc was updated.

LarsDenBakker:
There are some linting errors

LarsDenBakker:
You can run yarn format to fix the linting/formatting errors

pmcelhaney:
A snapshot is failing now but it's unrelated to my change. Looks like it has something to do with #1800

pmcelhaney:
Rebased and now the tests are passing.

PR #1920, opened 2022-04-02
Fix/node 17 hash compatibility

maartenst:
## What I did

1.Change the hashing for rollup-plugin-html from md4 to md5 to be compatible with Node 17 and up.
2.Add 'yarn build' to the Contributing file setup stage to avoid it being missed before trying to run tests

PR #1926, opened 2022-04-08
Feat/legacy polyfills loader

LarsDenBakker:
## What I did

It turns out that @web/dev-server-legacy is still using the old open-wc polyfills loader. This updates it to use @web/polyfills-loader. 

Needed for https://github.com/modernweb-dev/web/pull/1920


bicknellr:
Oh, nice! I was just trying to use the legacy plugin with `polyfills: false` but couldn't figure out why it wasn't working. Sounds like this will fix it.

PR #1930, opened 2022-04-09
fix / finish the demo app for esbuild plugin

pmcelhaney:
## What I did

I noticed the demo app for the ESBuild plugin was not finished. It had a syntax error, the config file had the wrong name, etc.


LarsDenBakker:
Thanks!

PR #1949, opened 2022-05-01
docs: add file globbing example to serverStart

bennypowers:
## What I did

1. add an example of how to add multiple files to the dev-server file watcher with `glob`

### DIFF
--- a/docs/docs/dev-server/writing-plugins/hooks.md
+++ b/docs/docs/dev-server/writing-plugins/hooks.md
@@ -332,6 +332,10 @@ The `serverStart` hook is called when the server starts. It is the ideal locatio
 Accessing the serverStart parameters:
 
 ```js
+import _glob from 'glob';
+import { promisify } from 'util';
+const glob = promisify(_glob);
+  
 function myFancyPlugin() {
   let rootDir;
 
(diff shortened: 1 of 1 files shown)


docs/docs/dev-server/writing-plugins/hooks.md
```diff
@@ -332,6 +332,10 @@ The `serverStart` hook is called when the server starts. It is the ideal locatio
 Accessing the serverStart parameters:
 
 ```js
+import _glob from 'glob';
```
Westbrook, line 335: FYI: https://www.npmjs.com/package/fast-glob does this without the promise wrapping, not sure if it much the same otherwise or not.

LGTM, otherwise.

Westbrook (approved):

docs/docs/dev-server/writing-plugins/hooks.md
bennypowers: yeah, `globby` as well

I figured `glob` is a 'lowest-common-denominator' kind of package

Westbrook: If wrapping is fine, seems fine. I’ve never wrapped like that to say if it feels “normal” or not.

PR #1959, opened 2022-05-11
Test runner logger and reporters modules

edoardocavazza:
This PR introduces two dedicated modules for test runner logger and test runner reporters. As discussed [here](https://github.com/modernweb-dev/web/discussions/1958), providing a dedicated module for reporters utils (such as the progress bar) is helpful plugins author.

## What I did

1. Created new package for `test-runner-logger`
2. Moved logger sources into the new module
3. Updated references in `test-runner` package
4. Created new package for `test-runner-reporters`
5. Moved reporters sources into the new module
7. Updated references in `test-runner` package

## Notes

* The `test-runner` package re-exports reporters for backward compatibility


bennypowers:
Some questions for review:
Should these be in
1. separate packages (as per initial pr),
2. a single `test-runner-helpers` package, or
3. a package export from `test-runner`?

edoardocavazza:
IMO, using the `test-runner` package as dependency is not preferable because it downloads and loads a lot of stuff for cli and the whole runner usage. May I add option 4: move logger and reporters to `test-runner-core`?

PR #1961, opened 2022-05-13
Implement SWC plugin

koddsson:
## What I did

1. I copied the esbuild plugin directory to a new directory `dev-server-swc`.
2. I then updated the implementation to use the SWC compiler instead of the esbuild one.
3. Then I opened up this PR!


packages/dev-server-swc/demo/ts/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
bennypowers, line 1: is this necessary for the demo? the root gitignore should cover this

packages/dev-server-swc/index.mjs
```diff
@@ -0,0 +1,6 @@
+// this file is autogenerated with the generate-mjs-dts-entrypoints script
```
bennypowers, line 1: if this is autogenerated, it should be gitignored

packages/dev-server-swc/package.json
```diff
@@ -0,0 +1,66 @@
...
+  "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/dev-server-swc",
+  "main": "dist/index.js",
+  "exports": {
+    ".": {
+      "import": "./index.mjs",
+      "require": "./dist/index.js"
+    }
+  },
+  "engines": {
+    "node": ">=10.0.0"
```
bennypowers, line 24: we've dropped support for node 12, so ideally this should be >=14.0.0
```suggestion
    "node": ">=14.0.0"
```

packages/dev-server-swc/test/target.test.ts
```diff
@@ -0,0 +1,388 @@
+import { expect } from 'chai';
+import fetch from 'node-fetch';
+import { createTestServer, expectIncludes } from '@web/dev-server-core/test-helpers';
+
+import { swcPlugin } from '../src/index';
+
+const modernJs = `
```
bennypowers, line 7: thanks for including these cases

packages/dev-server-swc/demo/ts/.gitignore
koddsson: The root gitignore doesn't seem to cover this.

```
koddsson@Kristjans-MBP dev-server-swc % git st
## swc-plugin
 D demo/ts/.gitignore
?? demo/ts/node_modules/
```

Note that I copied this over from `dev-server-esbuild` so the demo there also has this `.gitignore` file. 

https://github.com/modernweb-dev/web/blob/master/packages/dev-server-esbuild/demo/ts/.gitignore

packages/dev-server-swc/index.mjs
koddsson: I copied this over from the `dev-server-esbuild` plugin which doesn't seemed to be gitignored either. In fact all the `index.mjs` files are not gitignored.

packages/dev-server-swc/package.json
```diff
@@ -21,7 +21,7 @@
     }
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=12.0.0"
```
bennypowers, line 24: why?

koddsson: To match all other packages in the repo. I was worried that the tests were failing because of the change but I can totally revert this.

dgreif:
@bennypowers is there anything else specific that needs to be addressed before this could be merged?  Really excited to use SWC + web-test-runner 😁

PR #1977, opened 2022-05-30
Remove unused lit-html for deduplication

PonomareVlad:
Not compatible with new Lit

### DIFF
--- a/packages/dev-server/package.json
+++ b/packages/dev-server/package.json
@@ -72,7 +72,6 @@
   },
   "devDependencies": {
     "@types/command-line-usage": "^5.0.1",
-    "lit-html": "^2.0.0",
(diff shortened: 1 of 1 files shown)


web-padawan:
Please note that the lit-html dev dependency is used in the demo:

https://github.com/modernweb-dev/web/blob/72c070ebab35131e391695e4f80921d7c0a9c730/packages/dev-server/demo/node-resolve/extension-priority.js#L1

Also, the specified range is semver compatible with the latest `2.2.5` at the time of writing.
And there are no changes in `yarn.lock` after running `yarn` with this PR, so deduplication works.

Closing this PR, feel free to comment in case if there is something that I misunderstood.

PR #1978, opened 2022-05-31
Whether or

ryanauj:
## What I did

1. Fixed what I think is a typo. I think this sentence should should read `whether or` instead of `whether and`.


Westbrook (approved):
Thanks for the corrections!!

PR #1984, opened 2022-06-07
fix(rollup-plugin-import-meta-assets): completely overwrite import.meta.url

cprecioso:
This removes the duplicated `import.meta.url` in the output, like this:

```diff
 // Pre-bundling code
 new URL("./some-internal-path.png", import.meta.url)
 
 // Bundled code in ESM output 
-new URL(new URL("asset/bundled-asset.png", import.meta.url).href, import.meta.url) 
+new URL(new URL("asset/bundled-asset.png", import.meta.url).href) 

 // Bundled code in CJS output 
 new URL(
   "undefined" == typeof document
     ? new (require("url").URL)(
         "file:" + __dirname + "/assets/InterLatin-b0e9bc2a.woff2"
       ).href
     : new URL(
         "assets/InterLatin-b0e9bc2a.woff2",
         (document.currentScript && document.currentScript.src) ||
           document.baseURI
       ).href
-, import.meta.url); // Uncaught SyntaxError: Cannot use 'import.meta' outside a module
+);
```

## Why

This way, it should work exactly the same (Rollup makes sure we have `import.meta.url` or something similar to base our URL on); but the second `import.meta` gets eliminated, which reduces size, repetition, and most of all, doesn't trigger an error in Common JS modules.

## What I did

1. Updated the overwrite function

   ```js
   // Before
   new URL("myFile.tx", import.meta.url)
   //      |---------|
   //      Only this part was overwritten

   // After
   new URL("myFile.tx", import.meta.url)
   //      |--------------------------|
   //      Now everything is overwritten
   ```

2. Updated tests

> I made it a `minor` change as it should supposedly not modify the functionality of the plugin, but maybe it should be a `major` as it modifies its output?

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/test/snapshots/different-asset-levels-bundle.js
+++ b/packages/rollup-plugin-import-meta-assets/test/snapshots/different-asset-levels-bundle.js
@@ -1,14 +1,14 @@
 const nameOne = 'one-name';
-const imageOne = new URL(new URL('assets/one-deep-d40c1b4b.svg', import.meta.url).href, import.meta.url).href;
+const imageOne = new URL(new URL('assets/one-deep-d40c1b4b.svg', import.meta.url).href).href;
 
 const nameTwo = 'two-name';
-const imageTwo = new URL(new URL('assets/two-deep-e73b0d96.svg', import.meta.url).href, import.meta.url).href;
+const imageTwo = new URL(new URL('assets/two-deep-e73b0d96.svg', import.meta.url).href).href;
 
 const nameThree = 'three-name';
-const imageThree = new URL(new URL('assets/three-deep-801763e8.svg', import.meta.url).href, import.meta.url).href;
+const imageThree = new URL(new URL('assets/three-deep-801763e8.svg', import.meta.url).href).href;
 
 const nameFour = 'four-name';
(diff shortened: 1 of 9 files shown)


cprecioso:
Closing due to old age

bashmish:
Hey, it's actually a great change, too bad I didn't see it before today. I can sync with latest master, I think many tests have changed since this was created. But if you still wanna work on this, I'm happy to assist. Let me know what works best for you :)

cprecioso:
@bashmish Synced and fixed tests

packages/rollup-plugin-import-meta-assets/test/snapshots/different-asset-levels-bundle.js
```diff
@@ -1,14 +1,14 @@
 const nameOne = 'one-name';
-const imageOne = new URL(new URL('assets/one-deep-Bkie7h0E.svg', import.meta.url).href, import.meta.url).href;
+const imageOne = new URL(new URL('assets/one-deep-Bkie7h0E.svg', import.meta.url).href).href;
```
bashmish, line 2: maybe I mislooked first time, but I was under impression that the whole outer URL wrapper will be gone with you change, not only the `import.meta.url`
do you think it's a more complex change or is the outer one still needed?

PR #1987, opened 2022-06-19
Feat/improve snapshot handling

LarsDenBakker:
## What I did

Currently the snapshot plugin has some bugs around caching in-memory. Also when updating snapshots, each individual snapshot does a write to disk and the browser command blocks until this is finished. This makes the browser tests frequently run into timeouts. 

I fixed the caching bugs, and made the snapshots write to disk only once when a test file finishes.


bennypowers (approved):
LGTM

PR #1993, opened 2022-06-28
Replace navigations to `data:,` with `about:blank`.

bicknellr:
This PR replaces navigations to `data:,` with `about:blank`. This lets the Safari extension installed by the version of Selenium used in Saucelabs' Safari 9 and 10 environments [recognize that it should not run](https://github.com/SeleniumHQ/selenium/blob/selenium-2.48.0/javascript/safari-driver/inject/tab.js#L234).

(Fixes #1986)

PR #1994, opened 2022-06-30
test: unskip Selenium and Webdriver tests

web-padawan:
## What I did

1. Removed `describe.skip` in test suites for Selenium and Webdriver so we can run them again.


bicknellr:
Hi, is this PR still on the roadmap?

web-padawan:
Hi, sorry about the delay. Looks like our Selenium tests have issues with Node 18.

I think we can merge #1993 as the change is quite small. I will try to investigate this later when I have time.

bicknellr:
Ok, thanks!

PR #2013, opened 2022-08-05
fix(test-runner-playwright): create BrowserContext per test page

benjamind:
## What I did

Fixed playwright memory leaks by ensuring that we use a new BrowserContext and Page for each test session:

* Create a BrowserContext per Page
* Removed reusing of Pages (since this would retain temporary objects in Playwright and leak memory since we'd be reusing the same BrowserContext).

### DIFF
--- a/.changeset/grumpy-badgers-raise.md
+++ b/.changeset/grumpy-badgers-raise.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-playwright': patch
+---
+
+Fixes playwright memory leak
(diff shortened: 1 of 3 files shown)


Westbrook (commented):
@LarsDenBakker this all looks reason able to me as a change. The question would be whether the API changes here could be considered all "internal" in a way to afford this update as a "patch". I don't _think_ people will have been using these APIs, but can to differ to you on this.

LarsDenBakker (approved):
Looks good to me as well, I think this can be considered as internal details

benjamind:
Thanks folks. I assume this will go out on the automated release later today?

PR #2018, opened 2022-08-08
fix: es-module-lexer logic in dev-server-storybook

thepassle:
`es-module-lexer` previously returned a `exports` as a `string[]`, but since `^1.0.0`, `exports` is now an object.


thepassle:
It seems like TS doesn't want to believe that `exports` is now an array of `ExportSpecifiers`. Not sure how to fix it https://github.com/guybedford/es-module-lexer/blob/main/types/lexer.d.ts#L158

PR #2021, opened 2022-08-09
docs: update link to Playwright CI docs

enqueue:
The link to the Playwright manual was broken.

The use of the `microsoft/playwright-github-action` is discouraged ([You don't need this GitHub Action](https://github.com/microsoft/playwright-github-action#%EF%B8%8F-you-dont-need-this-github-action-%EF%B8%8F)). The recommended way is to install dependencies using the Playwright CLI commands.


web-padawan (approved):
Good catch, thanks! Actually, we no longer use this deprecated action since #1587.
Updated the PR to also use the correct link to Playwright CI documentation.

PR #2032, opened 2022-08-25
Upgrade TypeScript to 4.7.4

justinfagnani:
Also upgrades related ESLint plugins

## What I did

1. `yarn install -D typescript@latest -W`
2. Set version constraint to use `~` since TypeScript doesn't follow semver
3. Fix new errors
  a. Most were cause by catch parameters (`error`) being unknown. I cast these to specific Error types when possible.
  b. One set of errors was around very dynamic console patching that wasn't very type safe to begin with.
4. Upgrade the TypeScript ESLint parser and plugin

I'm unsure whether the upgraded `tsc` will cause .d.ts changes that are backwards incompatible. This should be checked before merging. It's possible, because I needed to upgrade TypeScript to be able to upgrade parse5, since parse5 7.0 uses `type` imports that TypeScript 4.1 can't parse.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "rollup": "^2.66.1",
     "rollup-plugin-terser": "^7.0.2",
     "ts-node": "^10.4.0",
-    "typescript": "4.1.3"
+    "typescript": "~4.7.4"
   },
   "resolutions": {
     "@mdjs/core": "^0.9.0",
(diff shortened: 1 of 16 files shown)


koddsson:
Hey @justinfagnani; I tried solving the merge conflict but am getting CI errors now 😬

koddsson:
I very annoyingly ain't getting these failures locally 😭 This PR might get a bit noisy if I keep trying to debug this against CI.

PR #2037, opened 2022-09-03
Add selectOption command in test-runner-commands

michaelwarren1106:
## What I did

**Background**
Having written a few tests where selecting an option in a `<select>` was necessary, I've always found that the "normal" method of setting a select's value then programmatically triggering the events with `dispatchEvent` kinda breaks the testing paradigm of "interact like a user would" in the case of e2e tests. Using a driver based "select option" command, imo, better fits the model of interacting like a user, since the test code doesn't manually emit events then test if they fired. Tests can just call the `selectOption` command, and the browser will act as if the user clicked an option on the page.

Playwright and Puppeteer both implement a browser driver command for option selection that automatically emits native `change` and `input` events. This PR adds those commands as a `test-runner-commands` plugin.

1. Added a `selectOption` command plugin with tests and documentation such that options can be selected via labels, values, or multiple values, depending on the support available in supported launchers. (unfortunately, selecting options doesn't seem to be implemented in WebDriver for JS)

### DIFF
--- a/docs/docs/test-runner/commands.md
+++ b/docs/docs/test-runner/commands.md
@@ -332,6 +332,84 @@ it('natively holds and then releases a key', async () => {
 
 </details>
 
+### Select option
+
+The `selectOption` command allows you to select an option in a `<select>` tag by either a single value, a label string or as multiple values, depending on the specific implementations of the browser laucher you're using. Once selected, the native `change` and `input` events are dispatched automatically. The `selectOption` command needs the CSS selector to locate the `<select>` by and the value, label, or values to select. The function is async and should be awaited.
+
+The three major launchers all have different support for selecting options, and the `selectOption` takes this into account.Attempting to select an option via a method that is not implemented in the provided launcher will produce errors. Each supported launcher's option selection support is documented below.
+
+- Playwright
+  - Supports selecting options by `label`, `value`, and `multiple values`
+- Puppeteer
+  - Supports selection options by `value`, and `multiple values` only. Selecting options by label is not implemented
+- WebDriver
+  - [No support for selecting options in JS](https://www.selenium.dev/documentation/webdriver/elements/select_lists/)
+
+<details>
+<summary>View example</summary>
+
+```js
+import { selectOption } from '@web/test-runner-commands';
+
+// Assuming the document.body has a select like:
+// <select id="testSelect">
+//   <option value="first">first option</option>
+//   <option value="second">second option</option>
+//   <option value="third">third option</option>
+// </select>
+
+it('natively selects an option by value', async () => {
+  const valueToSelect = 'first';
+  const select = document.querySelector('#testSelect');
+
+  await selectOption({ selector: '#testSelect', value: valueToSelect });
+
+  expect(select.value).to.equal(valueToSelect);
+});
+
+// PLAYWRIGHT ONLY
+it('natively selects an option by label', async () => {
+  const labelToSelect = 'second option';
+  const select = document.querySelector('#testSelect');
+
+  await selectOption({ selector: '#testSelect', label: labelToSelect });
+
+  const expectedSelectedOption = Array.from(document.querySelectorAll('option')).filter(
+    option => option.textContent === labelToSelect,
+  )[0];
+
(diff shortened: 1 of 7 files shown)


daKmoR:
it's a linting "error"

> /home/runner/work/web/web/packages/test-runner-commands/src/selectOptionPlugin.ts
  2:31  error  'puppeteerCore' is defined but never used      @typescript-eslint/no-unused-vars
  4:[15](https://github.com/modernweb-dev/web/runs/8172261151?check_suite_focus=true#step:8:16)  error  'WebdriverLauncher' is defined but never used  @typescript-eslint/no-unused-vars

michaelwarren1106:
oh ok! will do! i'll get that fixed asap! Thanks!

michaelwarren1106:
seems like an all green build, so all thats left is a review. Thoughts on this addition?

docs/docs/test-runner/commands.md
```diff
@@ -332,6 +332,84 @@ it('natively holds and then releases a key', async () => {
...
+//   <option value="first">first option</option>
+//   <option value="second">second option</option>
+//   <option value="third">third option</option>
+// </select>
+
+it('natively selects an option by value', async () => {
+  const valueToSelect = 'first';
+  const select = document.querySelector('#testSelect');
+
+  await selectOption({ selector: '#testSelect', value: valueToSelect });
```
daKmoR, line 365:
```suggestion
  await selectOption({ selector: select value: valueToSelect });
```

that should work right?

michaelwarren1106: yep, it should.

as far as I can tell, its any CSS selector. do you think I should include another test using another selector? I thought about it, but wasnt sure how much value that would provide since there's always several ways to target some HTML element, should we just trust the locator mechanism from the drivers, or should we test them explicitly?

daKmoR: yeah I think we can trust that it works 👍
in the docs, however, it's "strange" as you already have an instance via `querySelector('#testSelect');` and in the following line you "query it again"...

michaelwarren1106: oh, the query is for the `expect` so that I can compare the values. In Playwright and Puppeteer both, the selection commands I'm using only return the "values" that were selected. So I `querySelector` my templated `select` element so that I can make sure that using the `selectOption` command actually changed the `value`. I don't see how I would test that without querying the select for the expect.

If you have a suggestion about that, please let me know!

michaelwarren1106: if I'm reading right, your suggestion is about passing the `select` instance into the `selectOption` command, but the select option commands from  Playwright/Puppeteer only accept CSS selectors, not DOM instances.

https://playwright.dev/docs/api/class-page#page-locator
https://pptr.dev/api/puppeteer.page.select#parameters

daKmoR: yeah I was assuming the playwright API...  (which IMHO is nice)
```js
selectEl.selectOption(value);
```

in that case, you could pass `selectEl` directly or do a locator before (if it is a string)

but the puppeteer API is not as flexible
```js
page.select(selector, value); // where selector needs to be a string
```

so I guess the lowest common denominator means the "worse" API... but it's at least consistent 🤷‍♂️
otherwise, we would need 2 separate ways for playwright and puppeteer... worth it or not?

what do you think?

michaelwarren1106: hrm, good question. I don't really think its worth it. I think in common practice for testing, the selectors passed are probably going to be hardcoded a lot of the time. I know i hardcode test selectors a LOT :). So in that case, there doesnt need to be a querySelector for the `select` necessarily because selecting the option is probably just a middle step in some other test where you're testing some other UI functionality.

Therefore, I think I'd vote for the consistent API even if its "worse".

daKmoR: fair enough 👍

daKmoR:
thaaankkk you 🙇‍♂️

michaelwarren1106:
thanks!

daKmoR:
will be shortly released via `@web/test-runner-commands@0.6.5`

Issue #1334, opened 2021-03-04
appIndex and monorepos


jimsimon:
I'm trying to use the following config for my app and noticed that the `appIndex` setting doesn't seem to work when setting `rootDir` to `../../`:

```javascript
import { esbuildPlugin } from "@web/dev-server-esbuild";
import { fromRollup } from '@web/dev-server-rollup';
import rollupCommonjs from '@rollup/plugin-commonjs';

const commonjs = fromRollup(rollupCommonjs);

export default {
  nodeResolve: true,
  appIndex: `packages/ui/src/index.html`,
  rootDir: '../../',
  debug: true,
  plugins: [
    esbuildPlugin({ ts: true }),
    commonjs({
      include: [
        '../../node_modules/fast-json-stable-stringify/index.js',
        '../../node_modules/zen-observable/**/*'
      ]
    })
  ],
};
```
I've tried using an absolute path and various relative paths (`src/index.html`, `../../packages/ui/src/index.html`) with no success.

LarsDenBakker:
The app index should be within the scope of the root directory. When it's a relative path, it's evaluated relative to the current working directory. What do you mean by it's not working? Where is it failing?

steve2507:
Having stumbled upon the same issue; @jimsimon the appIndex is a relative path to your index.html. As a result, just `src/index.html` should suffice, or add a leading slash to your appIndex path.
After that the app should be available at `http://localhost:8000/packages/ui/src/` and anything you add to the path after that will host the same index.html.

jimsimon:
I've been meaning to circle back to this issue, but that's what I was seeing as well. I was expecting the underlying server to set the serving root to the folder index.html is in so that just visiting localhost:8000 would work.

I also recall running into an issue where I explicitly had to add index.html to the url (i.e. `http://localhost:8000/packages/ui/src/index.html`) which then caused issues with client-side routing iirc.

btelles:
I've identified a possible culprit, but the tool may be working as intended.

My directory structure:

``` 
project
  - packages
    - core
      - ... 
    - example
      - dev
        - index.html
      web-dev-server.config.js 
```

file web-dev-server.config.js
```js
import { legacyPlugin } from "@web/dev-server-legacy";

export default {
  nodeResolve: true,
  preserveSymlinks: true,
  appIndex: "dev/index.html",
  debug: true,
  rootDir: "../../",
  plugins: [
    legacyPlugin({
      polyfills: {
[...]

PR #2048, opened 2022-10-07
Delete prefix requirement for appIndex redirect.

btelles:
See issue #1334 for context.

This removes the requirement that the devserver see the appIndex relative path at the beginning of the url path in order for the devserver to redirect to the index page. As a developer, I'd expect any page that does not match a URL redirect to the index.html file. This is especially helpful for monorepos where the appIndex file ends up pointing to /packages/example/dev/index.html.


btelles:
Gentle ping regarding this PR. Anyone willing to merge this?

btelles:
Is there anything I can do to bring attention to this pull request? I think it's relatively simple and would fix a problem for at least two users.

Issue #1150, opened 2020-12-30
[test-runner-core] Don't watch files when not in "watch" mode


iFwu:
It seems the TestRunner always uses `chokidar` to watch files served. Is it possible to not watch files at all when not in "watch" mode?
The reason I ask is that `chokidar` has some bugs on macOS using `fsevents`. It won't release file descriptors after the watcher is closed. My scenario is to use thread workers to run TestRunner and exit. After running for a while, the number of file descriptors will exceed the system limit.

LarsDenBakker:
Chokidar is used by the dev server for caching. I think we can turn this off when the test runner isn't using watch mode. Is there on issue on chokidar's side about this?

iFwu:
> Chokidar is used by the dev server for caching. I think we can turn this off when the test runner isn't using watch mode. Is there on issue on chokidar's side about this?

I'm not sure if this is related https://github.com/paulmillr/chokidar/issues/1030.
Maybe it can provide an option to choose whether using `fsevents` for `chokidar`. `fsevents` may cause many stability problems from my experiences.

LarsDenBakker:
We should definitely file a bug upstream so that we have a better idea of what's going on. I've and many others been using chokidar on mac for years without any issues, so I'm curious what about your situation is different from our situation.

iFwu:
@LarsDenBakker Hello, I found the root cause of the file descriptors leaking issue.

https://github.com/modernweb-dev/web/blob/adb0833d3bac2ef78063f15062ed70e7aa014159/packages/test-runner-core/src/runner/TestRunner.ts#L152-L155

I changed these lines to:
```typescript
    await Promise.all([
      this.scheduler.stop(),
      this.server.stop().catch(error => {
        console.error(error);
      }),
    ]);
```
And the leaking issue was fixed.
I don't know if that's intentional behavior? I think it's necessary to ensure the file watcher of the dev server gets closed before test runner stops.

IanVS:
This seems reasonable, would you accept a PR, @LarsDenBakker?

PR #2057, opened 2022-10-18
Allow test-runner to run tests without adding files to the watcher

remcovaes:
## What I did

1. Added a `disableFileWatcher` as a flag to the config of the dev-server.
2. Pass `disableFileWatcher` from the test-runner to the dev-server depending on the watch mode of the test-runner.

## Why?

While using the `@web/test-runner` in a Gitlab CI, it was noticed that the test was sometimes failing due to the browser (Playwright) being killed. This was due the a file watch limit in the CI. Due to the flakyness I am not able to give a clear reproduction on this issue. It also does not help that this is a private Gitlab, but I can provide the error:
`2022-10-17T08:06:58.884Z pw:browser [pid=155][err] [1017/080658.883851:ERROR:file_path_watcher_inotify.cc(333)] inotify_init() failed: Too many open files (24)`

It is possible to fix this via a plugin, but I hope to also solve future issues by doing a proper fix. The plugin would look like this if anyone is interested:
``` TypeScript
const unwatchPlugin = (unwatch) => ({
	name: 'unwatch',
	serverStart: ({ fileWatcher }) => {
		if (unwatch) {
			fileWatcher.add = () => {};
		}
	},
});
```
Using the plugin I was able to verify that the watcher had to do with the issue. I am aware that the watcher should not trigger problems in the CI, but I think the simpler a test run is in the CI, the better.

### DIFF
--- a/.changeset/mighty-yaks-develop.md
+++ b/.changeset/mighty-yaks-develop.md
@@ -1,0 +1,6 @@
+---
+'@web/dev-server-core': minor
+'@web/test-runner-core': patch
+---
+
+Add option to disable the fileWatcher in the dev server, to allow the test-runner to run once without files added to the watcher.
--- a/docs/docs/dev-server/cli-and-configuration.md
+++ b/docs/docs/dev-server/cli-and-configuration.md
@@ -86,7 +86,7 @@ interface DevServerConfig {
   open?: 'string' | boolean;
   // index HTML to use for SPA routing / history API fallback
   appIndex?: string;
-  // run in watch mode, reloading when files change
+  // reload the brower on file changes.
   watch?: boolean;
   // resolve bare module imports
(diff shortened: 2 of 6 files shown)


remcovaes:
This pull request would also close #1150

Westbrook (approved):
LGTM. Thanks for the contribution, sorry it took so long to process.

tmsns:
It seems that this change is breaking the manual mode of the test runner. Files are not watched (and updated when changed) in manual mode.

I propose to change its initialization to not disabled the filewatcher when in manual mode. https://github.com/modernweb-dev/web/pull/2057/files#diff-04c616f06dfee655a5a69094b9eb017357a02eba8c977b338fb11e02e54e0456R49

I'll create an issue and a PR for it.

PR #2094, opened 2023-01-06
fix(all): add types to exports for node16/nodenext module resolution

bgotink:
Add types to exports for typescript's node16 module resolution.

Two packages were already fixed in #1973 and #1981, this PR fixes the rest.

I automated this change using [`jq`](https://stedolan.github.io/jq/) to prevent typos or other small mistakes in this repetitive task.

<details>
<summary>jq script</summary>

```jq
#!/usr/bin/env -S jq -f

def type_path:
	sub("\\.[mc]?js$"; ".d.ts")
	;

def add_types_to_export:
	if type == "string" then
		{ types: type_path, default: . }
	elif .types != null then
		.
	else
		map_values(add_types_to_export)
	end
	;

def add_types_to_exports:
	if keys[0] | startswith(".") then
		map_values(add_types_to_export)
	else
		add_types_to_export
	end
	;

if .exports != null then
	.exports |= add_types_to_exports
else
	.
end
```

</details>

### DIFF
--- a/packages/browser-logs/package.json
+++ b/packages/browser-logs/package.json
@@ -16,8 +16,14 @@
   "main": "dist/index.js",
   "exports": {
     ".": {
-      "import": "./index.mjs",
-      "require": "./dist/index.js"
+      "import": {
+        "types": "./index.d.ts",
+        "default": "./index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      }
     }
   },
   "engines": {
--- a/.changeset/tasty-waves-allow.md
+++ b/.changeset/tasty-waves-allow.md
@@ -1,0 +1,34 @@
+---
+'@web/browser-logs': patch
+'@web/config-loader': patch
+'@web/dev-server': patch
+'@web/dev-server-core': patch
+'@web/dev-server-esbuild': patch
+'@web/dev-server-hmr': patch
+'@web/dev-server-import-maps': patch
+'@web/dev-server-legacy': patch
+'@web/dev-server-rollup': patch
+'@web/dev-server-storybook': patch
+'@web/parse5-utils': patch
+'@web/polyfills-loader': patch
+'@web/rollup-plugin-copy': patch
+'@web/rollup-plugin-html': patch
+'@web/rollup-plugin-import-meta-assets': patch
+'@web/rollup-plugin-polyfills-loader': patch
+'rollup-plugin-workbox': patch
+'@web/test-runner-browserstack': patch
+'@web/test-runner-chrome': patch
+'@web/test-runner-cli': patch
+'@web/test-runner-commands': patch
+'@web/test-runner-core': patch
+'@web/test-runner-coverage-v8': patch
+'@web/test-runner-junit-reporter': patch
+'@web/test-runner-playwright': patch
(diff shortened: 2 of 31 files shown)


packages/browser-logs/package.json
```diff
@@ -16,8 +16,14 @@
   "main": "dist/index.js",
   "exports": {
     ".": {
-      "import": "./index.mjs",
-      "require": "./dist/index.js"
+      "import": {
+        "types": "./index.d.ts",
```
justinfagnani, line 20: These files should be named `index.d.mts`... and if they are named correctly and siblings of the files they describe, then they _don't_ actually need to be in the package exports. See some discussion [here](https://github.com/microsoft/TypeScript/issues/52363) and some docs [here](https://www.typescriptlang.org/docs/handbook/esm-node.html#:~:text=Furthermore%2C%20TypeScript%20also%20supports%20two%20new%20declaration%20file%20extensions%3A%20.d.mts%20and%20.d.cts.)

bgotink: That would work for me, but it would break all packages in typescript when not using node16/nodenext module resolution, because in that mode typescript only looks for `.d.ts` files.

justinfagnani: That shouldn't be true. My understanding, reiterated in the issue I liked to, is that TypeScript only looks for `.d.mts` declarations files for `.mjs` files, `.d.cts` declaration files for `.cjs` files. IOW, the declaration file _must_ be named after the file it's describing always, and there wasn't event a version of TypeScript that even supported `.mjs` files with `.d.ts` declarations.

bgotink: You're correct, but that doesn't stop it from being broken. If typescript isn't configured to use Node16/NodeNext resolution, typescript looks for the `d.ts` file extension but it doesn't look for `.d.mts` or `.d.cts` file extensions. That makes sense as Node doesn't autocomplete the `.cjs` or `.mjs` extensions either.

The `@web/dev-server-core` package has an export `./test-helpers`. To support this export in non-Node16/NodeNext resolutions the package must have a `./test-helpers.d.ts` file.

andrewbranch: > To support this export in non-Node16/NodeNext resolutions the package must have a ./test-helpers.d.ts file.

You’re referring to resolution in the TypeScript mode called `node`, which in 5.0 has been renamed to `node10` to reflect what it really is. Node 10 will fail at runtime on an import of `"@web/dev-server-core/test-helpers"`, so putting a `.d.ts` file at `./test-helpers.d.ts` to make TypeScript think it will resolve is strictly wrong. Users who are running on a Node version newer than v10 should not be using that setting anymore.

Of course, taking my pedant hat off, users _are_ still using that setting even though the vast majority of them are no longer using Node 10. It’s up to you to decide how and if you want to protect those users from their own misconfiguration. Placing the phantom `.d.ts` file there works. Other strategies are demonstrated at https://github.com/andrewbranch/example-subpath-exports-ts-compat. Particularly interesting is that you can use `typesVersions` to affect TypeScript `node10` resolution almost exactly the same way that `exports` affects the modern TypeScript resolution modes.

Whatever strategy you decide to go with, you can use https://arethetypeswrong.github.io to test it. I can confirm that the issue pointed out here by @justinfagnani is indeed a problem that will be recognized by this tool.

PR #2105, opened 2023-01-22
fix: avoid css modules from being preloaded as script

thepassle:
When using native css modules in rollup, the polyfills loader will try to add a preload link for them, e.g.:
`<link rel="preload" href="styles-78a368d53d94321d.css" as="script" crossorigin="anonymous" />`, should be preloaded as script instead


thepassle:
Awaiting https://github.com/whatwg/html/issues/8764

Issue #2066, opened 2022-10-29
Types are not working when using `moduleResolution: "NodeNext"`


daKmoR:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@web/dev-server-core@0.3.19` for the project I'm working on.

TS using `"moduleResolution": "NodeNext"` will say it can't find types for `@web/dev-server-core`...

If I add it like so it works fine 🤗

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@web/dev-server-core/package.json b/node_modules/@web/dev-server-core/package.json
index c65693c..0e0bdc9 100644
--- a/node_modules/@web/dev-server-core/package.json
+++ b/node_modules/@web/dev-server-core/package.json
@@ -16,6 +16,7 @@
   "main": "dist/index.js",
   "exports": {
     ".": {
+      "types": "./index.d.ts",
       "import": "./index.mjs",
       "require": "./dist/index.js"
     },
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

hungtcs:
+1 same issues for node16

PR #2108, opened 2023-01-24
fix(test-runner): Fix the typings file name for the main entrypoint

justinfagnani:
Fixes https://github.com/modernweb-dev/web/issues/2066

## What I did

1. Renamed `packages/test-runner/index.d.ts` to `index.d.mts`
2. Added `"*.d.mts"` to `package.json` files field

This fixes the problem slightly differently than @daKmoR 's patch in https://github.com/modernweb-dev/web/issues/2066. With TypeScript you do not need a types field if the source file has a sibling type declaration file of the correct name. The problem here is that `.mjs` files need a `.d.mts` declaration. See https://www.typescriptlang.org/docs/handbook/esm-node.html


Westbrook (approved):
LGTM.

Issue #1950, opened 2022-05-02
[dev-server] Default node resolution


justinfagnani:
I think it'd be really helpful to have a default resolution mode for `@web/dev-server`, and for that default to be a new mode closer to Node module and browser w/ import maps resolution.

There are two main reasons I'd like a default mode:
1. I have to pass `--node-resolution` _every_ time I use WDS. Basically no project I've ever had works without it, so it's pure boilerplate.
2. I want more users to use our Lit development builds, so I want the default to include the `development` export condition. This is a dev server after all :) and it would greatly improve DX for users to get our more extensive development build warnings. I hope this will be true of more projects that adopt export conditions of time too.

As for the mode, I really want a mode that resolves bare specifiers and package names, but not the rest of the path (except for when package exports are available). Node doesn't do path searching for modules, and import maps can't do suffix mappings, so the current node resolution can hide subtle mistakes where you intended to use a file extension, but forgot.

Westbrook:
We namely leverage the Resolve plug-in from Rollup to power this in dev server. Have you raised this question there?

daKmoR:
I also use `@web/dev-server` to check if a build output of an app actually would work on a "default web server".

So I do
1. Developing `web-dev-server --node-resolve ./page`
2. Build it `rollup ...`
3. Check build `web-dev-server ./_site`

it usually comes back to the question... do we wanna have a "magical" thing that just works... and you may have a surprise as you push it to a real server (e.g. if it "just" works - you may be inclined to just push your source)

so as you provide the option you are at least aware that your server will need to have the same option/functionality or you need to transform your code so that this functionality is no longer needed...

always a balancing act 😅

here is a crazy idea - ship "2 clis/servers"?
- `web-dev-server` comes with some default options like node resolve
- `web-server` comes with no default options

justinfagnani:
I hit this again today when I forgot that Web Dev Server does not set the `development` condition by default. I think a lot of other developer tools are doing this by default now (Vite and Next?) and it's a bit weird that a "dev" server isn't.

It's also cumbersome to have to write a config file to set the option. I think for a dev server this should just happen out of the box.

PR #2109, opened 2023-01-24
fix(dev-server): Enable the development export condition by default

justinfagnani:
Partial fix for https://github.com/modernweb-dev/web/issues/1950

This adds the "development" export condition _when_ node resolution is already enabled, which at least removes the need for a config file when using the `--node-resolve` flag or `nodeResolve: true` config option.

I still think the flag should be enabled by default as it's used in basically every invocation of wds that I've ever seen.

## What I did

1. Added `exportConditions: ['development']` to the default config options in the nodeResolvePlugin
2. Added a test that checks for a global that lit-html only sets in the development build.

### DIFF
--- a/.changeset/brown-lies-repeat.md
+++ b/.changeset/brown-lies-repeat.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server': patch
+---
+
+Enable the "development" export condition by default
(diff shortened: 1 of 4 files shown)


Westbrook:
Can n you rebase this to see if it'll kick the tests into gear?

justinfagnani:
@Westbrook just rebased and force-pushed

justinfagnani:
@Westbrook what else do I need to do here?

Westbrook:
Hrm. Looks like you caught the rebase while we had some issue in `master` 🙈. Sorry about that. Can you do another rebase so we can confirm CI before approving this?

Westbrook:
We'd love to get this in for you if you could try the rebase again.

justinfagnani:
@Westbrook rebased again

thepassle:
fixed in https://github.com/modernweb-dev/web/pull/2375

PR #2113, opened 2023-02-08
Format uncaught errors

koddsson:
Uncaught promise rejections aren't formatted with a nice diff (if they happen to be snapshot errors) so I wanted to make sure they are formatted.


koddsson:
Maybe I'll come back to this PR later but I don't know what I'm doing right now.

PR #2114, opened 2023-02-10
chore: update all rollup dependencies to run CI with

Westbrook:
## What I did

1. Take latest rollup dependencies across the board.
2. Open a draft PR to test the versions on CI.
3. If it works, we can have the discussion about mixed version resolutions or hard breaking release or whatever.
4. If not...something else 😄


Westbrook:
A follow up question for this (if we can get tests passing) will end up being "do we make a Rollup@3 only release?"

43081j:
@Westbrook it looks like rollup plugins have changed structure...

before, a plugin had `{resolveId: Function}`

now it has `{resolveId: {handler: Function, order: 'pre'|'post'}}`

[r3.patch](https://github.com/modernweb-dev/web/files/10826157/r3.patch)

Westbrook:
I've very naively taken all of the changes that come with this update and applied them to the tests. They _feel_ OK-ish for the most part, but I don't 100% know what _should_ be in this area.

koddsson:
This is great work! Is there something I can do to help with this PR?

Maybe we can do a pre-release and start testing in various environment. I'd love to take this for a spin on projects at ING to see if there's something broken that we can find.

Westbrook:
@koddsson A pre-release is probably a good idea! I've mainly just _assumed_ that the process worked and accepted all of the new test rules. They're stable... which is something.

If you were well versed in the sorts of transforms we're doing with CJS and ESM, extra eyes on the updated tests outputs would be a great first step. @43081j has been working a lot in this area recently, and I hope he has some time to check these changes out as well.

The other area that I can think of right now would be what sort of version change this is. Technically, I don't think our APIs change, but I guess we do remove support for Rollup@2, which would be a breaking change. Would _everything_ need to break? Just `@web/dev-server-rollup` and similar? Something else?

koddsson:
> @koddsson A pre-release is probably a good idea! I've mainly just _assumed_ that the process worked and accepted all of the new test rules. They're stable... which is something.

Yeah, I might be overly cautious here 😄 If the tests are green, that might just be fine. Then if there are issues we can make bugfix releases. 

> If you were well versed in the sorts of transforms we're doing with CJS and ESM, extra eyes on the updated tests outputs would be a great first step. @43081j has been working a lot in this area recently, and I hope he has some time to check these changes out as well.

I can take a look and tell you if I spot something out of the ordinary 😄 

>The other area that I can think of right now would be what sort of version change this is. Technically, I don't think our APIs change, but I guess we do remove support for Rollup@2, which would be a breaking change. Would everything need to break? Just @web/dev-server-rollup and similar? Something else?

I think a breaking change for `@web/dev-server-rollup` makes sense. Looking at the changes, I _think_ we should probably make a breaking changes for anything that upgrades from Rollup 2 to Rollup 3 so that we don't accidentally break people's build with this change. Packages with no Rollup updates like `test-runner-mocha` can just go out in a patch release.

Westbrook:
Going to prerelease this now! 🤞🏼

koddsson:
I've updated `@web/dev-server`, `@web/test-runner` and `@web/dev-server-storybook` in ~20 internal repositories and everything is looking great ✨

packages/dev-server-rollup/test/node/fixtures/commonjs/commonjs-browser-test.js
```diff
@@ -33,7 +34,7 @@ describe('commonjs', () => {
   });
 
   it('can handle require default', () => {
-    expect(requiredDefault).to.equal('foo');
+    expect(requiredDefault.default).to.equal('foo');
```
Westbrook, line 37: @43081j this feels like an error to have converted, in that this same wrapping of default is causing me some trouble with the canaries...thoughts?

packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -695,6 +695,7 @@ describe('rollup-plugin-html', () => {
     }
 
     const outputHtml = getAsset(output, 'index.html').source;
+    console.log(outputHtml);
```
koddsson, line 698: Looks like a `console.log` was left in here ☝🏻

packages/dev-server-rollup/test/node/fixtures/commonjs/commonjs-browser-test.js
```diff
@@ -24,6 +24,7 @@ describe('commonjs', () => {
   });
 
   it('can handle compiled es modules with named exports', () => {
+    console.log(compiledEsm);
```
koddsson, line 27: And this console.log

Issue #1832, opened 2022-01-03
[dev-server-storybook] start fails due to require() of ES modules is not supported when package.json type: module is used 


samvimes01:
Hello.
There is an issue starting dev-server-storybook when it's used in a library with `"type: module"`

###Steps to reproduce:
1. start a package with packcage.json `"type: module"` (or for example clone or download https://github.com/lit/lit-element-starter-ts)
2. according to the current [guide](https://modern-web.dev/guides/dev-server/storybook/) add the dev-server-storybook
  - create a `.storybook/main.js` with `exports.default = ` (N.B. - it's the source of issue)
  - create a `.storybook/server.config.mjs`
3. run `wds --config .storybook/server.config.mjs`
4. got the error `ERR_REQUIRE_ESM` `require() of ES modules is not supported`

```
> wds --config .storybook/server.config.mjs

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: ~/lit-element-starter-ts-main/.storybook/main.js
require() of ES modules is not supported.
require() of ~/lit-element-starter-ts-main/.storybook/main.js from ~/lit-element-starter-ts-main/node_modules/@web/dev-server-storybook/dist/shared/config/readStorybookConfig.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename main.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from ~/lit-element-starter-ts-main/package.json.
```

### Possible solution 1
remove `"type: "module"` from your package.json

### Possible solution 2
- fix `packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts`
rename hardcoded main.js  to main.cjs
- fix the docs and guide (rename hardcoded main.js  to main.cjs)

samvimes01:
This PR https://github.com/modernweb-dev/web/pull/1831 should fix the issue

PR #2116, opened 2023-02-13
[dev-server-storybook] Load a CommonJS config if one exists

koddsson:
This PR checks if there's a `main.cjs` file and loads that as the storybook configuration file if it exists instead of `main.js`.

Closes https://github.com/modernweb-dev/web/issues/1832
Closes https://github.com/modernweb-dev/web/pull/1831

### DIFF
--- a/packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts
+++ b/packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts
@@ -47,7 +47,10 @@ export function readStorybookConfig(pluginConfig: StorybookPluginConfig): Storyb
   let previewHead: string | undefined = undefined;
   let previewBody: string | undefined = undefined;
 
-  if (!fs.existsSync(mainJsPath) && !fs.existsSync(commonJsMainPath)) {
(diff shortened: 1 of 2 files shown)


packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts
```diff
@@ -60,7 +63,9 @@ export function readStorybookConfig(pluginConfig: StorybookPluginConfig): Storyb
     previewBody = fs.readFileSync(previewBodyPath, 'utf-8');
   }
 
-  const mainJs = validateMainJs(require(mainJsPath));
+  const mainJs = fs.existsSync(commonJsMainPath) ?
```
LarsDenBakker, line 66: We are doing `fs.existsSync(commonJsMainPath)` twice. Should we make it a variable?

LarsDenBakker (changes requested):

packages/dev-server-storybook/src/shared/config/readStorybookConfig.ts
koddsson: Sure!

Issue #1700, opened 2021-09-28
[@web/dev-server]: `@rollup/plugin-commonjs` version >18 breaks commonjs imports


peschee:
Consuming dependencies that have to be transformed using the `@rollup/plugin-commonjs` plugin break when version `>18` of the plugin is used.

I have a simple example repository: https://github.com/peschee/wds-commonjs-example

In the `main` branch, trying to run wds (`npm run start`) shows the following error when importing `apexcharts` or `moment` (those are the ones I tested):

```
Uncaught SyntaxError: The requested module '/__web-dev-server__/rollup/moment.js?web-dev-server-rollup-null-byte=%00%2FUsers%2Fpsiska%2Ftemp%2Fwds-commonjs-test%2Fnode_modules%2Fmoment%2Fmoment.js%3Fcommonjs-module' does not provide an export named 'exports'
```

I also have a simple rollup build in the repo, using the same plugins as in wds, and that one works (`npm run build`).

Downgrading `@rollup/plugin-commonjs` to `^18` makes the imports work again in wds: https://github.com/peschee/wds-commonjs-example/pull/1/files

Pratik29121:
const basename = process.env.NODE_ENV === 'production' ? '/my-react-app' : '/';

return (
  <Router basename={basename}>
     <App/>
  </Router>
);

Pratik29121:
const basename = process.env.NODE_ENV === 'production' ? '/my-react-app' : '/';

return (
  <Router basename={basename}>
     <App/>
  </Router>
);

phihu:
The problem still exists. Installing the @rollup/plugin-commonjs@22.0.0-4 (beta) did not help but produces another error: 
`Error while transforming node_modules/axios/index.js: Could not resolve import "commonjsHelpers.js".`

akrawitz:
I seem to have hit this as well. I was trying to run tests on code that uses [jStat](https://github.com/jstat/jstat). *Before* I added `@rollup/plugin-commonjs` I was getting the error:

~~~
 🚧 Browser logs:
      TypeError: Cannot set properties of undefined (setting 'jStat')
        at ..\..\node_modules\jstat\dist\jstat.js:7:22
        at ..\..\node_modules\jstat\dist\jstat.js:9:3
~~~

Fair enough, since jStat is only provided in UMD format, I added the CommonJS plugin to my Web Test Runner config:

~~~js
import rollupPluginCommonjs from '@rollup/plugin-commonjs';
[...]

PR #2129, opened 2023-02-22
Prepare release of CJS and Node Resolution updates

Westbrook:
## What I did

1. Accepted @43081j's updates and pre-released them.
2. Users are testing them in https://github.com/modernweb-dev/web/issues/1700#issuecomment-1432372016
3. It's possible (though I'm likely lying to myself) that this support avoiding some of the failures in https://github.com/modernweb-dev/web/pull/2114

### DIFF
--- a/.changeset/tasty-balloons-pay.md
+++ b/.changeset/tasty-balloons-pay.md
@@ -1,0 +1,6 @@
+---
+'@web/dev-server-core': minor
+'@web/dev-server-rollup': minor
+---
+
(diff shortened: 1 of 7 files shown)


Westbrook:
This seems to be working for people. Should we move ahead with merging it in? @43081j sound good?

43081j:
I'm pretty confident it works 👍 so it's a go from me

Worst case I'll pick up any related issues and fix them. But don't think there will be any 🤞 We've been using it for a while now without any hiccups

Issue #1929, opened 2022-04-09
Unable to use @web/dev-server-esbuild#0.3.0 with TS and lit


jarrodek:
Related to https://github.com/lit/lit/issues/2716

After updating to version 0.3.0 I started seeing errors related to the use of decorators. I was able to pinpoint the problem to the update of the `@web/dev-server-esbuild` library in my projects. I was able to run my projects again after downgrading to version `0.2.16`.

Please, refer to the referenced issue for more details.

LarsDenBakker:
This is likely an issue with esbuild, and/or your configuration. The plugin is just a passthrough.

jarrodek:
Mine configuration is pretty much what's in the @open-wc init with addition of `esbuildPlugin` and some local API for development:

```javascript
// import { hmrPlugin, presets } from '@open-wc/dev-server-hmr';
import { esbuildPlugin } from '@web/dev-server-esbuild';
import LocalApiStore from './demo/store/StorePlugin.js';

/** Use Hot Module replacement by adding --hmr to the start command */
const hmr = process.argv.includes('--hmr');

export default /** @type {import('@web/dev-server').DevServerConfig} */ ({
  open: '/demo/',
  /** Use regular watch mode if HMR is not enabled. */
  watch: !hmr,
  /** Resolve bare module imports */
  nodeResolve: {
    exportConditions: ['browser', 'development'],
  },

  /** Compile JS for older browsers. Requires @web/dev-server-esbuild plugin */
  // esbuildTarget: 'auto'

  /** Set appIndex to enable SPA routing */
  // appIndex: 'demo/index.html',

  plugins: [
    esbuildPlugin({ ts: true, target: 'auto' }),
    /** Use Hot Module Replacement by uncommenting. Requires @open-wc/dev-server-hmr plugin */
    // hmr && hmrPlugin({ exclude: ['**/*/node_modules/**/*'], presets: [presets.litElement] }),
    LocalApiStore,
  ],

  // preserveSymlinks: true,

  // See documentation for all available options
});
```

I also didn't change much in the tsconfig file:

```json
{
  "compilerOptions": {
    "target": "es2018",
    "module": "esnext",
    "moduleResolution": "node",
    "noEmitOnError": true,
    "lib": ["es2017", "dom"],
    "strict": true,
    "esModuleInterop": false,
    "allowSyntheticDefaultImports": true,
    "experimentalDecorators": true,
    "importHelpers": true,
    "outDir": "dist",
[...]

PR #2137, opened 2023-03-10
Update `esbuild` dependency of `@web/dev-server-esbuild` to fix bug with decorators

jpzwarte:
Version `0.3.3` of `@web/dev-server-esbuild` uses `0.14.54` of esbuild. This version has a bug that breaks experimental typescript decorators. Apparently this has been known quite a while and the workaround is to fix the version to `0.2.16` of dev-server-esbuild.

This PR updates the esbuild dependency to 0.16 || 0.17.

All tests still pass.

fixes https://github.com/modernweb-dev/web/issues/1929

### DIFF
--- a/packages/dev-server-esbuild/package.json
+++ b/packages/dev-server-esbuild/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mdn/browser-compat-data": "^4.0.0",
     "@web/dev-server-core": "^0.3.19",
-    "esbuild": "^0.12 || ^0.13 || ^0.14",
+    "esbuild": "^0.16 || ^0.17",
     "parse5": "^6.0.1",
     "ua-parser-js": "^1.0.2"
   },
--- a/.changeset/kind-cycles-listen.md
+++ b/.changeset/kind-cycles-listen.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server-esbuild': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/dev-server-esbuild/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mdn/browser-compat-data": "^4.0.0",
     "@web/dev-server-core": "^0.3.19",
-    "esbuild": "^0.12 || ^0.13 || ^0.14",
+    "esbuild": "^0.16 || ^0.17",
```
Westbrook, line 55: If there are no consumed APIs changing as part of this update, is there anything wrong with leaving the previous verions? `0.14.0` is at least a year old, so I'm not against dropping them, just wanted to get your thoughts. As a "breaking" change (patches before 1.0) it's not an issue either way.

Westbrook (approved):
This update make sense. If there are no opinions to the contrary, don't let me forget to merge this next week sometime.

packages/dev-server-esbuild/package.json
jpzwarte: I suppose we could just leave the old ones. On the other hand; *some* old ones are broken, so we could prevent anyone from going down the same rabbit hole?

Westbrook: Oh, yeah, if they are broken, let's use the breaking semantics of the change you've listed and exclude them. Great call!

PR #2161, opened 2023-03-15
feat(dev-server-esbuild): allow extended tsconfig

venikx:
## What I did

1. Added a test case when referencing a `tsconfig.json` in the `esbuildPlugin` that uses extends to extend from a base `tsconfig.json`
2. Added a possible implementation how it could work using `--showConfig` when running `tsc`. See https://www.typescriptlang.org/docs/handbook/compiler-options.html
3. Change the `target` of the fixture, since es2022 was reported as invalid by `tsc`

I'm not sure about the actual implementation, but it covers my test case, without breaking the other tests 🤷🏽


venikx:
> ### ⚠️ No Changeset found
> 
> Latest commit: [ad0ca62](https://github.com/modernweb-dev/web/commit/ad0ca62055e59308aff65949246ae984d94fc2f4)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> This PR includes no changesets
> 
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/master/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/venikx/modernweb-dev-web/new/feat/extends-tsconfig?filename=.changeset/popular-dancers-raise.md&value=---%0A%22%40web%2Fdev-server-esbuild%22%3A%20patch%0A%22%40web%2Fdev-server%22%3A%20patch%0A---%0A%0Afeat(dev-server-esbuild)%3A%20allow%20extended%20tsconfig%0A)

So if I understand this correctly I can trigger an automatic release of the package, by applying a changeset?

Westbrook:
Apply the changeset to clarify that there are changes requiring a release.

Full the changeset with a lingual representation of the changes so that they are included in the CHANGELOG.md.

venikx:
> Apply the changeset to clarify that there are changes requiring a release.
> 
> Full the changeset with a lingual representation of the changes so that they are included in the CHANGELOG.md.

Let me know if the formatting of the changeset should be adjusted!

packages/dev-server-esbuild/src/EsbuildPlugin.ts
```diff
@@ -61,7 +62,15 @@ export class EsbuildPlugin implements Plugin {
     this.config = config;
     this.logger = logger;
     if (this.esbuildConfig.tsconfig) {
-      this.tsconfigRaw = await promisify(fs.readFile)(this.esbuildConfig.tsconfig, 'utf8');
+      try {
+        const { stderr, stdout } = await promisify(child_process.exec)(
```
koddsson, line 66: Does reading the file not work? We have to shell out to `tsc`?

koddsson: Oh, interesting! I'm wondering if we should really be using the `.tsconfig` setter instead of `.tsconfigRaw`. I'm not sure, but from reading the docs, it seems like that setting will work with extending configs?

If not, I think this is good!

venikx: What do you mean with the `.tsconfig` setter? And are you referring to the dev-server-esbuild setting, or esbuild itself?

venikx: @koddsson I just tested this again, and with the test case I added the test fails, without the change I've proposed here.

koddsson (approved):
This is great! I'm slightly worried about performance here since we'll be shelling out to `tsc` but it's probably fine.

We should probably have some performance testing in the future to help us get insights into how changes affect performance.

koddsson:
We've got #2180 queued up to merge to `master` soon. I want to make sure we land that first so that we don't delay that release any more. We'll merge this once that PR has landed.

venikx:
No worries, I've come across another way to possibly solve it while going through the dependabot PR's at work. Would something like this be preferred over shelling out `tsc`? https://github.com/qmhc/vite-plugin-dts/pull/202

koddsson:
> No worries, I've come across another way to possibly solve it while going through the dependabot PR's at work. Would something like this be preferred over shelling out `tsc`? [qmhc/vite-plugin-dts#202](https://github.com/qmhc/vite-plugin-dts/pull/202)

Oh, yeah! I would love to do this instead. Looks like we need to upgrade to TypeScript 5 first, which I think is also a positive, so I'd prefer that.

What do you think?

venikx:
> > No worries, I've come across another way to possibly solve it while going through the dependabot PR's at work. Would something like this be preferred over shelling out `tsc`? [qmhc/vite-plugin-dts#202](https://github.com/qmhc/vite-plugin-dts/pull/202)
> 
> Oh, yeah! I would love to do this instead. Looks like we need to upgrade to TypeScript 5 first, which I think is also a positive, so I'd prefer that.
> 
> What do you think?

Yeah, I think it might be better, but it does indeed seem to require typescript 5. Would that then be considered a breaking change for dev-server-esbuild?

koddsson:
> Would that then be considered a breaking change for dev-server-esbuild?

Probably! But that's no worries, we can just make a breaking change.

koddsson:
We should be on TypeScript 5 now if you pull master 👯

venikx:
> We should be on TypeScript 5 now if you pull master 👯

I'll modify my PR somewhere this weekend, I have busy schedule for this week unfortunately.

koddsson:
> > We should be on TypeScript 5 now if you pull master 👯
> 
> I'll modify my PR somewhere this weekend, I have busy schedule for this week unfortunately.

No worries! Let me know if we can help :)

PR #2176, opened 2023-03-24
Try `chai@next`

koddsson:
Trying out the `chai@5` prerelease so we can see if there are issues that we can feed back to the chai team.


koddsson:
This is looking good! I'm thinking I'll let this sit until Chai@5 lands.

integration/test-runner/package.json
```diff
@@ -14,7 +14,6 @@
   },
   "author": "modern-web",
   "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/test-runner-integration-tests",
-  "main": "index.js",
```
thepassle, line 17: Why remove this?

package.json
```diff
@@ -6,6 +6,7 @@
     "type": "git",
     "url": "https://github.com/modernweb-dev/web.git"
   },
+  "type": "module",
```
thepassle, line 9: Why was this needed? I do think this can have impact on packages that use it

thepassle:
This is cool dude, nice work. Changes look pretty straightforward (mostly import changes), nice.

integration/test-runner/package.json
koddsson: It's unrelated to the chai changes. I was just getting warnings when running the tests:

```
(node:87634) [DEP0128] DeprecationWarning: Invalid 'main' field in '/src/web/integration/test-runner/package.json' of 'index.js'. Please either fix that or report it to the module author
```

Westbrook:
while you're killing it here, any thoughts on a `mocha@10.x` PR? 😉

koddsson:
> while you're killing it here, any thoughts on a `mocha@10.x` PR? 😉

Sure! I'll look into that 😄

PR #2177, opened 2023-03-26
Upgrade mocha

koddsson:
Update Mocha from 8.2.1 to 10.2.0!


Westbrook (commented):
The breaking changes in `v9` are where Dependabot failed us on this too, but since then it seems that Mocha has moved to `v10` ([npm listing](https://www.npmjs.com/package/mocha)). Would be great to move all the way through to that version, here.

koddsson:
> The breaking changes in `v9` are where Dependabot failed us on this too, but since then it seems that Mocha has moved to `v10` ([npm listing](https://www.npmjs.com/package/mocha)). Would be great to move all the way through to that version, here.

For sure, I did that initially, but just had so much breakage that I thought I'd try to do this in smaller steps. Trying to figure out what breaks still 😄

Westbrook (commented):
Does this count as a breaking change? _Likely_.

Should we target #2180 for this PR?

Westbrook:
Looks like you _may_ just be missing these two https://github.com/koddsson/web/blob/update-mocha/package.json#L53-L68 and then the breaks of the change will open unto you?

koddsson:
> Looks like you _may_ just be missing these two https://github.com/koddsson/web/blob/update-mocha/package.json#L53-L68 and then the breaks of the change will open unto you?

Yup! I've done this locally but didn't push it since it was broken. But I can push so we can all see the error in CI :)

43081j:
i think you need to still use the global since its testing that the setup worked correctly (i.e. its testing the `window` object has mocha on it)

but even then i think ci will fail 😭

43081j:
@koddsson got it:

```ts
const mocha = (window as any).mocha;
const mochaExports = (window as any).Mocha;

class SilentReporter extends mochaExports.reporters.Base {
}
```

the exports got moved to `window.Mocha` while the mocha instance is at `window.mocha`

koddsson:
@43081j Let's hope this works! 🤞🏻

koddsson:
Amazing work here @43081j! I had done some changes with the `BaseReporter` locally, but never got it to work. Really happy you were able to figure this out <3 <3

43081j:
Awesome, glad it worked!

So the original code wasn't far off actually

Pretty much `window.mocha.Mocha` moved to `window.Mocha`

packages/test-runner-mocha/src/mochaSetup.ts
```diff
@@ -1,8 +1,9 @@
 import { styles } from './styles.js';
 
-const mocha = (window as any).mocha as BrowserMocha;
-const BaseReporter = (mocha as any).Mocha.reporters.Base;
-class SilentReporter extends BaseReporter {}
+const mocha = (window as any).mocha;
```
Westbrook, line 3: Is it the removal of `as BrowserMocha` that’s making TS angry here?

43081j: You can keep the cast fyi, should work fine

koddsson:
Looks like we are all green! Let me know what you think, and feel free to merge at your own convenience.

Westbrook (commented):
Exciting! Need to do a double check on a PC in the morning, but looking good. Nice merges we’ll get one last canary, and assuming there’s no issues, and no last minute feedback to the contrary, we’ll go to `master` with the whole set of change! 🎊

Westbrook (approved):
LGTM!

This merge should auto release a new canary, and with a little testing there, maybe we're good to go?

Ecosystem wise, are most other people also leveraging this through `@open-wc/testing-helpers`? Or is it just me? We'll need to update those as well.

Issue #2212, opened 2023-04-04
[rollup-plugin-workbox] recently upgraded dependency on pretty-bytes fails builds


dlockhart:
The [recent upgrade of pretty-bytes](https://github.com/modernweb-dev/web/pull/2208) from version 5.6.0 to 6.1.0 which was released in `rollup-plugin-workbox` version 6.2.1 is causing Rollup builds using `@open-wc/building-rollup` to fail with the following error:

```
Error: While loading the Rollup configuration from "rollup.config.js", Node tried to require an ES module from a CommonJS file, which is not supported. A common cause is if there is a package.json file with "type": "module" in the same folder. You can try to fix this by changing the extension of your configuration file to ".cjs" or ".mjs" depending on the content, which will prevent Rollup from trying to preprocess the file but rather hand it to Node directly.
```

It seems related to the fact that `pretty-bytes` [upgraded to be ESM-based](https://github.com/sindresorhus/pretty-bytes/compare/v5.6.0...v6.1.0) and now has `"type": "module"`. The way in which it gets imported into `rollup-plugin-workbox` may need to be adjusted.

## Workaround

You can work around this by pinning the version of `pretty-bytes` in your `package.json`:

```json
{
  "overrides": {
    "pretty-bytes": "5.6.0"
  }
}
```

thepassle:
Would you mind creating a pr to fix it? :)

dlockhart:
Fixed in #2213.

PR #2213, opened 2023-04-04
fix(rollup-plugin-workbox): revert back to non-ESM version of pretty-bytes

dlockhart:
As raised in #2212, this reverts #2208, which upgraded `pretty-bytes` from 5.6.0 to 6.1.0.

In that major version bump, `pretty-bytes` became an ESM-only package. Based on [their description](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), this package would also need to convert to ESM in order to adopt it, hence the revert.

Dependabot will likely attempt to upgrade this again, so not sure if you'd like me to add this to its ignore list?

### DIFF
--- a/.changeset/smooth-zoos-provide.md
+++ b/.changeset/smooth-zoos-provide.md
@@ -1,0 +1,5 @@
+---
+'rollup-plugin-workbox': patch
+---
+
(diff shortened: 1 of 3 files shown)


thepassle:
Thanks!

PR #2218, opened 2023-04-07
Prepare for mocha upgrade

koddsson:


koddsson:
I _think_ this is the change needed for #2177 but I figured I'd test it out a bit here before pushing to that branch.

koddsson:
Cherry picking the commits from the mocha upgrade branch to see if this is actually the fix.

PR #2228, opened 2023-04-20
Try running tests in parallel

koddsson:


koddsson:
Doesn't look like there's a lot of time saved compared to a recent build, like https://github.com/modernweb-dev/web/actions/runs/4709931672/jobs/8353291487. They are both around the same execution time, so I feel good about just closing this.

PR #2243, opened 2023-04-24
Move mocha options to config in package.json

koddsson:
I also moved from `--require ts-node/register` to `--loader ts-node/esm` which will make the ESM upgrade easier.


Westbrook (approved):
LGTM.

PR #2246, opened 2023-04-25
chore: lint versions --fix

### LINKED ISSUE #2245: Adopt `yarn version` in ci
yarn has a `version` command which can sync sibling packages in a monorepo

https://yarnpkg.com/features/release-workflow

Lately, I've noticed that having more than one version of dev-server or test-runner-core in a project leads to [annoying typescript errors](https://github.com/patternfly/patternfly-elements/actions/runs/4793707596/jobs/8526399489). This would be less of an issue if versions were aligned. In the example linked, the recently released 0.6.0 version of junit-reporter depends on test-runner 0.11.0 when the current version is 0.11.1.

It seems to me that using `yarn version` in CI (perhaps during the changesets versions command) would be a partial fix here.

bennypowers:
## What I did

1. fix the `lint-versions.js` script
1. add a `--fix` flag to `lint-versions.js`

## Notes to Reviewers
Seems to work, but I'm not sure how to handle the `newVersions` object that's returned from `compareVersions`, so this script might not be 100% correct

Closes #2245

### DIFF
--- a/scripts/lint-versions.js
+++ b/scripts/lint-versions.js
@@ -1,4 +1,4 @@
-/* eslint-disable */
+#!/usr/bin/env node
 const { readdirSync, existsSync, readFileSync } = require('fs');
 
 const getDirectories = source =>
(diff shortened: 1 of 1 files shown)


koddsson (approved):
Looks good to me 👍🏻

PR #2247, opened 2023-04-25
fix: reset coverage result converters between Chrome session runs

tmsns:
## What I did

@web/test-runner-coverage-v8 0.6.0 introduces a performance improvement when calculating code coverage by caching coverage converters between browser sessions. (coverage converters are creating by loading the original file from disk)
Unfortunately, converters will alter their own state when applying the coverage from a specific session. This means that next sessions will be "poisoned" by previous ones.

This PR will ensure converters are reset (to their initial state of zero coverage) before applying a new one.

Ideally we would use a clone of the converter before using. But as there is no stable, ootb, way to create a clone (without also knowing the internals of the object), this workaround will just reset the internally used variables.

/cc: @43081j @koddsson

### DIFF
--- a/.changeset/breezy-crabs-watch.md
+++ b/.changeset/breezy-crabs-watch.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-coverage-v8': patch
+---
+
+fix: reset coverage result converters between Chrome session runs
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "devDependencies": {
     "@changesets/cli": "^2.16.0",
     "@rollup/plugin-node-resolve": "^15.0.1",
+    "@rollup/plugin-terser": "^0.4.1",
     "@rollup/plugin-typescript": "^11.0.0",
     "@types/chai": "^4.2.22",
     "@types/mocha": "^10.0.1",
@@ -71,7 +72,6 @@
     "prettier-plugin-package": "^1.3.0",
     "rimraf": "^4.4.1",
     "rollup": "^3.15.0",
-    "@rollup/plugin-terser": "^0.4.1",
     "ts-node": "^10.4.0",
     "typescript": "~5.0.4"
   },
@@ -111,11 +111,6 @@
       "@typescript-eslint/ban-ts-comment": "off"
     }
   },
-  "mocha": {
-    "loader": "ts-node/esm",
-    "exit": true,
-    "retries": 3
-  },
   "lint-staged": {
     "*.js": [
       "eslint --fix",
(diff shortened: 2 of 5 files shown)


koddsson:
Looks like my formatting commit fixes unrelated stuff, not sure why it wasn't caught in earlier PRs but 🤷🏻‍♂️

koddsson:
@43081j I think you did the LRU cache changes. What do you think about this?

Westbrook (approved):
LGTM! Just tested it on my project locally and it's doing all the right things. Thanks for catching this 🙇🏼

tmsns:
hi @43081j

Although @westbrook might have jumped the gun a bit too with approving & merging, I still wanted to take the time to address your concerns:

- first off, I created a repo repo for the issue (in hindsight I should have created a proper issue first):
  https://github.com/tmsns/wtr-cached-coverage (sorry for the manual repo, I couldn't get an easy repro working with Stackblitz)
  Just run it using `npm install && npx wtr`. You'll see that the coverage fails with the current version of `@web/test-runner`, although the two test files cover both functions.
- _the point: collect coverage in one cached converter_
  This fix will not undo the caching improvements. For each file still, only one converter will be created (and cached, and reused). This means each file will only be loaded once.
  As far as I can see this is what the [PR ](https://github.com/modernweb-dev/web/pull/2169)was trying to achieve: "_reading the specified filePath from disk, which results in incredibly slow coverage computation_"

- The root of the issue is that the converter class `V8ToIstanbul` is not a stateless object, it mutates itself. When the converter is loaded, it does not offer a stateless function `convertCoveragetoIstanbul()`, instead it first needs to change its inner state using `applyCoverage()` and then allows this coverage to be converted to Istanbul format using `toIstanbul()`. 
  I also just found some more info on it in the original parsing of the coverage data by Lars: https://github.com/modernweb-dev/web/blob/4d773009b719c3852d52865ae960107fa63d8c62/packages/test-runner-core/src/coverage/getTestCoverage.ts#L186

- I agree, it's not great to use private API. I haven't found a better way to clone the loaded converter object though (in order to keep the perf improvements) :-/

- There might be a way to adjust `getTestCoverage` to take into account the accumulated istanbul data of the cached converter. But this will probably mean you need to cache all converters, always, and not use a LRU.

Hope this helps :-) And don't hesitate if you need more info!

43081j:
He did jump the gun. Any ts-ignore is pretty nasty

We should be contributing upstream to make a pubic API we can call to clear/reset the state really

When I'm back from vacation I will take a proper look

Westbrook:
I pushed forward because without this change the coverage numbers were broken and teams couldn't accurately adopt the new versions we published recently. We should absolutely contribute upstream, but needing to not have broken packages on `latest` was more important to me.

43081j:
👍 when I'm actually back I will open whatever PRs we need to do it properly

No rush, this will do for now

PR #2255, opened 2023-04-28
Node fetch updates

koddsson:
Update `node-fetch` to version 3.3.1.


thepassle:
we're on node 18 now, so do we even need node-fetch anymore?

koddsson:
> > we're on node 18 now, so do we even need node-fetch anymore?
> 
> Good questions! I thought we were just on node16?

In any case I'd love to just use fetch

PR #2257, opened 2023-05-01
Test on stable targeting Node releases

Westbrook:
## What I did

1. @43081j v20 is fully available, so it seemed like the right thing to move off of 19.


koddsson (approved):
Oh yeah

web-padawan:
Covered by #2601.

PR #2269, opened 2023-05-05
Use the new headless Chrome mode when launching puppeteer

koddsson:
I added `headless: 'new'` to the lunch options of puppeteer.

See https://github.com/puppeteer/puppeteer#default-runtime-settings

>[Chrome 112 launched a new Headless mode](https://developer.chrome.com/articles/new-headless/) that might cause some differences in behavior compared to the old Headless implementation.
In the future Puppeteer will start defaulting to new implementation.
We recommend you try it out before the switch:
>
>```ts
>const browser = await puppeteer.launch({headless: 'new'});
>```

### DIFF
--- a/.github/workflows/verify-node.yml
+++ b/.github/workflows/verify-node.yml
@@ -30,12 +30,6 @@ jobs:
       - name: Install Playwright dependencies
         run: npx playwright install-deps
 
-      - name: Install Chrome 113
-        run: |
-          VERSION_STRING="113.0.5672.63-1"
-          wget "https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_${VERSION_STRING}_amd64.deb"
-          sudo dpkg -i "google-chrome-stable_${VERSION_STRING}_amd64.deb"
-
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
--- a/.github/workflows/verify-node.yml
+++ b/.github/workflows/verify-node.yml
@@ -30,6 +30,12 @@ jobs:
       - name: Install Playwright dependencies
         run: npx playwright install-deps
 
+      - name: Install Chrome 113
+        run: |
+          VERSION_STRING="113.0.5672.63-1"
+          wget "https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_${VERSION_STRING}_amd64.deb"
+          sudo dpkg -i "google-chrome-stable_${VERSION_STRING}_amd64.deb"
+
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "start": "rocket start",
     "start:build": "node packages/dev-server/dist/bin.js --root-dir _site --open",
     "test": "yarn test:node && yarn test:browser && node scripts/workspaces-scripts-bin.mjs test:ci",
-    "test:browser": "node packages/test-runner/dist/bin.js \"packages/*/test-browser/**/*.test.{js,ts}\"",
+    "test:browser": "node packages/test-runner/dist/bin.js \"packages/*/test-browser/**/*.test.{js,ts}\" --coverage",
```
koddsson, line 34:
```suggestion
    "test:browser": "node packages/test-runner/dist/bin.js \"packages/*/test-browser/**/*.test.{js,ts}\"",
```

packages/test-runner-chrome/package.json
```diff
@@ -48,6 +48,7 @@
     "@web/test-runner-core": "^0.11.0",
     "@web/test-runner-coverage-v8": "^0.6.0",
     "chrome-launcher": "^0.15.0",
+    "puppeteer": "^20.1.1",
```
koddsson, line 51:
```suggestion
```

koddsson:
I think this is a bug with the new headless chrome in puppeteer. I've opened up a issue here: https://github.com/puppeteer/puppeteer/issues/10350

.github/workflows/verify-node.yml
```diff
@@ -30,6 +30,12 @@ jobs:
       - name: Install Playwright dependencies
         run: npx playwright install-deps
 
+      - name: Install Chrome 113
+        run: |
+          VERSION_STRING="113.0.5672.63-1"
+          wget "https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_${VERSION_STRING}_amd64.deb"
+          sudo dpkg -i "google-chrome-stable_${VERSION_STRING}_amd64.deb"
+
```
koddsson, line 38: I think Chrome has been updated in actions since I opened up this PR so I'm gonna see if we can remove this.

packages/test-runner-chrome/package.json
```diff
@@ -48,6 +48,7 @@
     "@web/test-runner-core": "^0.11.0",
     "@web/test-runner-coverage-v8": "^0.7.0",
     "chrome-launcher": "^0.15.0",
+    "puppeteer": "19.10",
```
thepassle, line 51: was this dep added intentionally? I think test-runner-chrome uses puppeteer-core for a specific reason (to avoid depending on puppeteer directly), can you check this with lars?

thepassle: iirc puppeteer-core depends on using the chrome version installed on the users machine, whereas puppeteer ships it with the package or something like that

koddsson: Ah! I think this is accidental then!

thepassle (approved):

PR #2277, opened 2023-05-11
fix: Export types from config-loader

Matsuuu:
## What I did

1. Added an export for types for the config-loader package

I was about to start using this package but had my typescript setup not finding the type declarations for it. 

Turns out you have to add reference to the types as well, if you're declaring e.g. the import and require exports.

Source: https://webpack.js.org/guides/package-exports/#conditions

Should fix the "could not find a declaration file for module types" errors for this package

Not sure if there's other packages in this repo that could use this touch up too


Matsuuu:
I think this should cause a patch level version bump to the package, I'll try n figure out how to set that up

Matsuuu:
I'm like 100% that the tests aren't failing due to this change :D Could this get a look/merge?

thepassle:
@Matsuuu can you bring this branch up to date with the latest changes on master? that should make CI pass

Matsuuu:
@thepassle Sure can. Gimme a couple of hours and I'll push the latest

Matsuuu:
Welp. Time escaped me last week. Did it now, let's see if the CI runs nicely

PR #2278, opened 2023-05-16
fix: do not swallow tests

mokimo:
## What I did
Fix an issue that leads to failing tests to be potentially swallowed from the defaultReporter `import { defaultReporter } from '@web/test-runner';`

It's a very tedious issue to debug and might fail tests to fail on CI systems as well as locally.

## Reproduce
```
import { readFile } from '@web/test-runner-commands';

describe('reading-time estimate', () => {
  it('supports longer articles with 2+ min read time', async () => {
    document.body.innerHTML = await readFile({ path: './mocks/just-return-empty-body.html' });
    expect(true).to.be.true;
  });

  describe('this describe fails', async () => {
    document.body.innerHTML = await readFile({ path: './mocks/just-return-empty-body-2.html' });

    it('failing test', () => {
      expect(true).to.be.true;
    });
  });
});
```

## Output before
Test fails without any logging
Screenshot 2023-05-16 at 20 16 55

## Output after
Provide context and *which* test is failing for further debugging
Screenshot 2023-05-16 at 20 19 03

### DIFF
--- a/packages/test-runner/src/reporter/reportTestsErrors.ts
+++ b/packages/test-runner/src/reporter/reportTestsErrors.ts
@@ -75,15 +75,21 @@ export function reportTestsErrors(
     if (session.testResults) {
       const flattenedTests = getFlattenedTestResults(session.testResults);
       for (const test of flattenedTests) {
-        if (test.error) {
+        if (!test.passed && !test.skipped) {
           let testErrorsForBrowser = testErrorsPerBrowser.get(test.name);
           if (!testErrorsForBrowser) {
             testErrorsForBrowser = new Map<string, TestResultError>();
             testErrorsPerBrowser.set(test.name, testErrorsForBrowser);
           }
-          if (test.error) {
-            testErrorsForBrowser.set(session.browser.name, test.error);
-          }
+
+          test.error = test.error || { 
+            name: 'Unknown error', 
+            message: 'Test failed without any errors. One reason could be that you have code outside of your test.',
+            expected: "",
+            actual: "",
+            stack: "",
+          };
+          testErrorsForBrowser.set(session.browser.name, test.error);
         }
       }
     }


packages/test-runner/src/reporter/reportTestsErrors.ts
```diff
@@ -75,15 +75,21 @@ export function reportTestsErrors(
...
-          if (test.error) {
-            testErrorsForBrowser.set(session.browser.name, test.error);
-          }
+
+          test.error = test.error || { 
+            name: 'Unknown error', 
+            message: 'Test failed without any errors. One reason could be that you have code outside of your test.',
+            expected: "",
+            actual: "",
+            stack: "",
```
koddsson, line 90: It would be great to preserve the stack if possible. Is that something you think we can do here? I'd like to give the user more information.

mokimo: Nope sadly not - we can just hint which test failed. We only get `{ name: 'failing test', passed: false }` as context.

Compared to actually getting the error that we can forward as it normally would be the case, e.g.
```js
{
  duration: 1,
  error: {
    message: 'some error within a test',
    name: 'Error',
    stack: '  at qy.<anonymous> (test/blocks/global-navigation/foo.test.js:7:11)'
  },
  name: 'minimal reproducible example > this test works',
  passed: false
}
```

joekukish:
Is there an update on this? It makes it very hard to debug since we don't know where it failed.

mokimo:
Hey @joekukish we mitigated this in our project by adding a custom reporter https://github.com/adobecom/milo/blob/stage/web-test-runner.config.mjs#L7-L19

It would be better if we can get this in more generally of course.

mokimo:
Heya @talentlessguy or @bashmish anything I can do to push this forward? IMO this is pretty valuable and causes a bunch of dev-hours to be lost on strange failures!

bashmish:
@mokimo I think the solutions is nice

we need some tests for it, also to understand the exact use case, can you please add them?
and can you also change the `minor` to `patch`

PR #2297, opened 2023-06-08
feat: update workbox v7

thepassle:
## What I did

1. Update workbox to v7
2. Removed the `mode` option in favour of just always bundling/removing `process.env` for the service worker when using `injectManifest`


koddsson (approved):
Looks good to me :)

PR #2307, opened 2023-06-19
Reduce the Playwright dependency

thewilkybarkid:
I use `@playwright/test` and `@web/test-runner-playwright` in a project. They both have a pinned version of the `playwright-core` library (the latter has it through the `playwright` dependency). I end up with multiple versions of the `playwright-core` library installed whenever Dependabot tries to update the former. When I [install the browsers in my `web-test-runner` job](https://github.com/PREreview/prereview.org/blob/96927194aba4070df65223c3f8dbd086024634e0/.github/workflows/ci.yml#L247-L254) it picks up the wrong browser versions and fails (`npx playwright` running the command from `@playwright/test` rather than `playwright`).

This change reduces the dependency here to just `playwright-core`, with a relaxed constraint that should mean they use the same version.

### DIFF
--- a/.github/workflows/verify-node.yml
+++ b/.github/workflows/verify-node.yml
@@ -28,7 +28,7 @@ jobs:
           cache: yarn
 
       - name: Install Playwright dependencies
-        run: npx playwright install-deps
+        run: npx playwright-core install-deps
 
       - name: Install dependencies
         run: yarn --frozen-lockfile
--- a/.github/workflows/verify-windows.yml
+++ b/.github/workflows/verify-windows.yml
@@ -23,7 +23,7 @@ jobs:
           cache: 'yarn'
 
       - name: Install Playwright dependencies
-        run: npx playwright install-deps
+        run: npx playwright-core install-deps
 
       - name: Install dependencies
         run: yarn --frozen-lockfile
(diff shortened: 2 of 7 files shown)


Westbrook:
Playwright specifically says to [install `playwright`](https://playwright.dev/docs/intro#installing-playwright). Is there a path to resolving your issue without changing the installed package? Possibly opening the version string, or similar?

thewilkybarkid:
It seems quite confusing. The docs mention starting with `npm init playwright@latest` which sets up a local project that depends on `@playwright/test`. (It seems like actually uses the `create-playwright` library to set it up though.)

I've just been through the Playwright docs and they don't mention using `playwright` (only ever `@playwright/test`). The API does mention it at https://playwright.dev/docs/api/class-playwright. `playwright-core` is never mentioned.

Both `@playwright/test` and `playwright` are pinned to a specific version of `playwright-core`. They both try and produce a binary called `playwright` and `playwright-core`. When installing _both_ `@playwright/test` and `playwright` locally, the `playwright` binary seems to always point to the `@playwright/test` library, which may not have the same version as the `playwright` library. It seems that using `@playwright/test` and `playwright` in the same doesn't work.

I'll open an issue on the Playwright repo to see what is expected here.

thewilkybarkid:
Playwright 1.38 has changed its setup so `@playwright/test` now depends on `playwright`. As there's an open constraint here npm/Dependabot won't now install multiple versions; I'll close this.

PR #2344, opened 2023-07-02
chore: move to npm workspaces

daKmoR:
## What I did

1. Use npm workspaces instead of yarn workspaces
2. Replace all yarn usage with npm (run) in ci, package.json and docs

PS: seems the `serialize/deserialize` seem to fail on windows for quite some time - not gonna fix it here


bennypowers:
Nice move

Maybe we'll consider adopting wireit, next

.github/workflows/canary.yml
```diff
@@ -25,27 +25,27 @@ jobs:
           FORCE_COLOR: 0
         with:
           node-version: '18'
-          cache: 'yarn'
+          cache: 'npm'
           registry-url: 'https://registry.npmjs.org'
 
       - name: Install Dependencies
-        run: yarn --frozen-lockfile
+        run: npm ci
```
bennypowers, line 32:
```suggestion
        run: npm ci --prefer-offline
```

.github/workflows/canary.yml
```diff
@@ -25,27 +25,27 @@ jobs:
...
-        run: yarn build
+        run: npm run build
 
       - name: Build for production
-        run: yarn build:production
+        run: npm run build:production
 
       - name: Version canary release
-        run: yarn changeset version --snapshot canary
+        run: npm run changeset version --snapshot canary
```
bennypowers, line 41:
```suggestion
        run: npx changeset version --snapshot canary
```

bennypowers: Etc

.github/workflows/canary.yml
bennypowers: And elsewhere

bennypowers (changes requested):

.github/workflows/canary.yml
daKmoR: what is that for? just did a quick google and wouldn't that also require to specify the cache location? 🤔

I think it's best to keep `npm ci` for now and then in a separate PR check the difference using `--prefer-offline` as I'm curious how big of a difference it would be 🤗

.github/workflows/canary.yml
daKmoR: changed

.github/workflows/canary.yml
bennypowers: Speeds up ci a bit. Not critical

bennypowers (approved):

daKmoR:
yeah wireit to revamp the tsc build step would imho the best next step 👍

PR #2353, opened 2023-07-03
chore(mocks): generate and publish types + setup wireit

daKmoR:
## What I did

1. Basic Setup for Wireit
2. Add `all` and `head` to http
2. Generate basic types for `@web/mocks` using wireit


packages/mocks/http.js
```diff
@@ -6,22 +6,68 @@ const VERSION = Symbol.for('msw-integration-layer::version');
...
+   */
   options: (endpoint, handler) => {
     return { [VERSION]: '1.x.x', method: 'options', endpoint, handler };
   },
+  /**
+   * @param {string} endpoint
+   * @param {import('./types.js').handler} handler
+   * @returns {import('./types.js').Mock}
+   */
+  all: (endpoint, handler) => {
```
thepassle, line 62: It was a conscious decision to _not_ add `all`, because its magic-y. Our current mocking solution has tons of magic, and we're trying to get rid of magic and have people be explicit in what they're trying to mock; so also their http methods.

packages/mocks/http.js
```diff
@@ -6,22 +6,68 @@ const VERSION = Symbol.for('msw-integration-layer::version');
...
+    return { [VERSION]: '1.x.x', method: 'all', endpoint, handler };
+  },
+  /**
+   * @param {string} endpoint
+   * @param {import('./types.js').handler} handler
+   * @returns {import('./types.js').Mock}
+   */
+  head: (endpoint, handler) => {
```
thepassle, line 70: I dont think MSW supports `head`, so this will cause an error

thepassle (changes requested):

thepassle:
2 comments, types are looking good though 👍

packages/mocks/http.js
daKmoR: makes sense 👍

packages/mocks/http.js
daKmoR: it's in msw's rest
https://github.com/mswjs/msw/blob/v0.0.0-fetch.rc-15/src/core/rest.ts#L35

packages/mocks/http.js
daKmoR: removed

packages/mocks/http.js
thepassle: Interesting, its not documented on this page: https://mswjs.io/docs/api/rest

Ok, lets keep it 👍

thepassle: Make sure to add `'head'` here: https://github.com/modernweb-dev/web/pull/2353/files#diff-dfdc098b54356afdd499c56eec03801b27a275a1436dbe8cf6d77eb468ae70a3R3

packages/mocks/registerMockRoutes.js
```diff
@@ -1,20 +1,13 @@
 import { rest } from 'msw';
 
-/**
- * @typedef {Object} Mock
- * @property {string} method
- * @property {string} endpoint
- * @property {({request}: {request: Request}) => Response | Promise<Response>} handler
- */
-
 const SUPPORTED_METHODS = ['get', 'post', 'put', 'patch', 'delete', 'options'];
```
thepassle, line 3:
```suggestion
const SUPPORTED_METHODS = ['get', 'post', 'put', 'head', 'patch', 'delete', 'options'];
```

thepassle:
LGTM

packages/mocks/http.js
daKmoR: good catch - added 👍

thepassle (approved):

PR #2365, opened 2023-07-05
fix(dev-server-rollup): await buildStart in serverStart

bashmish:
## What I did

1. make rollup adapter wait for `buildStart` in `serverStart`
2. add a test for it

Not sure why it wasn't the case already, I think this is expected to work like that.
I ran into a bug here because I do [pre-bundling for the new Storybook builder with @web/dev-server, using wrapped rollup plugins](https://github.com/bashmish/storybook-builder-wds/pull/2/files#diff-6cbe122f09cd8872e3496aab2d5335a29f2f6571c9a85552e6b0c860defa94a2) and this code breaks when pre-bundling takes longer when the first request for a pre-bundled module (usually with auto-open of the browser).

When I run a static build with rollup alone, it works as expected and awaits the `buildStart` before doing the resolutions and proceeding with the build, so this also aligns this behavior which I think is good.

### DIFF
--- a/packages/dev-server-rollup/test/node/unit.test.ts
+++ b/packages/dev-server-rollup/test/node/unit.test.ts
@@ -2,10 +2,47 @@ import { Plugin as RollupPlugin, AcornNode } from 'rollup';
 import { expect } from 'chai';
 import path from 'path';
 
-import { createTestServer, fetchText, expectIncludes } from './test-helpers';
+import { createTestServer, fetchText, expectIncludes, timeout } from './test-helpers';
 import { fromRollup } from '../../src/index';
 
 describe('@web/dev-server-rollup', () => {
+  describe('buildStart', () => {
+    it('awaits buildStart when starting a server', async () => {
+      let resolveBuildStartCalled: undefined | ((value: void) => void);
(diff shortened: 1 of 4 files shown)


packages/dev-server-rollup/test/node/unit.test.ts
```diff
@@ -2,10 +2,47 @@ import { Plugin as RollupPlugin, AcornNode } from 'rollup';
...
+          resolveBuildStart();
+        }
+        const res2 = await Promise.race([serverPromise, timeout(0, 'timeout')]);
+        expect(res2).not.to.equal('timeout');
+      } finally {
+        const { server } = await serverPromise;
+        server.stop();
+      }
+    });
+  });
```
bashmish, line 44: this is pretty hard to test, I was a few time thinking maybe it's best to ditch this completely, also given it's disproportionate to the complexity of the logic and potential bugs in it

thepassle: I agree this is a pretty complex test just for sake of adding a missing `await`

bashmish: good, I'll remove it

bashmish: done

PR #2366, opened 2023-07-06
fix: add workaround for headless issue

tmsns:
## What I did

This PR will add a workaround for an issue in the new headless mode of Puppeteer https://github.com/puppeteer/puppeteer/issues/10350.

The PR implements the proposal in the issue by patching the functions that are timing out (`window.requestIdleCallback` and `window.requestAnimationFrame`)

I'm not to sure yet what the best approach is to add tests for this.

### DIFF
--- a/.changeset/itchy-socks-rule.md
+++ b/.changeset/itchy-socks-rule.md
@@ -1,0 +1,7 @@
+---
+'@web/test-runner-chrome': patch
+---
+
+fix: add workaround for headless issue
+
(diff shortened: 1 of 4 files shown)


koddsson (approved):
I think this is a great patch that matches the issues that we've seen with the new chrome headless mode and the suggested solution by the puppeteer team!

:shipit:

koddsson:
I think this Windows build has been broken for a bit so I feel fine about merging and coming back to fix the Windows build.

PR #2370, opened 2023-07-07
chore: reenable all tests and update snapshots

daKmoR:
## What I did

1.

### DIFF
--- a/.github/workflows/verify-windows.yml
+++ b/.github/workflows/verify-windows.yml
@@ -4,6 +4,7 @@ on: pull_request
 
 jobs:
   verify-windows:
+    timeout-minutes: 30
     name: Verify windows
     runs-on: windows-2022
     steps:
--- a/packages/rollup-plugin-polyfills-loader/test/snapshots/flattened.html
+++ b/packages/rollup-plugin-polyfills-loader/test/snapshots/flattened.html
@@ -1,8 +1,7 @@
 <html><head>
 
-<link rel="preload" href="./entrypoint-a.js" as="script" crossorigin="anonymous" />
-<link rel="preload" href="shared-4322372e.js" as="script" crossorigin="anonymous" />
 <link rel="preload" href="./entrypoint-b.js" as="script" crossorigin="anonymous" />
+<link rel="preload" href="./entrypoint-a.js" as="script" crossorigin="anonymous" />
 </head><body><script>(function () {
   function loadScript(src, type, attributes) {
     return new Promise(function (resolve) {
@@ -34,9 +33,9 @@
   }
   function loadFiles() {
     [function () {
-      return loadScript('./entrypoint-a.js', 'module', []);
-    }, function () {
       return loadScript('./entrypoint-b.js', 'module', []);
+    }, function () {
+      return loadScript('./entrypoint-a.js', 'module', []);
     }].reduce(function (a, c) {
       return a.then(c);
     }, Promise.resolve());
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   release:
+    timeout-minutes: 30
     # Prevents changesets action from creating a PR on forks
     if: github.repository == 'modernweb-dev/web'
     name: Pre-release
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -38,6 +38,9 @@ jobs:
       - name: Build packages
         run: npm run build
 
+      - name: Build types for js packages
+        run: npm run types
+
       - name: Build for production
         run: npm run build:production
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   release:
+    timeout-minutes: 30
     # Prevents changesets action from creating a PR on forks
     if: github.repository == 'modernweb-dev/web'
     name: Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,6 +37,9 @@ jobs:
       - name: Build packages
         run: npm run build
 
+      - name: Build types for js packages
+        run: npm run types
+
       - name: Build for production
         run: npm run build:production
 
--- a/.github/workflows/verify-browserstack.yml
+++ b/.github/workflows/verify-browserstack.yml
@@ -4,6 +4,7 @@ on: pull_request
 
 jobs:
   verify-browserstack:
+    timeout-minutes: 30
     name: Verify Browserstack
 
     # Run only from the original repository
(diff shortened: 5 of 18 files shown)


packages/rollup-plugin-polyfills-loader/test/snapshots/flattened.html
```diff
@@ -1,8 +1,8 @@
 <html><head>
 
-<link rel="preload" href="./entrypoint-a.js" as="script" crossorigin="anonymous" />
-<link rel="preload" href="shared-4322372e.js" as="script" crossorigin="anonymous" />
 <link rel="preload" href="./entrypoint-b.js" as="script" crossorigin="anonymous" />
+<link rel="preload" href="entrypoint-a.js" as="script" crossorigin="anonymous" />
+<link rel="preload" href="./entrypoint-a.js" as="script" crossorigin="anonymous" />
```
thepassle, line 5: This seems like a bug. We shouldnt add duplicate links for the same filepath. Seems like something is going wrong with the handling of the paths.

thepassle: I think we need to change this to normalize the paths and make them consistent:
https://github.com/modernweb-dev/web/blob/718e1524a6e13e04f22f2854657bf3bc00f644a5/packages/rollup-plugin-polyfills-loader/src/rollupPluginPolyfillsLoader.ts#L70-L79

Something like:
```js
        function normalize(path: string) {
          if (path.startsWith('../')) {
            return path;
          } else if (path.startsWith('./')) {
            return path;
          } else if (path.startsWith('/')) {
            return '.' + path;
          } else {
            return './' + path;
          }
        }

        for (const entrypoint of entrypoints) {
          const importPath = normalize(path.posix.relative('', entrypoint.importPath));
          preloaded.push(importPath);

          // js files (incl. chunks) will always be in the root directory
          const pathToRoot = path.posix.dirname(importPath);
          for (const chunkPath of entrypoint.chunk.imports) {
            const relativeChunkPath = normalize(path.posix.join(pathToRoot, chunkPath));
            preloaded.push(relativeChunkPath);
          }
        }

        preloaded = [...new Set(preloaded)];
```

thepassle:
This PR seems to mostly fix the failing tests on master, but now the tests that were supposed to be fixed by https://github.com/modernweb-dev/web/pull/2366 (by @tmsns ) are broken again, seemingly without any apparent reason. @tmsns would you mind taking a look at this?

image

koddsson:
If we don't come up with a quick solution to this, we can look into disabling these tests again temporarily in order to get all the tests running again on `master`.

tmsns:
@thepassle here is a proposal to fix the "random" timeouts. Locally it works. Let's see if it works correctly in a pipeline.

koddsson:
@daKmoR; wanna pull `master` into this branch after we merged #2371? That might have fixed the failing tests 😄

packages/rollup-plugin-polyfills-loader/test/snapshots/flattened.html
```diff
@@ -34,9 +34,9 @@
   }
   function loadFiles() {
     [function () {
-      return loadScript('./entrypoint-a.js', 'module', []);
-    }, function () {
       return loadScript('./entrypoint-b.js', 'module', []);
+    }, function () {
+      return loadScript('./entrypoint-a.js', 'module', []);
```
daKmoR, line 39: hmmm I'm just wondering is the order here maybe wrong?

as it's now `entrypoint-b` and then `entrypoint-a` 🤔

koddsson:
This is looking green except for the windows tests hanging, which I believe is a separate issue.

tmsns:
Took a first look at the Windows pipeline. Seemed to be hanging once wireit starts the `tsc —build —pretty` command in the new mocks project. Not sure whether wireit can’t start it or typescript can’t complete it.

packages/rollup-plugin-polyfills-loader/test/snapshots/flattened.html
daKmoR: that worked perfectly 👍

.github/workflows/verify-windows.yml
```diff
@@ -4,6 +4,7 @@ on: pull_request
 
 jobs:
   verify-windows:
+    timeout-minutes: 30
```
koddsson, line 7: I like that we have a timeout now. I'd argue that we can probably make it even lower but maybe in a future PR.

koddsson (approved):
This looks pretty good to me! I'm not sure on the snapshot tests but I'm not sure if the order matters.

thepassle:
So is this an issue with wireit? It seems unlikely that its an issue with TSC, right? Should we disable wireit for now, and just run TSC directly? What do you think @daKmoR ?

image

thepassle:
Not really sure why, but I'm still seeing wireit running in the pipeline, so it's still hanging on **Verify Windows**

image

tmsns:
`@web/mocks` itself is also using wireit 🙈. It’s not only on top level.
https://github.com/modernweb-dev/web/blob/6544b78fc2e5489ffa531b373a3b349cb60adf85/packages/mocks/package.json#L41

koddsson:
Now this is happening without wireit! 🤔 

```
Run npm run build
  npm run build
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    WIREIT_CACHE: github
    ACTIONS_CACHE_URL: https://artifactcache.actions.githubusercontent.com/ccnqSCum0KFcxsiFoFI7NcNXpjM[3](https://github.com/modernweb-dev/web/actions/runs/5523695137/jobs/10074963603?pr=2370#step:8:3)eDUNQ90mJ1DwqHCbmkBIOA/
    ACTIONS_RUNTIME_TOKEN: ***

> build
> rimraf --glob packages/*/tsconfig.tsbuildinfo && tsc --build && npm run types


> types
> cd packages/mocks && npm run types


> @web/mocks@0.1.[5](https://github.com/modernweb-dev/web/actions/runs/5523695137/jobs/10074963603?pr=2370#step:8:5) types
> tsc --build --pretty

Terminate batch job (Y/N)? 
Error: The operation was canceled.
```

koddsson:
Maybe we can enable the [`diagnostics`](https://www.typescriptlang.org/tsconfig#diagnostics) flag to get a better idea of what's going on?

I googled around and someone talked about this happening on machines with lower memory and I did discover that the Windows machines have 7GB memory while macOS has 14GB which might explain why it would only hang on Windows machines.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

daKmoR:
ok, I now specifically NOT do the types action on windows (there is no need to do it anyways)

still there is one failing test... but I will leave this for another PR

PR #2371, opened 2023-07-10
fix(test-runner-chrome): add mutex when bringing tabs to front

tmsns:
## What I did

This is a potential fix for the timeout issues when running test code in parallel that is requiring to be run in an active tab.


koddsson (approved):
Looks good to me 👍🏻

PR #2373, opened 2023-07-10
feat(dev-server-core): support middleware mode

bashmish:
## What I did

1. add `middlewareMode` option to dev server config, make `hostname` and `port` optional in types
2. implement middleware mode to prevent starting an actual HTTP/S server
3. make `server` optional on `DevServer` which also has an impact on Plugin API
4. add test and docs

## Breaking change

Theoretically it's a breaking change for Plugin API since now `serverStart` hook might not have a param `server`.
It won't impact the codebase immediately since you need to first activate `middlewareMode` to actually break stuff if any plugin depends on `server` existence, but type wise it might break compilation of such plugins right away.

As agreed with @thepassle we decided to take an easy path and not create a SemVer mess with a minor update as the impact of this is too small.

### DIFF
--- a/.changeset/fluffy-snails-share.md
+++ b/.changeset/fluffy-snails-share.md
@@ -7,4 +7,4 @@ support middleware mode
 BREAKING CHANGE
 
 Theoretically it's a breaking change for Plugin API since now `serverStart` hook might not have a param `server`.
-It won't impact the codebase immediately since you need to first activate `middlewareMode` to actually break stuff if any plugin depends on `server` existence, but type wise it might break compilation of such plugins right away. We decided to take an easy path and not create a SemVer mess with a minor update as the impact of this is too small and the risk is acceptable.
+The breaking change shouldnt impact your codebase immediately since you need to first activate `middlewareMode` to actually break stuff if any plugin depends on the existence of the `server` arg that gets passed to the plugin, but type wise it might break compilation of such plugins right away. Considering that `@web/dev-server` is still on semver 0.x.x, and since the impact of this breaking change should be very minimal, we decided to make the breaking change in this patch version.
--- a/.changeset/fluffy-snails-share.md
+++ b/.changeset/fluffy-snails-share.md
@@ -1,0 +1,10 @@
+---
+'@web/dev-server-core': patch
+---
+
+support middleware mode
+
+BREAKING CHANGE
+
+Theoretically it's a breaking change for Plugin API since now `serverStart` hook might not have a param `server`.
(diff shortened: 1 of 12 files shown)


bashmish:
Not sure why this error

```
TSError: ⨯ Unable to compile TypeScript:
Error: packages/dev-server-hmr/test/browser.test.ts(49,41): error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.
```

but maybe related to the package-lock.json update. I just ran `npm i`, nothing special.

docs/docs/dev-server/node-api.md
```diff
@@ -152,6 +152,29 @@ wss.on('connection', ws => {
 });
 ```
 
+## Middleware mode
+
+If you need to connect to an existing running web server with a compatible node middleware API (e.g. `express`), you can use a middleware mode.
```
thepassle, line 157:
```suggestion
If you need to connect to an existing running web server with a compatible node middleware API (e.g. `express`), you can use `@web/dev-server` in middleware mode.
```

thepassle:
One comment, otherwise LGTM :)

docs/docs/dev-server/node-api.md
bashmish: fixed!

bashmish:
@thepassle I added a note about potential breaking change in the changeset

.changeset/fluffy-snails-share.md
```diff
@@ -0,0 +1,10 @@
+---
+'@web/dev-server-core': patch
+---
+
+support middleware mode
+
+BREAKING CHANGE
+
+Theoretically it's a breaking change for Plugin API since now `serverStart` hook might not have a param `server`.
+It won't impact the codebase immediately since you need to first activate `middlewareMode` to actually break stuff if any plugin depends on `server` existence, but type wise it might break compilation of such plugins right away. We decided to take an easy path and not create a SemVer mess with a minor update as the impact of this is too small and the risk is acceptable.
```
thepassle, line 10:
```suggestion
The breaking change shouldnt impact your codebase immediately since you need to first activate `middlewareMode` to actually break stuff if any plugin depends on the existence of the `server` arg that gets passed to the plugin, but type wise it might break compilation of such plugins right away. Considering that `@web/dev-server` is still on semver 0.x.x, and since the impact of this breaking change should be very minimal, we decided to make the breaking change in this patch version.
```

thepassle (approved):

thepassle:
Made a small edit on the changeset, but LGTM 👍

Issue #2282, opened 2023-05-24
Update to pupeteer 20.x to resolve download behind proxy issues


wapophis:
We're having issues trying to download the chrome-browser with proper proxy configuration. Seems like pupeteer 19.x have issues working behind proxy which affect us, with download code error, because this bugs in pupeteer 19.x. After upgrading pupeteer to 20.x the bugs are gone, but our projects has dependency with @web/test-runner-puppeteer which resolve to 19.x.

Westbrook:
We'd love to accept a PR in this area. Could we interest you in a contribution?

Issue #2262, opened 2023-05-04
[test-runner-commands] `reset-mouse` command is broken with newer versions of puppeteer


melink14:
This is broken since [v19.10.1](https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-v19.10.1) (due to https://github.com/puppeteer/puppeteer/pull/10021)  and unfortunately wasn't marked as a breaking change and so is in scope of the current version designation of the puppeteer package.

The breakage is caused by these lines:
https://github.com/modernweb-dev/web/blob/e0d362fcdbedf6a72a6c78ebc2e3a3cffa922dbc/packages/test-runner-commands/src/sendMousePlugin.ts#L170

In latest versions of puppeteer, it's an error to call `up` if the mouse button is not actually pressed.

Users can implement their own reset commands to compensate (based on actual use) but I think the only easy common reset implementation would be to use `try...catch` to prevent bubbling expected errors.

dlockhart:
I'm seeing this as well. Another alternative solution perhaps: track whether the 3 mouse buttons are up or down and then `reset-mouse` would only call `mouse.up` for the ones that require it.

web-padawan:
Update: the last working version is `19.9.0` and the change was introduced in [19.9.1](https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v19.9.1).

melink14:
The reset command was added in https://github.com/puppeteer/puppeteer/pull/10340 and it looks like it's been released since [puppeteer-core: v20.7.3](https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v20.7.3) and [browsers: v1.4.2](https://github.com/puppeteer/puppeteer/releases/tag/browsers-v1.4.2) so I think this problem could be fixed with another update to those deps and then chaning the commands code to use `reset` instead of manually calling the mouse commands.

I guess since the deps are new major versions (some sort of breaking change) then it would mean the modernweb package would also need a new major version?

web-padawan:
> it would mean the modernweb package would also need a new major version?

Yes, we probably should update to these versions and use `await page.mouse.reset()` directly.
This would mean releasing `@web/test-runner-commands` version `0.8.0` due to major bump.

PR #2375, opened 2023-07-12
feat/various fixes

thepassle:
## What I did

1. Update puppeteer to 20.0.0, fixes https://github.com/modernweb-dev/web/issues/2282
2. Use puppeteer's new `page.mouse.reset()` in `sendMousePlugin`, fixes https://github.com/modernweb-dev/web/issues/2262
3. Use `development` export condition by default


thepassle:
@tmsns FYI the focus tests are still failing, I'm skipping them for now

image

daKmoR:
needs a rebase for the conflict... but rest LGTM 👍

koddsson:
Does this also need a changeset, so it's released?

thepassle:
Huh, weird, pretty sure I added a changeset. Maybe I fudged something. I'll re-add the changeset.

bearfriend:
The switch to `development` exports caused some minor issues for us. Easy enough to work around but I was just curious what the reasoning was for this switch. It doesn't seem like it's linked to any issues like the other fixes.

bearfriend:
Sure. We have some `lit` components that have their own `_initialize` methods. Those of course extend `LitElement` (which itself extend `ReactiveElement`), and eventually tries to initialize itself with `this._initialize()`. With `development` exports, though, this doesn't get renamed, so they collide and `ReactiveElement` ends up calling _my_ `_initialize` method instead.

Using `development` exports by default seems like there is no limit to the things that could collide if they're being extended, and the devs of those packages would/should fully expect that not to be an issue, so that could start happening even with minor or patch releases.

thepassle:
Cc @justinfagnani this feedback may be of interest to you

justinfagnani:
Sounds like this highlighted some real problems in user code!

fwiw, I tried to change `_initialize` to avoid collisions: https://github.com/lit/lit/pull/3895 but we had to revert that specific PR. I can take another shot with a different approach, which will alleviate this issue.

bearfriend:
I suppose I'm still a bit confused as to why we want to default to `development`. With no other setup, nothing else would resolve to `development` exports. The `default` export should always be the ~transpiled version (and in lit it is), no? To default the conditions to anything that isn't `default` seems odd.

Perhaps I'm misunderstanding something, though.

bearfriend:
Also, I'm not sure why this would be considered a problem with user code, since `_initialize` is not a documented public method of the class.

augustjk:
> I suppose I'm still a bit confused as to why we want to default to `development`.

I'm for development condition being default as this is web *dev* server, after all. Some libraries like Lit have extra checks and warnings that only show in dev mode, it's nice to be able to see those while developing. I feel like it probably should default to the production build for testing though.

bearfriend:
> as this is web dev server, after all. ...  I feel like it probably should default to the production build for testing though.

Thanks for pointing this out. I _was_ thinking about this just from the testing side. I agree with all of that.

PR #2390, opened 2023-07-17
Added types to exports in package.json

Joel-Levi:
## What I did

This package has types but it does not provide them, this commit adds the missing types to the exports in the package.json.


koddsson:
Looks good to me

PR #2394, opened 2023-07-19
set node version to 18

koddsson:
## What I did

1. I set `engines.node` version to at least 18.

I figured since we aren't even testing on node16 that we don't actually support it.


koddsson:
Would this be all breaking changes? 🤔

bashmish:
> Would this be all breaking changes? 🤔

Theoretically. But for pragmatic reasons I'd keep it a patch

bashmish (approved):
Makes sense to me indeed!

koddsson:
> Makes sense to me indeed!

Nice! I'll update the versions to be a minor bump and if no one else chimes in soon I'll release this 👍🏻

Issue #2397, opened 2023-07-22
[rollup-plugin-html] does not extract assets within `<template>` tags


WillsonSmith:
When processing HTML the `rollup-plugin-html` plugin skips extracting assets within `<template>` tags.

I've made myself a small build system that uses `@lit-labs/ssr` to generate HTML pages and noticed that the image assets of a components were not being extracted. I've generated this minimal page as an example. The `/public/img/me.png` asset is not being extracted or included in the build. 

I've also reproduced using a css asset. 

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>My app</title>
    <meta name="viewport" content="width=device-width, initial-scale=1">
    </head>
  <body>
    <!--lit-part S7bmmAfnsS0=-->
  <template>
    <img src="/public/img/me.png" />
  </template>
<!--/lit-part-->
  </body>
</html>
```

PR #2398, opened 2023-07-22
fix(parse5-utils): add support for templates

WillsonSmith:
Proposed fix for https://github.com/modernweb-dev/web/issues/2397

parse5-utils does not currently handle tags within a `<template>` tag, this prevents `rollup-plugin-html` from extracting assets that are within a template.

## What I did

1. Add a condition to `findNodes` that will get child nodes of a `<template>` tag
2. Add a test to ensure the change functions as expected

### DIFF
--- a/.changeset/soft-meals-smash.md
+++ b/.changeset/soft-meals-smash.md
@@ -1,0 +1,5 @@
+---
+'@web/parse5-utils': patch
+---
+
+Adds support for traversing <template> elements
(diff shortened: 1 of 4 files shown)


koddsson:
@43081j; You've been doing parse5 stuff lately? What do you think?

43081j:
Makes sense to me!

I've been building an official parse5 tools package but we can't move mw over yet since it doesn't build esm. 

Once we can use that, you would still do this same code though. But there's a type guard to check if a node is a template

PR #2407, opened 2023-08-04
Fix output format of `@web/dev-server-storybook`

### LINKED ISSUE #2405: [dev-server-storybook] v1 published files are wrong format
Since v1 of web dev server storybook, it no longer works due to wrong file formats.

```
> wds --config .storybook/server.config.mjs

file:///path/to/node_modules/@web/dev-server-storybook/index.mjs:2
import cjsEntrypoint from './dist/index.js';
       ^^^^^^^^^^^^^
SyntaxError: The requested module './dist/index.js' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
```

Main MJS entrypoint points to `index.mjs`:
```
// this file is autogenerated with the generate-mjs-dts-entrypoints script
import cjsEntrypoint from './dist/index.js';

const { storybookPlugin } = cjsEntrypoint;

export { storybookPlugin };
```

This is attempting to import a default import from a CJS file, which makes no sense:
- Why would MJS import from a CJS file
- There is no default export in that file, only named

I think something might have gone wrong in your build process in the v1 publish

koddsson:
Set `moduleResolution` to `node16` so that TypeScript will output ES Modules and changed the `module` property so that it's not a moving target.

Fixes #2405

### DIFF
--- a/packages/dev-server-storybook/src/shared/mdx/transformMdxToCsf.ts
+++ b/packages/dev-server-storybook/src/shared/mdx/transformMdxToCsf.ts
@@ -1,5 +1,6 @@
 import mdx from '@mdx-js/mdx';
 import { transformAsync } from '@babel/core';
+// @ts-ignore
 import { createCompiler } from '@storybook/csf-tools/mdx';
 import { createError } from '../utils.js';
 
--- a/packages/dev-server-storybook/src/shared/mdx/transformMdxToCsf.ts
+++ b/packages/dev-server-storybook/src/shared/mdx/transformMdxToCsf.ts
@@ -1,7 +1,7 @@
 import mdx from '@mdx-js/mdx';
 import { transformAsync } from '@babel/core';
 import { createCompiler } from '@storybook/csf-tools/mdx';
-import { createError } from '../utils';
+import { createError } from '../utils.js';
 
 const compilers = [createCompiler({})];
 
--- a/.changeset/stale-cooks-cover.md
+++ b/.changeset/stale-cooks-cover.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server-storybook': patch
+---
+
+Fix output format of package.
--- a/packages/dev-server-storybook/src/build/build.ts
+++ b/packages/dev-server-storybook/src/build/build.ts
@@ -1,13 +1,13 @@
-import { readStorybookConfig } from '../shared/config/readStorybookConfig';
-import { validatePluginConfig } from '../shared/config/validatePluginConfig';
+import { readStorybookConfig } from '../shared/config/readStorybookConfig.js';
+import { validatePluginConfig } from '../shared/config/validatePluginConfig.js';
 
-import { createRollupConfig } from './rollup/createRollupConfig';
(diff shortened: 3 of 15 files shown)


packages/dev-server-storybook/src/shared/mdx/transformMdxToCsf.ts
```diff
@@ -1,7 +1,8 @@
 import mdx from '@mdx-js/mdx';
 import { transformAsync } from '@babel/core';
+// @ts-ignore
 import { createCompiler } from '@storybook/csf-tools/mdx';
```
Westbrook, line 4: I get this error:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/westbrookjohnson/Documents/repos/spectrum-web-components/node_modules/@storybook/csf-tools/mdx' imported from /Users/westbrookjohnson/Documents/repos/spectrum-web-components/node_modules/@web/dev-server-storybook/dist/shared/mdx/transformMdxToCsf.js
Did you mean to import @storybook/csf-tools/mdx.js?
    at new NodeError (node:internal/errors:387:5)
    at finalizeResolution (node:internal/modules/esm/resolve:429:11)
    at moduleResolve (node:internal/modules/esm/resolve:1006:10)
    at defaultResolve (node:internal/modules/esm/resolve:1214:11)
    at nextResolve (node:internal/modules/esm/loader:165:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:844:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:431:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)
    at link (node:internal/modules/esm/module_job:75:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```
Goes away if I add `.js` locally, but then I get a different error.

bashmish: I confirm this

@koddsson did you test this change with some storybook?

koddsson: I didn't manually test this no. I was kind of relying on the tests to catch this. I find it weird that they didn't 🤔 

I haven't had a lot of time to look into this but I am trying to fix this today. I've gotten to a point where I don't get any errors in a local project but I am just getting a empty page so I'm looking further.

bashmish: @koddsson are there integration tests that run a real storybook in the test environment or smth like that? if so, maybe good to improve that to detect such cases
but I thought we didn't have them, maybe some unit tests only or so, but I didn't look into that yet
for the Storybook v7 migration I think we need to improve a lot in this space

koddsson: I think I got this fixed now. It's at least working in my local project and the tests are all running green locally. Could you take a look @Westbrook and @bashmish ?

PR #2411, opened 2023-08-08
fix(rollup-plugin-import-meta-assets): allow simple backticks strings

jorenbroekema:
## What I did

1. allow simple backticks strings (Template Literals with 1 quasi and 0 expressions), since those are statically analyzable.

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/test/fixtures/backticks.js
+++ b/packages/rollup-plugin-import-meta-assets/test/fixtures/backticks.js
@@ -1,0 +1,7 @@
+const backticksImg = new URL(`./one.svg`, import.meta.url);
+const backticksImg2 = new URL(`./two.svg`, import.meta.url);
+
+const three = 'three';
+const backticksImg3 = new URL(`./${three}.svg`, import.meta.url);
+
+console.log(backticksImg, backticksImg2, backticksImg3);
--- a/.changeset/thirty-planes-arrive.md
+++ b/.changeset/thirty-planes-arrive.md
@@ -1,0 +1,5 @@
+---
+'@web/rollup-plugin-import-meta-assets': patch
+---
+
+Allow import-meta-assets rollup plugin to bundle meta assets that use backticks, but do not contain any dynamic expressions inside the template literal.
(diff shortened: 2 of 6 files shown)


packages/rollup-plugin-import-meta-assets/test/fixtures/backticks.js
```diff
@@ -0,0 +1,4 @@
+const backticksImg = new URL(`./one.svg`, import.meta.url);
+const backticksImg2 = new URL(`./two.svg`, import.meta.url);
```
bashmish, line 2: I'd add a negative scenario case with a template literal and check in the test that it didn't change

bashmish:
Looks great!

packages/rollup-plugin-import-meta-assets/test/fixtures/backticks.js
jorenbroekema: Hmm I see your point, currently I guess it ignores it and leaves such statements as is. I've added it to the test.

I was actually thinking of raising another PR that detects template literals with expressions in new URL() or other cases of dynamic-ness that cannot be statically analysed. In those cases I **think** the plugin should throw with a helpful error message explaining what's wrong... Why throw? Because it's a bit strange I think to just ignore it and have a path that's untransformed and most definitely pointing to a non-existent file in the final bundle, leading to a 404 error, so it's better to error earlier and make it clear what's wrong. And there's always warnOnError option in the plugin which would then cause the current behavior, where it's just ignored and untransformed (except with a nice warning). 
I imagine that's a breaking change though? Thoughts?

jorenbroekema: Should also consider btw that maybe it should not throw, because if you use it in combination with @rollup/plugin-dynamic-import-vars, such paths might actually work..

bashmish (approved):

jorenbroekema:
can we merge & release this?

bashmish:
> can we merge & release this?

sure, I thought you can do it with my approval yourself

I'm new to this org, let me know if anything else was expected from my side besides review/approval :)

PR #2412, opened 2023-08-09
fix(test-runner-chrome): fix debug mode

remcovaes:
## What I did

1. Fix debug mode by passing `headless: false`

This most likely was broken by upgrading to puppeteer 20.

My OS:
Ubuntu 22

To reproduce:
1. Run a test using the ChromeLauncher
2. Press `D` in the terminal

Actual:
No browser window opened

Expected:
Browser window to open


Westbrook (approved):
LGTM!

PR #2414, opened 2023-08-10
fix(test-runner-core): destroy consumer after usage to prevent a leak

remcovaes:
## What I did

1. Cleanup the caching code for readability
2. Destroy the consumer to prevent a leak that causes errors to not be displayed correctly

## Context

When running tests against a folder containing source maps, the source maps will point to the place where the actual error was thrown. When the source map is really big and the tests are run a lot of times (this problem persists during watch mode reruns) there is overflow happening somewhere. I have created an small reproduction repo here: https://github.com/remcovaes/source-map-bug-web-test-runner

Also, the consumer caching was broken, because it the cache was only filled after a promise was resolved, which means that multiple promises where started and all those promises had cache misses. I think removing the consumer cache will not result a performance issue as it was not working correct anyway. Also, I think destroying the consumer is more important and there is not another logical place where this can be done easily.

### DIFF
--- a/.changeset/strong-fans-invent.md
+++ b/.changeset/strong-fans-invent.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-core': patch
+---
+
+Destroy consumer when it is used and refactor the caching code for readability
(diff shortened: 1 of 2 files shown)


koddsson (approved):
Looks good to me :)

PR #2415, opened 2023-08-14
Fix dev server storybook once more

koddsson:
- I removed some `require` calls from `dev-server-storybook`.
- Updated `mdx` and changed the API calls to it.
- Made sure the entrypoint into `dev-server-storybook` works correctly.


.changeset/nice-ligers-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web/dev-server-storybook': patch
```
koddsson, line 2: While _technically_ these are breaking changes, I'm fine with this since the current major version never worked and these changes should have been included in that breaking change. This is just fixing the broken `1.0` version.

Westbrook (commented):
This looks good to me. If I have some time later today or tomorrow, I could build, copy/paste, and test locally, but it feels like we're moving in a good direction.

Some clarity on the breaking parts, and I'd be find to release and test, just the same. It's broken on `latest` right now, anyways.

.changeset/nice-ligers-look.md
koddsson: The API is the same but I'm deprecating the CommonJS built code that's used when the package is "required".

Westbrook: That is what the 1.0 intended to do, right? In which case, I’m even more down with this as a patch.

packages/dev-server-storybook/package.json
```diff
@@ -58,11 +57,10 @@
   "dependencies": {
     "@babel/core": "^7.16.0",
     "@babel/preset-env": "^7.16.4",
-    "@mdx-js/mdx": "^1.6.22",
+    "@mdx-js/mdx": "^2.3.0",
```
bashmish, line 60: mdx 2 is a breaking change that not directly related to this update
is it strictly necessary?

koddsson: Nope, it wasn't strictly necessary. Figured it would be good to fold into another breaking change.

bashmish: I think we should have this in Storybook 7 only
but for Storybook 6 users I'd stick to MDX 1
MDX 2 is pretty big in terms of breaking changes, I don't see how users of our current Storybook 6 prebuild setup will benefit from this

PR #2418, opened 2023-08-17
test on node 20

koddsson:
## What I did

1. Added node version 20 to the test matrix so that we are testing against the latest node runtime.


koddsson:
I can't find what breaking change caused this in node 🤔

PR #2422, opened 2023-08-22
fix(rollup-plugin-import-meta-assets): fix #2421

### LINKED ISSUE #2421: @web/rollup-plugin-import-meta-assets fails if encountering any reference to a directory
### Observed behavior

If any module contains a pattern that references a _directory_, such as
```javascript
new URL('./', import.meta.url)
```
then this plugin fails with the message `EISDIR: illegal operation on a directory, read`. Setting `warnOnError: true` does not help: if the offending module also contains a pattern that references a _file_ then that file will not be added to the rollup pipeline.

### Expected behavior

Patterns that reference directories should be silently ignored by default.

### Edit:

My apologies, I just noticed that -- as opposed to what I wrote above -- setting `warnOnError: true` _does_ help, and other file references are added to the rollup pipeline.
Nevertheless, IMHO if a pattern references a directory then this was probably intended and should not be considered as an error, not even as a not-fatal one.

undecaf:
Makes rollup-plugin-import-meta-assets ignore references to directories such as e.g. `new URL('./', import.meta.url)`

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
+++ b/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
@@ -100,10 +100,12 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
               );
               modifiedCode = true;
             } catch (error) {
-              if (warnOnError) {
-                this.warn(error, node.arguments[0].start);
-              } else {
-                this.error(error, node.arguments[0].start);
+              if (error.code !== 'EISDIR') {
+                if (warnOnError) {
+                  this.warn(error, node.arguments[0].start);
+                } else {
+                  this.error(error, node.arguments[0].start);
+                }
               }
             }
           }
--- a/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
+++ b/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
@@ -48,6 +48,7 @@ function isNewUrlImportMetaUrl(node) {
 /**
  * Detects assets references relative to modules using patterns such as `new URL('./path/to/asset.ext', import.meta.url)`.
  * The assets are added to the rollup pipeline, allowing them to be transformed and hash the filenames.
+ * Patterns that represent directories are skipped.
  *
  * @param {object} options
  * @param {string|string[]} [options.include] A picomatch pattern, or array of patterns, which specifies the files in the build the plugin should operate on. By default all files are targeted.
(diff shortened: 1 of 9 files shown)


packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -100,10 +100,12 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
-              if (warnOnError) {
-                this.warn(error, node.arguments[0].start);
-              } else {
-                this.error(error, node.arguments[0].start);
+              if (error.code !== 'EISDIR') {
+                if (warnOnError) {
+                  this.warn(error, node.arguments[0].start);
+                } else {
+                  this.error(error, node.arguments[0].start);
+                }
```
bashmish, line 108: I think we should instead match only expressions which contain an extension in the path by patching `isNewUrlImportMetaUrl` method, so that we don't even run into this error

also maybe good to rename `isNewUrlImportMetaUrl ` to `isNewUrlAssetUrl`

bashmish: [this comment](https://github.com/modernweb-dev/web/issues/2421#issuecomment-1690682097) address this, so I'm OK with the current check for `EISDIR`

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -100,10 +100,12 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
               );
               modifiedCode = true;
             } catch (error) {
-              if (warnOnError) {
-                this.warn(error, node.arguments[0].start);
-              } else {
-                this.error(error, node.arguments[0].start);
+              if (error.code !== 'EISDIR') {
```
bashmish, line 103: I'd add a comment that this ignores dirs and needs to be done like this to have confidence it's a dir on the file system level

bashmish (changes requested):
thanks for the fix! nice work 👍 

just a small remark to clarify this specific logic in code for future generations :)

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
bashmish: actually, you can also make a test by creating a file without an extension and making sure it's processed correctly

undecaf: Done

PR #2432, opened 2023-08-30
Publish dist lol

koddsson:
Added `dev-server-storybook` to the root `tsconfig.json`.

### DIFF
--- a/.changeset/purple-mirrors-clap.md
+++ b/.changeset/purple-mirrors-clap.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server-storybook': patch
+---
+
+Make sure to publish `@web/dev-server-storybook`
--- a/package-lock.json
+++ b/package-lock.json
@@ -27142,7 +27142,7 @@
     },
     "packages/dev-server": {
       "name": "@web/dev-server",
-      "version": "0.3.0",
+      "version": "0.3.1",
       "license": "MIT",
       "dependencies": {
         "@babel/code-frame": "^7.12.11",
@@ -27385,7 +27385,7 @@
     },
     "packages/dev-server-storybook": {
       "name": "@web/dev-server-storybook",
-      "version": "1.0.2",
+      "version": "1.0.3",
       "license": "MIT",
       "dependencies": {
         "@babel/core": "^7.16.0",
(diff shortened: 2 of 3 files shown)


bashmish (approved):

koddsson:
@bashmish:

Any idea on how I can make sure `module` and `moduleResolution` are set to `node16`? Running the steps you mentioned makes the following diff:

```diff
index d23e2f60..7430b12e 100644
--- a/packages/dev-server-storybook/tsconfig.json
+++ b/packages/dev-server-storybook/tsconfig.json
@@ -3,8 +3,7 @@
 {
   "extends": "../../tsconfig.node-base.json",
   "compilerOptions": {
-    "module": "node16",
-    "moduleResolution": "node16",
+    "module": "commonjs",
     "outDir": "./dist",
     "rootDir": "./src",
     "composite": true,
```

koddsson:
Looks like some projects like `@web/mocks` don't use the `workspace-packages` stuff so I think it's good that we don't either.

bashmish:
@koddsson oh, I see you updated the main "tsconfig.json"
this one will update on `npm run update` next time and remove your package again, maybe that's what happened recently

koddsson:
Hmm, but it needs to be in there so it's built during `npm run build` :(

Let me try to figure stuff out.

bashmish:
@koddsson min Node is now 16, right? can we just update it in all tsconfigs then? I'll make another PR

packages/dev-server-storybook/tsconfig.json
```diff
@@ -1,5 +1,3 @@
-// Don't edit this file directly. It is generated by generate-ts-configs script
-
 {
   "extends": "../../tsconfig.node-base.json",
   "compilerOptions": {
```
bashmish, line 3: the lines are gone now :(

koddsson: Yeah let me revert back to a good state.

bashmish (approved):

bashmish:
> > @koddsson min Node is now 16, right? can we just update it in all tsconfigs then? I'll make another PR
> 
> I created a PR to update to Node 18 a while back. We aren't running node 16 in CI anyways.

do IIUC that we generate Node16-compatible JS while having Node18 minimum requirement?

koddsson:
Ah I think the issue of `npm update` blowing away the reference in `tsconfig.json` is still present here though. Working on that now.

PR #2449, opened 2023-09-11
Dynamic import meta assets

Soesah:
## What I did

1. Allow dynamic vars when using `new URL` re-using code from https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars. 
2. Updated tests
3. Updated documentation

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/package.json
+++ b/packages/rollup-plugin-import-meta-assets/package.json
@@ -41,8 +41,10 @@
     "import-meta"
   ],
   "dependencies": {
+    "@rollup/plugin-dynamic-import-vars": "^2.0.5",
     "@rollup/pluginutils": "^5.0.2",
     "estree-walker": "^2.0.2",
+    "globby": "^13.2.2",
     "magic-string": "^0.30.0"
   },
   "types": "dist/index"
--- a/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
+++ b/packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
@@ -7,6 +7,10 @@ const path = require('path');
 const { createFilter } = require('@rollup/pluginutils');
 const { asyncWalk } = require('estree-walker');
 const MagicString = require('magic-string');
+const {
+  dynamicImportToGlob: dynamicURLToGlob,
+  VariableDynamicImportError: VariableURLError,
+} = require('@rollup/plugin-dynamic-import-vars');
 
 /**
  * Extract the relative path from an AST node representing this kind of expression `new URL('./path/to/asset.ext', import.meta.url)`.
@@ -24,25 +28,32 @@ function getRelativeAssetPath(node) {
  * Checks if a AST node represents this kind of expression: `new URL('./path/to/asset.ext', import.meta.url)`.
  *
  * @param {import('estree').Node} node - The AST node
- * @returns {boolean}
+ * @returns {undefined | 'static' | 'dynamic'}
  */
-function isNewUrlImportMetaUrl(node) {
-  return (
+function getImportMetaUrlType(node) {
+  const isNewURL =
     node.type === 'NewExpression' &&
     node.callee.type === 'Identifier' &&
     node.callee.name === 'URL' &&
     node.arguments.length === 2 &&
     (node.arguments[0].type === 'Literal' ||
-      // Allow template literals if it does not contain any dynamic expressions
-      (node.arguments[0].type === 'TemplateLiteral' &&
-        node.arguments[0].quasis.length === 1 &&
-        node.arguments[0].expressions.length === 0)) &&
+      // Allow template literals, reuses @rollup/plugin-dynamic-import-vars logic
+      node.arguments[0].type === 'TemplateLiteral') &&
     typeof getRelativeAssetPath(node) === 'string' &&
     node.arguments[1].type === 'MemberExpression' &&
     node.arguments[1].object.type === 'MetaProperty' &&
     node.arguments[1].property.type === 'Identifier' &&
-    node.arguments[1].property.name === 'url'
-  );
+    node.arguments[1].property.name === 'url';
+
+  if (!isNewURL) {
+    return undefined;
+  }
+
+  if (node.arguments[0].type === 'TemplateLiteral' && node.arguments[0].expressions.length > 0) {
+    return 'dynamic';
+  }
+
+  return 'static';
 }
 
 /**
@@ -62,19 +73,92 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
 
   return {
     name: 'rollup-plugin-import-meta-assets',
-
     async transform(code, id) {
       if (!filter(id)) {
         return null;
       }
 
-      const ast = this.parse(code);
-      const magicString = new MagicString(code);
+      // Part 1: resolve dynamic template literal expressions
+      const parsed = this.parse(code);
+
+      let dynamicURLIndex = -1;
+      let ms;
+
+      await asyncWalk(parsed, {
+        enter: async node => {
+          const importMetaUrlType = getImportMetaUrlType(node);
+
+          if (importMetaUrlType !== 'dynamic') {
+            return;
+          }
+          dynamicURLIndex += 1;
+
+          try {
+            // see if this is a Template Literal with expressions inside, and generate a glob expression
+            const glob = dynamicURLToGlob(node.arguments[0], code.substring(node.start, node.end));
+
+            if (!glob) {
+              // this was not a variable dynamic url
+              return;
+            }
+
+            const { globby } = await import('globby');
+            // execute the glob
+            const result = await globby(glob, { cwd: path.dirname(id) });
+            const paths = result.map(r =>
+              r.startsWith('./') || r.startsWith('../') ? r : `./${r}`,
+            );
+
+            // create magic string if it wasn't created already
+            ms = ms || new MagicString(code);
+            // unpack variable dynamic url into a function with url statements per file, rollup
+            // will turn these into chunks automatically
+            ms.prepend(
+              `function __variableDynamicURLRuntime${dynamicURLIndex}__(path) {
+  switch (path) {
+${paths.map(p => `    case '${p}': return new URL('${p}', import.meta.url);`).join('\n')}
+${`    default: return new Promise(function(resolve, reject) {
+      (typeof queueMicrotask === 'function' ? queueMicrotask : setTimeout)(
+        reject.bind(null, new Error("Unknown variable dynamic new URL statement: " + path))
+      );
+    })\n`}   }
+ }\n\n`,
+            );
+            // call the runtime function instead of doing a dynamic url, the url specifier will
+            // be evaluated at runtime and the correct url will be returned by the injected function
+            ms.overwrite(
+              node.start,
+              node.start + 7,
+              `__variableDynamicURLRuntime${dynamicURLIndex}__`,
+            );
+          } catch (error) {
+            if (error instanceof VariableURLError && warnOnError) {
+              this.warn(error);
+            } else {
+              this.error(error);
+            }
+          }
+        },
+      });
+
+      let newCode = code;
+      if (ms && dynamicURLIndex !== -1) {
+        newCode = ms.toString();
+      }
+
+      // Part 2: emit asset files
+      const ast = this.parse(newCode);
+      const magicString = new MagicString(newCode);
       let modifiedCode = false;
 
       await asyncWalk(ast, {
         enter: async node => {
-          if (isNewUrlImportMetaUrl(node)) {
+          const importMetaUrlType = getImportMetaUrlType(node);
+          if (!importMetaUrlType) {
+            return;
+          }
+
+          if (importMetaUrlType === 'static') {
             const absoluteScriptDir = path.dirname(id);
             const relativeAssetPath = getRelativeAssetPath(node);
             const absoluteAssetPath = path.resolve(absoluteScriptDir, relativeAssetPath);
@@ -116,7 +200,7 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
 
       return {
         code: magicString.toString(),
-        map: modifiedCode ? magicString.generateMap({ hires: true }) : null,
+        map: modifiedCode ? magicString.generateMap({ hires: true, includeContent: true }) : null,
       };
     },
   };
--- a/packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
+++ b/packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
@@ -1,0 +1,8 @@
+const names = ['one', 'two'];
+// value of one could also be "two" or "three", bundler does not analyze the value itself
+// Therefore, we expect both one.svg, two.svg and three.svg to be bundled, and this to turn into a switch statement
+// with 3 cases (for all 3 assets in the dynamic-assets folder)
+const dynamicImgs = names.map(n => new URL(`./dynamic-assets/${n}.svg`,import.meta.url));
+const backticksImg = new URL(`./dynamic-assets/three.svg`, import.meta.url);
+
+console.log(dynamicImgs, backticksImg);
--- a/packages/rollup-plugin-import-meta-assets/test/integration.test.js
+++ b/packages/rollup-plugin-import-meta-assets/test/integration.test.js
@@ -258,19 +258,20 @@ describe('rollup-plugin-import-meta-assets', () => {
     );
   });
 
-  it('allows backticks in path if there are no dynamic parts included', async () => {
+  it('allows backtics and dynamic vars in path', async () => {
     const config = {
-      input: { backticks: require.resolve('./fixtures/backticks.js') },
+      input: { 'dynamic-vars': require.resolve('./fixtures/dynamic-vars.js') },
       plugins: [importMetaAssets({ warnOnError: true })],
     };
 
     const bundle = await rollup.rollup(config);
     const { output } = await bundle.generate(outputConfig);
 
-    expect(output.length).to.equal(3);
-    expectChunk(output, 'snapshots/backticks.js', 'backticks.js', [
-      expectAsset(output, 'snapshots/one.svg', 'one.svg', 'assets/one-824f522a.svg'),
+    expect(output.length).to.equal(4);
+    expectChunk(output, 'snapshots/dynamic-vars.js', 'dynamic-vars.js', [
+      expectAsset(output, 'snapshots/three.svg', 'three.svg', 'assets/three-63bfb103.svg'),
       expectAsset(output, 'snapshots/two.svg', 'two.svg', 'assets/two-efaa9ab3.svg'),
+      expectAsset(output, 'snapshots/one.svg', 'one.svg', 'assets/one-824f522a.svg'),
     ]);
   });
 
--- a/packages/rollup-plugin-import-meta-assets/test/snapshots/dynamic-vars.js
+++ b/packages/rollup-plugin-import-meta-assets/test/snapshots/dynamic-vars.js
@@ -1,0 +1,21 @@
+function __variableDynamicURLRuntime0__(path) {
+  switch (path) {
+    case './dynamic-assets/one.svg': return new URL(new URL('assets/one-824f522a.svg', import.meta.url).href, import.meta.url);
+    case './dynamic-assets/three.svg': return new URL(new URL('assets/three-63bfb103.svg', import.meta.url).href, import.meta.url);
+    case './dynamic-assets/two.svg': return new URL(new URL('assets/two-efaa9ab3.svg', import.meta.url).href, import.meta.url);
+    default: return new Promise(function(resolve, reject) {
+      (typeof queueMicrotask === 'function' ? queueMicrotask : setTimeout)(
+        reject.bind(null, new Error("Unknown variable dynamic new URL statement: " + path))
+      );
+    })
+   }
+ }
+
+const names = ['one', 'two'];
+// value of one could also be "two" or "three", bundler does not analyze the value itself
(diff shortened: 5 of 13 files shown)


packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
```diff
@@ -0,0 +1,8 @@
+const one = 'one';
+// value of one could also be "two", bundler does not analyze that format
+// hence, we expect both one.svg and two.svg to be bundled, and this to turn into a switch statement
+// with 2 cases (for both assets in the dynamic-assets folder)
```
jorenbroekema, line 4:
```suggestion
// value of one could also be "two" or "three", bundler does not analyze the value itself
// Therefore, we expect both one.svg, two.svg and three.svg to be bundled, and this to turn into a switch statement
// with 3 cases (for all 3 assets in the dynamic-assets folder)
```

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @web/rollup-plugin-import-meta-assets
 
+## 2.0.3
+
+### Patch Changes
+
+- Enabled using dynamic import vars when importing assets
```
bashmish, line 7: did you add it manually?
if yes, please remove it and also the version bump in `package.json`
and instead use `npm changelog` to make a patch changelog file

packages/rollup-plugin-import-meta-assets/test/integration.test.js
```diff
@@ -258,19 +258,20 @@ describe('rollup-plugin-import-meta-assets', () => {
     );
   });
 
-  it('allows backticks in path if there are no dynamic parts included', async () => {
+  it.only('allows backtics and dynamic vars in path', async () => {
```
jorenbroekema, line 261:
```suggestion
  it('allows backtics and dynamic vars in path', async () => {
```

packages/rollup-plugin-import-meta-assets/package.json
```diff
@@ -41,8 +41,10 @@
     "import-meta"
   ],
   "dependencies": {
+    "@rollup/plugin-dynamic-import-vars": "^2.0.5",
     "@rollup/pluginutils": "^5.0.2",
     "estree-walker": "^2.0.2",
+    "fast-glob": "^3.3.1",
```
koddsson, line 47: This would be the third globbing library that we have in the monorepo. Do you mind using `globby` so that it's a bit easier to manage our dependencies? 😄

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -24,25 +29,32 @@ function getRelativeAssetPath(node) {
  * Checks if a AST node represents this kind of expression: `new URL('./path/to/asset.ext', import.meta.url)`.
  *
  * @param {import('estree').Node} node - The AST node
- * @returns {boolean}
+ * @returns {false | 'static' | 'dynamic'}
```
bashmish, line 32: `false` doesn't play nicely with the new method name and purpose, instead use `undefined` when type can't be determined

packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
```diff
@@ -0,0 +1,8 @@
+const one = 'one';
+// value of one could also be "two", bundler does not analyze that format
```
koddsson, line 2: Could we elaborate on how the variable `one` could have the value `"two"` here? It's not apparent to me at least.

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +74,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+            ms = ms || new MagicString(code);
+            // unpack variable dynamic url into a function with url statements per file, rollup
+            // will turn these into chunks automatically
+            ms.prepend(
+              `function __variableDynamicURLRuntime${dynamicURLIndex}__(path) {
+  switch (path) {
+${paths.map(p => `    case '${p}': return new URL('${p}', import.meta.url);`).join('\n')}
+${`    default: return new Promise(function(resolve, reject) {
+      (typeof queueMicrotask === 'function' ? queueMicrotask : setTimeout)(
+        reject.bind(null, new Error("Unknown variable dynamic import: " + path))
```
jorenbroekema, line 122:
```suggestion
        reject.bind(null, new Error("Unknown variable dynamic new URL statement: " + path))
```

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @web/rollup-plugin-import-meta-assets
 
+## 2.0.3
```
koddsson, line 3: Technically a minor version according to semver, right?

packages/rollup-plugin-import-meta-assets/test/integration.test.js
```diff
@@ -258,19 +258,20 @@ describe('rollup-plugin-import-meta-assets', () => {
...
     const { output } = await bundle.generate(outputConfig);
 
-    expect(output.length).to.equal(3);
-    expectChunk(output, 'snapshots/backticks.js', 'backticks.js', [
-      expectAsset(output, 'snapshots/one.svg', 'one.svg', 'assets/one-824f522a.svg'),
+    expect(output.length).to.equal(4);
+    expectChunk(output, 'snapshots/dynamic-vars.js', 'dynamic-vars.js', [
+      expectAsset(output, 'snapshots/three.svg', 'three.svg', 'assets/three-63bfb103.svg'),
       expectAsset(output, 'snapshots/two.svg', 'two.svg', 'assets/two-efaa9ab3.svg'),
+      expectAsset(output, 'snapshots/one.svg', 'one.svg', 'assets/one-824f522a.svg'),
```
bashmish, line 274: I'm curious what happen with `three.svg` given it can be both included as a static one and a dynaminc one?
won't it be bundled twice?

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -116,7 +205,7 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
 
       return {
         code: magicString.toString(),
-        map: modifiedCode ? magicString.generateMap({ hires: true }) : null,
+        map: modifiedCode ? magicString.generateMap({ hires: true, includeContents: true }) : null,
```
bashmish, line 208: what does `includeContents: true` change?

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
Soesah: Will do

packages/rollup-plugin-import-meta-assets/test/integration.test.js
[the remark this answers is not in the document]
Soesah: Will do

packages/rollup-plugin-import-meta-assets/test/integration.test.js
jorenbroekema: The asset name and source will be identical in the `this.emitFile()` call, which should mean that either it is ignored or the same file is overwritten, either way you will have 1 asset file in the final bundle, so I think it's fine

Soesah: In the output it is only there once.

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
jorenbroekema: `npx changeset` indeed 👍🏻

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
Soesah: ok

packages/rollup-plugin-import-meta-assets/test/integration.test.js
Soesah: Will change

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
Soesah: @bashmish I'm a little new to changesets and changelogs. Do you want me to do `npx changeset` and commit or `npx changelog`? And does this count as a patch or a minor?

packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
koddsson: > You could write
> 
> ```
> const urls = ['one', 'two'].map( v => new URL(`./dynamic-assets/${one}.svg`, import.meta.url));
> ```

Ahh, could we use this snippet in the test fixture to make it obvious to readers?

packages/rollup-plugin-import-meta-assets/test/integration.test.js
koddsson: We really should have a lint rule for this :)

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
Soesah: Will fix

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
bashmish: yes, and then you see how the system will bump the version in the GitHub comment in your PR
the changelog will also include the text which will end up in CHANGELOG.md of the corresponding package

image

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
bashmish: thanks!

packages/rollup-plugin-import-meta-assets/CHANGELOG.md
bashmish: actually I agree, it's kinda a new feature, so should be 2.1.0 which in this case is a `minor`

@Soesah can you please fix the changelog file accordingly? you can manually update `patch` to `minor` in it without regenerating the file

packages/rollup-plugin-import-meta-assets/test/fixtures/dynamic-vars.js
Soesah: Yes!

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +74,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+            // see if this is a Template Literal with expressions inside, and generate a glob expression
+            const glob = dynamicURLToGlob(node.arguments[0], code.substring(node.start, node.end));
+
+            if (!glob) {
+              // this was not a variable dynamic url
+              return;
+            }
+
+            // execute the glob
+            const result = fastGlob.sync(glob, { cwd: path.dirname(id) });
```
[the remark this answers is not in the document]
Soesah: done

packages/rollup-plugin-import-meta-assets/package.json
Soesah: done.

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +73,97 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+            // call the runtime function instead of doing a dynamic url, the url specifier will
+            // be evaluated at runtime and the correct url will be returned by the injected function
+            ms.overwrite(
+              node.start,
+              node.start + 7,
+              `__variableDynamicURLRuntime${dynamicURLIndex}__`,
+            );
+          } catch (error) {
+            if (error instanceof VariableURLError) {
+              // TODO: line number
```
bashmish, line 136: @Soesah is this for now or for later?

Soesah: "That ain't mine"

Soesah: We copied some code for the implementation from the dynamic import plugin. This was part of it. I'll remove the line to avoid confusion.

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +74,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+    })\n`}   }
+ }\n\n`,
+            );
+            // call the runtime function instead of doing a dynamic url, the url specifier will
+            // be evaluated at runtime and the correct url will be returned by the injected function
+            ms.overwrite(
+              node.start,
+              node.start + 7,
+              `__variableDynamicURLRuntime${dynamicURLIndex}__`,
+            );
```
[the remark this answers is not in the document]
bashmish: `import.meta.url` is still used in the output. but just not passed as a param

I'm more concerned to see that it's used twice in it and URL constructor is also called twice, maybe that should be cleaned up somehow, but I don't know how

packages/rollup-plugin-import-meta-assets/test/snapshots/dynamic-vars.js
```diff
@@ -0,0 +1,21 @@
+function __variableDynamicURLRuntime0__(path) {
+  switch (path) {
+    case './dynamic-assets/one.svg': return new URL(new URL('assets/one-824f522a.svg', import.meta.url).href, import.meta.url);
```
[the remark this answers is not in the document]
Soesah: Yes. They are needed here, but not needed in the original code, which was rewritten. Removing it from the rewritten original code is tricky because of spaces.

bashmish:
This is looking great! I have one question left.

Before merging it would be nice to rebase with latest master and regenerate lock file, as I think there are some gaps there right now. I also ran `npm i` on my machine and it does a little change even with the current state of the branch.

packages/rollup-plugin-import-meta-assets/test/snapshots/dynamic-vars.js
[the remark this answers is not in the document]
bashmish: the outside URL wrapper is not needed, is it?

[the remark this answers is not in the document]
Soesah: Ah, I see what you mean. The doubling up comes from the process being performed in two parts. Not sure if we can prevent that.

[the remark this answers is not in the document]
Soesah: We need to wrap it in `new URL` so that the second part also parses it and adds the hash.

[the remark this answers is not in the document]
bashmish: not sure if that's a good enough example, but

```
> new URL(new URL('assets/one-824f522a.svg', 'http://localhost/some-path/').href, 'http://localhost/some-path/').href
'http://localhost/some-path/assets/one-824f522a.svg'
> new URL('assets/one-824f522a.svg', 'http://localhost/some-path/').href
'http://localhost/some-path/assets/one-824f522a.svg'
```

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +73,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+              this.error(error);
+            }
+          }
+        },
+      });
+
+      let newCode = code;
+      if (ms && dynamicURLIndex !== -1) {
+        newCode = ms.toString();
+      }
```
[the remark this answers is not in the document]
Soesah: It also has some minor specific changes.

packages/rollup-plugin-import-meta-assets/test/snapshots/dynamic-vars.js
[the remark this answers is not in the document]
bashmish: I'm missing smth here
- how does the 2nd one add the hash?
- parsing is just parsing, without generating code it won't change much from the functional perspective?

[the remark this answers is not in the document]
jorenbroekema: This behavior of an extra wrapping `new URL()` is not something that we added in this PR, it happens for every snapshot in the test if you look at them. 

It's probably required for when you have this code:

```js
const searchParams = new URL('./four.svg', import.meta.url).searchParams;
```

and that becomes 

```js
const searchParams = new URL(new URL('assets/four-97848687.svg', import.meta.url).href, import.meta.url).searchParams;
```

which is not the same as

```js
const searchParams = new URL('assets/four-97848687.svg', import.meta.url).searchParams;
```
So the wrapping new URL() + href prop is needed then?

bashmish:
@Soesah can you please rebase and regenerate package-lock?

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +73,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+          } catch (error) {
+            if (error instanceof VariableURLError) {
+              if (warnOnError) {
+                this.warn(error);
+              } else {
+                this.error(error);
+              }
+            } else {
+              this.error(error);
+            }
```
koddsson, line 143: Could be simplified to this?

```suggestion
            if (error instanceof VariableURLError && warnOnError) {
                this.warn(error);
            } else {
              this.error(error);
            }
```

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +73,96 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+              // this was not a variable dynamic url
+              return;
+            }
+
+            const { globby } = await import('globby');
+            // execute the glob
+            const result = await globby(glob, { cwd: path.dirname(id) });
+            const paths = result.map(r =>
+              r.startsWith('./') || r.startsWith('../') ? r : `./${r}`,
+            );
```
koddsson, line 110: I guess we gotta double check here in the test results if the returned paths have POSIX separators on Windows. I guess this ran green in CI earlier using `fast-glob` so it should be all good.

bashmish: Joren @jorenbroekema is on Windows, I assume he tested this?

Soesah:
> @Soesah can you please rebase with latest master and regenerate package-lock?

I have.

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
Soesah: Done

bashmish:
> > @Soesah can you please rebase with latest master and regenerate package-lock?
> 
> I have.

I think you have old master in your fork

image

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -62,19 +73,91 @@ function importMetaAssets({ include, exclude, warnOnError, transform } = {}) {
...
+            ms.overwrite(
+              node.start,
+              node.start + 7,
+              `__variableDynamicURLRuntime${dynamicURLIndex}__`,
+            );
+          } catch (error) {
+            if (error instanceof VariableURLError && warnOnError) {
+              this.warn(error);
+            }
+            this.error(error);
```
koddsson, line 138: This will now call both `this.warn` and `this.error` if the condition on line 135 is true.

Soesah: You're right, I've fixed it.

Soesah:
> > > @Soesah can you please rebase with latest master and regenerate package-lock?
> > 
> > 
> > I have.
> 
> I think you have old master in your fork
> 
> image

You're right, strange though. I've rectified it.

bashmish (approved):

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
koddsson: Looks like the tests are running on Windows so it's probably fine 🤷🏻

bashmish: they do and they fail due to timeouts quite often, so it's hard to make them all run and turn green
I restarted 3-4 times yesterday, no luck, not sure why so unstable though
so I won't say it's 100% working, but in one of the runs I saw lots of tests passing

koddsson: Yup the flaky tests are in `browser-logs` I think so if all the other tests ran green then we are all good.

koddsson (approved):
Looks good to me!

PR #2452, opened 2023-09-12
Include package.json in exports

Westbrook:
## What I did

1. Add an export that yelled at me when I tried to build Storybook

### DIFF
--- a/.changeset/polite-fans-tie.md
+++ b/.changeset/polite-fans-tie.md
@@ -1,0 +1,5 @@
+---
+"@web/storybook-framework-web-components": patch
+---
+
+Include package.json in package exports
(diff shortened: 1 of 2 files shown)


Westbrook:
@bashmish were we not able to get a build test into the initial PR for this? Seems like something that would have gotten caught in that context, worth thinking about for later (sooner?). For now, this isn't something that can be managed with `patch-package`, so moving this forward will be needed to continue to proof casing I'm doing on my end.

bashmish:
> @bashmish were we not able to get a build test into the initial PR for this? Seems like something that would have gotten caught in that context, worth thinking about for later (sooner?). For now, this isn't something that can be managed with `patch-package`, so moving this forward will be needed to continue to proof casing I'm doing on my end.

I tested locally with `npm install ../path-to-folder`, and I checked that exports were used with such installation (although why wouldn't they?), not sure why I couldn't catch this bug.

We will work on a good testing setup, hopefully sooner.

Westbrook:
Local to the project the code may implicitly have access to this file, but hard to say. This release works on my end. Now I just have to figure out why web sockets are being started in one of my projects so I can manifest the "big win" from this conversion.

PR #2457, opened 2023-09-18
chore: update import meta assets docs

jorenbroekema:
## What I did

1. Fix missing word in readme
2. Copy the dynamic import section to the website docs

### DIFF
--- a/docs/docs/building/rollup-plugin-import-meta-assets.md
+++ b/docs/docs/building/rollup-plugin-import-meta-assets.md
@@ -18,6 +18,16 @@ This doesn't work in the browser without transformation. This plugin makes it po
 const myImg = new URL('./assets/my-img.png', import.meta.url);
 ```
 
+### Dynamic variables
+
+You can also use dynamic variables like so:
+
+```js
+const myImg = new URL(`./assets/${myImg}.png`, import.meta.url);
+```
+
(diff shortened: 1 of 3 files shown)


bashmish (approved):

docs/docs/building/rollup-plugin-import-meta-assets.md
```diff
@@ -18,6 +18,16 @@ This doesn't work in the browser without transformation. This plugin makes it po
...
+### Dynamic variables
+
+You can also use dynamic variables like so:
+
+```js
+const myImg = new URL(`./assets/${myImg}.png`, import.meta.url);
+```
+
+Please consult the [dynamic-import-vars plugin](https://github.com/rollup/plugins/blob/master/packages/dynamic-import-vars) documentation for options and limitations.
+
```
bashmish, line 30: is it common that we need to duplicate this?

in this specific case I feel like it definitely needs to be here in docs, but README is usually for a link to docs or for a very basic docs without going into config options, so maybe remove it from there? up to you

jorenbroekema: Yeah it's definitely a fair bit of overlap that I see, not really a consistent convention as far as I can tell where to draw the line for overlapping stuff. Since it's really short section, I think I'd like to keep it in the README as well.

PR #2458, opened 2023-09-20
fix: fix infinite loop if no package.json is present

jean-loutre:
When launching web-test-runner from global npm packages, and without a package.json in the current or any of the parent directory, it hangs on an infinite loop searching for a package.json, due to path.resolve('/', '..') returning '/'.

### DIFF
--- a/packages/config-loader/src/getPackageType.js
+++ b/packages/config-loader/src/getPackageType.js
@@ -20,7 +20,12 @@ async function getPackageType(basedir) {
         return pkgJson.type || 'commonjs';
       }
 
+      const previousPath = currentPath
       currentPath = path.resolve(currentPath, '..');
+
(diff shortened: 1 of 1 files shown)


koddsson:
Hey @jean-loutre; could you add a description to this PR so we can figure out what you are trying to achieve.

koddsson:
> When launching web-test-runner from global npm packages, and without a package.json in the current or any of the parent directory, it hangs on an infinite loop searching for a package.json, due to path.resolve('/', '..') returning '/'.

Hey, I'm having a hard time reproducing this.

```
 ~/temp/ npx @web/test-runner foo.js

Chrome: |██████████████████████████████| 1/1 test files | 0 passed, 0 failed

Finished running tests in 5.1s, all tests passed! 🎉
```

PR #2464, opened 2023-09-27
Avoid network errors when a large number of files are served with http2 ERR_HTTP2_PROTOCOL_ERROR

GabrielDelepine:
## Issue

I am experiencing network errors. It is due to the large number of files my page is loading (300 to 400 hundred).

The number of files shouldn't be a problem, especially in a project that advocate ESM, importmap, modulepreload, http2/3 multiplexing, etc...

ERR_HTTP2_PROTOCOL_ERROR

## Solution

Raise-up the maxSessionMemory of the http2 server from the default value (10) to 20

### DIFF
--- a/.changeset/smart-mice-relax.md
+++ b/.changeset/smart-mice-relax.md
@@ -1,0 +1,5 @@
+---
+'@web/dev-server-core': minor
+---
+
+Raise-up the maxSessionMemory of the http2 server to avoid network errors when a large number of files are served
--- a/packages/dev-server-core/src/server/createServer.ts
+++ b/packages/dev-server-core/src/server/createServer.ts
@@ -79,6 +79,7 @@ export function createServer(
           : path.join(dir, '..', '.self-signed-dev-server-ssl.cert'),
       ),
       allowHTTP1: true,
+      maxSessionMemory: 20,
     };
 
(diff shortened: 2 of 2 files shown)


koddsson:
Hey @GabrielDelepine; this looks ok but I'm slightly vary of the jump from the default 10 megabytes to 300 megabytes. Can you add a changeset to this as well?

koddsson:
I also want tot land #2467 before this one so we can at least verify that these changes pass the "smoke" test.

GabrielDelepine:
> Hey @GabrielDelepine; this looks ok but I'm slightly vary of the jump from the default 10 megabytes to 300 megabytes. Can you add a changeset to this as well?

Good point, the default value is `10`, `300` is too rough.
I am currently performing some tests with more reasonable values

koddsson:
> I also want to land #2467 before this one so we can at least verify that these changes pass the "smoke" test.

I've landed this so if you rebase this PR on `main` we'll have the new `http2` test for a little bit of more confidence.

GabrielDelepine:
Thanks, the branch is now rebased with `master`.

After conducting some performance tests, a value of 20 is enough to serve few hundred files per page. I updated the PR accordingly

Issue #2450, opened 2023-09-11
Only report browser logs for failed sessions


jasonmobley:
I'd like to be able to use the default reporter but have it only report the browser logs in the terminal output for failed sessions. We run a lot of test files at once, and we have test files that produce quite a few error and warning level log entries even for tests that pass. When the tests are passing fine, all that log content is just noise that I'd rather not see. It's kinda similar to how the default reporter doesn't print the name of each individual test that _passes_ ... only the failures are shown.

I could probably implement this myself via [filterBrowserLogs](https://github.com/modernweb-dev/web/blob/033b302c5f70edb1dc0b67e162c62b32c99db034/packages/test-runner-core/src/config/TestRunnerCoreConfig.ts#L49) if the result were passed into the filter function, e.g.

```
export default {
  ...
  filterBrowserLogs: ({ log: type, args: any[] }, session) => {
    return !session.passed;
  },
  ...
}
```

koddsson:
> I could probably implement this myself via [filterBrowserLogs](https://github.com/modernweb-dev/web/blob/033b302c5f70edb1dc0b67e162c62b32c99db034/packages/test-runner-core/src/config/TestRunnerCoreConfig.ts#L49) if the result were passed into the filter function, e.g.

This sounds like the bst way to solve this I think.

jasonmobley:
@koddsson see PR #2476

koddsson:
Closing this as fixed!

PR #2476, opened 2023-10-04
Add TestSession to filterBrowserLogs args to enable filtering by session result

jasonmobley:
## What I did

1. added the TestSession as an arg to filterBrowserLogs so that I can only print logs for failed sessions, see #2450 
2. updated paths in one of the test-runner-core tests to be package relative so that the test would pass for me (not sure if something about my env makes these fail, but those paths should be package-relative anyway IMO... commands run within a workspace context have CWD set to that package's root, not the multi-package root)

### DIFF
--- a/.changeset/spicy-kangaroos-search.md
+++ b/.changeset/spicy-kangaroos-search.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-core': minor
+---
+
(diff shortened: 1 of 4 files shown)


koddsson:
Hey @jasonmobley; looks like you need to run the formatter.

koddsson (approved):
Looks good to me 👍🏻

PR #2482, opened 2023-10-14
Allow web socket based refresh in middleware mode

### LINKED ISSUE #2471: WDS as Middleware with refresh on file change.
When WDS is a middleware, it will spin up an `app` but not a `server`. Without a `server` it cannot leverage web sockets in order to refresh the page when there are file changes:
```
Error: Cannot use watch mode when web sockets are disabled.
```
When leveraging `@web/storybook-builder` and `@web/storybook-framework-web-components`, WDS is started up as a middleware, which means it disallows `watch` and this update mechanism. Does that mean we need some added setting in Storybook in order to trigger refresh, or is there something else missing here?

@bashmish maybe you have thoughts?

Westbrook:
## What I did

1. Accept an external server when running in middleware mode.
2. Pass the server from Storybook Builder as an external server.
3. Use the external server, when available, to hang the web socket needed to watch changes.

Fixes #2471

### DIFF
--- a/packages/dev-server-core/src/server/DevServer.ts
+++ b/packages/dev-server-core/src/server/DevServer.ts
@@ -19,7 +19,6 @@ export class DevServer {
     public config: DevServerCoreConfig,
     public logger: Logger,
     public fileWatcher = chokidar.watch([]),
-    externalServer?: Server,
   ) {
     if (!config) throw new Error('Missing config.');
     if (!logger) throw new Error('Missing logger.');
@@ -28,7 +27,7 @@ export class DevServer {
       this.logger,
       this.config,
       this.fileWatcher,
-      config.middlewareMode,
+      !!config.middlewareMode,
     );
     this.koaApp = app;
     if (server) {
@@ -40,8 +39,11 @@ export class DevServer {
           this.connections.delete(connection);
         });
       });
-    } else if (externalServer) {
-      this.webSockets = new WebSocketsManager(externalServer);
+    } else if (
+      typeof this.config.middlewareMode === 'object' &&
+      this.config.middlewareMode.server
+    ) {
+      this.webSockets = new WebSocketsManager(this.config.middlewareMode.server);
     }
   }
 
--- a/packages/dev-server-core/src/server/DevServer.ts
+++ b/packages/dev-server-core/src/server/DevServer.ts
@@ -19,6 +19,7 @@ export class DevServer {
     public config: DevServerCoreConfig,
     public logger: Logger,
     public fileWatcher = chokidar.watch([]),
+    externalServer?: Server,
   ) {
     if (!config) throw new Error('Missing config.');
     if (!logger) throw new Error('Missing logger.');
@@ -39,6 +40,8 @@ export class DevServer {
           this.connections.delete(connection);
         });
       });
+    } else if (externalServer) {
+      this.webSockets = new WebSocketsManager(externalServer);
     }
   }
 
--- a/.changeset/green-arachnid-crossing.md
+++ b/.changeset/green-arachnid-crossing.md
@@ -1,0 +1,5 @@
+---
+"@web/storybook-builder": patch
+---
+
+Pass Storybook server into middleware mode to enable live refresh
--- a/.changeset/orange-octopus-murmurs.md
+++ b/.changeset/orange-octopus-murmurs.md
@@ -1,0 +1,5 @@
+---
+"@web/dev-server-core": patch
+---
+
+Allow web socket based refresh in middleware mode, by accepting the source `server` as an option for the `middlewareMode` config property
--- a/packages/dev-server-core/src/server/DevServerCoreConfig.ts
+++ b/packages/dev-server-core/src/server/DevServerCoreConfig.ts
@@ -1,5 +1,6 @@
 import { Middleware } from 'koa';
 import { Plugin } from '../plugins/Plugin';
+import { Server } from 'net';
 
 export type MimeTypeMappings = Record<string, string>;
 
@@ -21,7 +22,7 @@ export interface DevServerCoreConfig {
   /**
    * Whether to run server or not and allow to use as a middleware connected to another server.
    */
-  middlewareMode?: boolean;
+  middlewareMode?: boolean | { server: Server };
   basePath?: string;
   /**
    * The app's index.html file. When set, serves the index.html for non-file requests. Use this to enable SPA routing
--- a/packages/dev-server/src/startDevServer.ts
+++ b/packages/dev-server/src/startDevServer.ts
@@ -7,7 +7,6 @@ import { readFileConfig } from './config/readFileConfig';
 import { DevServerStartError } from './DevServerStartError';
 import { createLogger } from './logger/createLogger';
 import { openBrowser } from './openBrowser';
-import { Server } from 'net';
 
 export interface StartDevServerParams {
   /**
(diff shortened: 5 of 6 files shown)


packages/dev-server-core/src/server/DevServer.ts
```diff
@@ -39,6 +40,8 @@ export class DevServer {
           this.connections.delete(connection);
         });
       });
+    } else if (externalServer) {
+      this.webSockets = new WebSocketsManager(externalServer);
```
bashmish, line 44: is `WebSocketsManager` compatible with the Storybook manager server? with other servers?
I'm not aware of how WebSockets layer works on Storybook or WDS side

packages/dev-server-core/src/server/DevServer.ts
```diff
@@ -19,6 +19,7 @@ export class DevServer {
     public config: DevServerCoreConfig,
     public logger: Logger,
     public fileWatcher = chokidar.watch([]),
+    externalServer?: Server,
```
bashmish, line 22: I think we have 2 optional params here now: `fileWatcher` and `externalServer`
so I'd make a new "options" argument, and try to support passing `fileWatcher` in an old manner for backwards compatibility and make `options.fileWatcher` and `options.externalServer` as new ones

bashmish, line 22: another question on the same line: is that public/private? looks like it should always explicit in this file

bashmish:
Storybook has this code in the vite-builder:

https://github.com/storybookjs/storybook/blob/86ca49742641b6cc36a300ddff768da618ccb67c/code/builders/builder-vite/src/codegen-modern-iframe-script.ts#L32

```js
  const generateHMRHandler = (frameworkName: string): string => {
    // Web components are not compatible with HMR, so disable HMR, reload page instead.
    if (frameworkName === '@storybook/web-components-vite') {
      return `
      if (import.meta.hot) {
        import.meta.hot.decline();
      }`.trim();
    }

```

Do you think we need it too? Or is your solution not using `import.meta.hot`?

packages/dev-server-core/src/server/DevServer.ts
Westbrook: Yeah, I wondered if it made sense to have `externalServer` be part of `config`, being it's paired a bit with `middlewareMode`.

In that way, we could even make `externalServer` the truthy value of `middlewareMode`?

bashmish:
> Storybook has this code in the vite-builder:
> 
> https://github.com/storybookjs/storybook/blob/86ca49742641b6cc36a300ddff768da618ccb67c/code/builders/builder-vite/src/codegen-modern-iframe-script.ts#L32
> 
> ```js
>   const generateHMRHandler = (frameworkName: string): string => {
>     // Web components are not compatible with HMR, so disable HMR, reload page instead.
>     if (frameworkName === '@storybook/web-components-vite') {
>       return `
>       if (import.meta.hot) {
>         import.meta.hot.decline();
>       }`.trim();
>     }
> ```
> 
> Do you think we need it too? Or is your solution not using `import.meta.hot`?

I think I figured myself. We don't have `import.meta.hot` in the preview iframe unless we configure `dev-server-hmr` plugin, which we don't do right now. I know we don't need real HMR, but maybe we could reuse it's API to reload page instead? Curious if you considered this idea? In any case, I like your solution too.

packages/dev-server-core/src/server/DevServer.ts
Westbrook: What should be explicit? My understanding is the use of `public/private` in the constructor triggers some syntax sugar that automatically sets these values to `this.config = config`, etc. However, `externalServer` isn't needed beyond the call to `new WebSocketsManager(externalServer)`, as far as I can understand.

packages/dev-server-core/src/server/DevServer.ts
bashmish: Storybook is using "connect" / "express" API
WDS used koa API
they do have compatibility issues

for middleware I was able to find a way to make it work without any special compatibility layer
maybe it's simple here too

packages/dev-server-core/src/server/DevServer.ts
bashmish: my bad, I thought it was used outside of the constructor, but it's not
so code is good

Westbrook:
> I think I figured myself. We don't have `import.meta.hot` in the preview iframe unless we configure `dev-server-hmr` plugin, which we don't do right now. I know we don't need real HMR, but maybe we could reuse it's API to reload page instead? Curious if you considered this idea?

I had not run into this as a path. It _seems_ like HMR is mainly handled in Vite as opposed to Storybook, which might mean we'd need expanded support for it in WDS to take this path? I could spend some more time investigating this path if it felt productive, might take a little time for me to get to the bottom of this side.

packages/dev-server-core/src/server/DevServer.ts
Westbrook: The only responsibility of the `server` in `WebSocketsManager` appears to be to have an `on()` method that can listen for an `upgrade` event. This is on the shared type, so, hopefully, any compatibility issues would have to be quite serious to effect this cross-over?

packages/dev-server-core/src/server/DevServer.ts
Westbrook: Took: 
```
{
  middlewareMode: { server }
}
```

bashmish: more flexible indeed and still easy to use, great!

bashmish:
> > I think I figured myself. We don't have `import.meta.hot` in the preview iframe unless we configure `dev-server-hmr` plugin, which we don't do right now. I know we don't need real HMR, but maybe we could reuse it's API to reload page instead? Curious if you considered this idea?
> 
> I had not run into this as a path. It _seems_ like HMR is mainly handled in Vite as opposed to Storybook, which might mean we'd need expanded support for it in WDS to take this path? I could spend some more time investigating this path if it felt productive, might take a little time for me to get to the bottom of this side.

If we start supporting a wider range of frameworks in the future, e.g. Angular as it seems to be recommending WTR right, will we be able to add HMR without breaking this setup?

packages/dev-server-core/src/server/DevServer.ts
bashmish: I don't get though how this reloads the page?

Westbrook:
There are some docs about the larger HMR plugin ecosystem: https://modern-web.dev/docs/dev-server/plugins/hmr/

packages/dev-server-core/src/server/DevServer.ts
Westbrook: This watches any files that the server has served: https://github.com/modernweb-dev/web/blob/master/packages/dev-server-core/src/middleware/watchServedFilesMiddleware.ts

bashmish:
This PR seems ready to me :)
Can I assume you were testing it in your storybook when developing?

Westbrook:
I need to update my local patches for the changes from our conversation, but yes. I should be able to test that to be sure today and then I’ll report back!

Westbrook:
This is now working for me locally in a test project and in SWC. If you're interested in testing it out, https://github.com/westbrook/wds-storybook-test and https://github.com/adobe/spectrum-web-components/tree/storybook both have equivalent patches to this change.

Westbrook:
Storybook released https://www.npmjs.com/package/@storybook/addon-a11y/v/7.5.0 yesterday, which was also blocking my adoption, so it's all getting quite exciting now!! How are we looking on this PR?

bashmish (approved):

bashmish:
Awesome news! Thanks for your work @Westbrook !

PR #2483, opened 2023-10-16
Increase timeout for serialise/deserialise tests

koddsson:
These keep flaking on CI on Windows which makes them kind of useless. A quick fix for us is to just increase the timeout.


koddsson:
I have no idea why this would start failing. I'll try to just increase the timeout instead.

PR #2486, opened 2023-10-17
Support mocks in Storybook 7

bashmish:
I did:
- added new entry points for the mocks addon and decorator for Storybook 7
  - `addons: ['@web/mocks/storybook/addon/manager.js'],`
  - `import { withMocks } from '@web/mocks/storybook/addon/decorator.js';` 
  - the old "addon.js" path was not valid from the Storybook 7 perspective anyway, so I had to come up with a new one
  - we can keep paths different for Storybook 6 and 7 which allows supporting both in the same package with the same version
- restructured the code to reuse same code to create a decorator and register
- made `@web/storybook-utils` with `createAddon` copy from `@web/storybook-prebuilt`
  - puts docs in README and updated them to make it a nice public API

This is a non-breaking change and allows to support both Storybook 6 & 7 in the same package without managing multiple versions of `@web/mocks` or creating extra packages.


packages/storybook-utils/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "url": "https://github.com/modernweb-dev/web.git",
+    "directory": "packages/storybook-utils"
+  },
+  "author": "modern-web",
+  "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/storybook-utils",
+  "type": "module",
+  "exports": {
+    ".": {
+      "default": "./index.js",
+      "types": "./index.d.ts"
```
thepassle, line 20: types should be the first key, before "default"

https://publint.dev/rules#exports_types_should_be_first

bashmish: thanks!

packages/storybook-utils/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@web/storybook-utils",
+  "version": "0.0.0",
```
thepassle, line 3: should we just release this as v1? we have a bunch of packages to indefinitely tend to stay on 0.0.0 😅

bashmish: I thought it was "the way" :D hahaha
but no problem, will make it 1.0, I don't have a problem going to 2.0 if needed later

bashmish: I changed the changeset to "major"

thepassle: > I thought it was "the way" :D hahaha

yeah historically we did tend to do this, but I think for a utils package like this its fine to just go 1.0.0

thepassle (approved):

PR #2496, opened 2023-10-19
fix: lock dep

thepassle:
## What I did

1.

### DIFF
--- a/packages/mocks/package.json
+++ b/packages/mocks/package.json
@@ -65,7 +65,7 @@
     "@web/storybook-prebuilt": "^0.1.37",
     "@web/storybook-utils": "^1.0.0",
     "lit": "^2.7.5",
-    "msw": "^0.0.0-fetch.rc-23"
+    "msw": "0.0.0-fetch.rc-23"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


packages/mocks/package.json
```diff
@@ -65,7 +65,7 @@
     "@web/storybook-prebuilt": "^0.1.37",
     "@web/storybook-utils": "^1.0.0",
     "lit": "^2.7.5",
-    "msw": "^0.0.0-fetch.rc-23"
```
bashmish, line 68: I was always thinking that in this case it will be practically pinned even with the caret
thinking so based on recent experience testing storybook 7 prereleases where new rc's were never automatically picked up

bashmish: from a chat with Pascal

> somehow installing msw@^0.0.0-fetch.rc-23 results in 0.0.0-fetch.rc-9 to be installed instead o.O

apparently the suffix is interpreted as a string

thepassle: yeah I thought the same thing. We did have the version locked on the previous rc version, we used to have: `"0.0.0-fetch.rc-15"` for a long time. Silly mistake to have the `^` in there

PR #2532, opened 2023-11-01
Remove node-fetch since fetch is already implemented in node

koddsson:
We don't need `node-fetch` any more as the platform (nodejs) has caught up with standards and has `fetch` as a built-in global variable since node v18.


koddsson:
Since this change is (essentially) just a change in our tests I feel pretty confident in merging this without a changelog or version bump.

PR #2541, opened 2023-11-07
convert all projects to esm

koddsson:
Here's a first stab at converting these projects to ESM. I'm probably gonna keep this PR around and ship some of the less consequential changes in smaller PRs to shrink this PR down to something that's managable.

PR #2544, opened 2023-11-09
Include file extensions in relative imports.

koddsson:
I wanna slice off bits of the huge PR in #2541. This seems like a big piece that we can ship safely and therefore shrink the large PR.

While I _think_ we could configure the projects to have `"moduleResolution": "bundler"` I feel like this change also moves us closer to the platform and will make further platform or tool changes easier so I think it's worth doing.


packages/test-runner/src/reporter/reportTestFileResults.ts
```diff
@@ -1,12 +1,11 @@
-import { TestSession } from '@web/test-runner-core';
+import { TestSession, BufferedLogger } from '@web/test-runner-core';
```
koddsson, line 1: Fixed this import while I was here.

tsconfig.browser-base.json
```diff
@@ -1,4 +1,7 @@
 {
+  "ts-node": {
+    "experimentalResolver": true
```
koddsson, line 3: Here's the change I had to make in all the tsconfig files to get this to work.

tsconfig.node-base.json
```diff
@@ -1,4 +1,7 @@
 {
+  "ts-node": {
+    "experimentalResolver": true
```
koddsson, line 3: Here's the change I had to make in all the tsconfig files to get this to work.

tsconfig.json
```diff
@@ -37,6 +37,9 @@
     {
       "path": "./packages/dev-server-rollup/tsconfig.json"
     },
+    {
+      "path": "./packages/dev-server/tsconfig.json"
```
koddsson, line 41: These are shuffling around after I ran the generating tsconfig script. Not really a change.

koddsson:
The windows tests keep timing out. I don't think it's because of these changes but maybe #2542.

bashmish:
Why are you adding JS extension in TS files if the sources you are importing are TS?
TS has this ambiguity that files can be both TS and JS depending if it's source or dist, so no extension was also fine for TS I think, but if I were to add an extension, adding JS seems weird to me.

koddsson:
> Why are you adding JS extension in TS files if the sources you are importing are TS?

It is weird but perfectly legal TypeScript and in certain modes required. With `.js` extensions TypeScript will still try to look up the `.ts` file first before falling back to `.js` (and other extensions).

https://www.typescriptlang.org/docs/handbook/modules/reference.html#common-features-and-processes

This change will make it easier to move to ES Modules in the future.

PR #2545, opened 2023-11-09
fix(test-runner-commands): support empty string snapshots

jorenbroekema:
## What I did

1. Rather than a truthy check, check if the content property is of type string and thereby allow empty string snapshots.


packages/test-runner-commands/src/snapshotPlugin.ts
```diff
@@ -91,9 +91,10 @@ class SnapshotStore {
     const nameStr = JSON.stringify(name);
     const startMarker = `snapshots[${nameStr}]`;
     const endMarker = `/* end snapshot ${name} */\n\n`;
-    const replacement = updatedSnapshot
-      ? `${startMarker} = \n\`${updatedSnapshot}\`;\n${endMarker}`
-      : '';
+    const replacement =
+      typeof updatedSnapshot === 'string'
```
koddsson, line 95: This is interesting! The TypeScript type seems to indicate that this can only ever be a string? Seems like a issue with our types that allow this to happen. Maybe we can hunt that that error down and fix it. Doesn't have to be in this PR.

koddsson: Now that I've read the test I'm even more confused 😄 Does this need to be a `typeof` check? I realise you're doing this fix because an empty string is falsey in JavaScript but could we do a more explicit check?

Something like:

```suggestion
     updatedSnapshot?.length > 0
```

Also does this what you want it to do? `typeof ''` is still a `string`.

```
> typeof ''
'string'
```

jorenbroekema: It was doing a truthy/falsy check, and empty string means it coerces to false.
If you're correct and it's always a string, then we can delete the ternary altogether and just have it like this:
```js
const replacement = `${startMarker} = \n\`${updatedSnapshot}\`;\n${endMarker}`;
```

koddsson: > I was just under the assumption that it can happen where the updatedSnapshot is undefined, null, false, whatever and then I guess the ternary statement makes sense, but if this is impossible then I'm cool with changing it.

Ah I understand. Maybe I'm giving TypeScript too much credit in it's ability to make sure that this will never be `undefined`, `null` or `false`.

If we're guarding against those values I think this code could do with a comment explaining that is what we're doing as it stands in contrast with the type signature.

Tests that actually pass in `undefined`, `null` and `false` would also be good then.

jorenbroekema: Sorry this took a while, but I double checked and we can trust TypeScript in this case, also because the updatedSnapshot is first ran through `isSaveSnapshotPayload` function which checks for string type and a few other things. So this means the ternary checking for falsy is useless and can be deleted :). I adjusted the PR accordingly.

PR #2546, opened 2023-11-09
Add @web/test-runner-intercept

tmsns:
## What I did

This PR contains the initial commit for the new @web/test-runner-intercept project. For more info on this initiative, checkout out the discussion #2503.

- [ ] initial code
- [ ] transform to ts
- [ ] working project structure
- [ ] comments
- [ ] tests
- [ ] readme


packages/test-runner-intercept/README.md
```diff
@@ -0,0 +1,78 @@
+## Usage
+
+Setup
+
+```js
+// web-test-runner.config.mjs
+import { interceptModulePlugin } from '@web/test-runner-intercept/plugin';
```
bashmish, line 7: I think we don't have a convention to export plugins for `/plugin`. I'd stick to just the module name

but I also have a question about the module name: the plugin is focused on module interception, so cutting the name to `@web/test-runner-intercept` seems like a bad choice since the term `interceptor` can mean anything and often used in the context of AJAX, so I'd rather give the module the full name `@web/test-runner-intercept-module` or smth like that

packages/test-runner-intercept/browser/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Intercept a module in order to change its implementation.
+ * @param {string} moduleName The name of the module to intercept. When intercepting the import of a module, the name should exactly match the name of the import.
+ * @returns {Promise<Record<string, func> | void >} Writable object with every named export of the intercepted module as a property.
+ * If the module to be intercepted contains a default export, the export will be available in the `default` property. Only function expressions are supported when intercepting default exports.
+ */
+export async function interceptModule(moduleName) {
+  let module;
+  try {
+    module = await import(`/__intercept__/${moduleName}`);
```
bashmish, line 10:
```suggestion
    module = await import(`/__intercept-module__/${moduleName}`);
```

packages/test-runner-intercept/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@web/test-runner-intercept",
+  "version": "0.0.1",
```
bashmish, line 3: I'd start with at least 0.1.0

packages/test-runner-intercept/src/interceptModulePlugin.ts
```diff
@@ -0,0 +1,84 @@
...
+              return Reflect.set(obj, prop, value);
+            },
+            defineProperty(target, key, descriptor) {
+              ${namedExports.map(x => `if (key === '${x}') { ${x} = descriptor.value;}`).join('\n')}
+              return Reflect.defineProperty(target, key, descriptor);
+            },
+          });
+        `;
+        } catch (error) {
+          body = `export const __wtr_error__ = '${stripColor((error as Error).message)}';`;
```
bashmish, line 70: what makes you need stripColor?

packages/test-runner-intercept/src/interceptModulePlugin.ts
```diff
@@ -0,0 +1,84 @@
...
+    name: 'intercept-module',
+
+    serverStart(params) {
+      resolveImport = createResolveImport(params, this);
+    },
+    async serve(context) {
+      if (context.path.startsWith('/__intercept__/')) {
+        let body;
+        try {
+          const source = context.path.replace('/__intercept__/', '');
```
bashmish, line 24: how is this impacting tests running in separate tabs / concurrent?

PS: despite some issues with latest Chrome, we might still need to support that

packages/test-runner-intercept/browser/index.js
```diff
@@ -0,0 +1,20 @@
...
+  } catch (e) {
+    throw new Error(
+      `Module interception is not active. Make sure the \`interceptModulePlugin\` of \`@web/test-runner-intercept\` is added to the Test Runner config.`,
+    );
+  }
+  if (module.__wtr_error__) {
+    throw new Error(module.__wtr_error__);
+  }
+  return module.__wtr_intercepted_module__;
+}
```
bashmish, line 20: if this is a TS package, I think it makes sense to use TS here too as we do in other similar cases?

packages/test-runner-intercept/browser/index.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Intercept a module in order to change its implementation.
+ * @param moduleName The name of the module to intercept. When intercepting the import of a module, the name should exactly match the name of the import.
+ * @returns Writable object with every named export of the intercepted module as a property. 
+ * If the module to be intercepted contains a default export, the export will be available in the `default` property. Only function expressions are supported when intercepting default exports.
+ */
+export function interceptModule(moduleName: string): Promise<Record<string, func> | void>;
```
bashmish, line 7: is this auto-generated?

packages/test-runner-intercept/package.json
```diff
@@ -0,0 +1,63 @@
...
+      "import": "./plugin.mjs",
+      "require": "./dist/interceptModulePlugin.js"
+    }
+  },
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "scripts": {
+    "build": "tsc",
+    "test": "mocha \"test/**/*.test.ts\" --require ts-node/register && npm run test:ci",
```
bashmish, line 33: are you planning to add tests for this?

packages/test-runner-intercept/README.md
```diff
@@ -0,0 +1,78 @@
...
+    const stubGetCurrentHour = stub(timeLibrary, 'getCurrentHour').returns(14);
+    try {
+      const result = getTimeOfDay();
+      expect(result).to.equal('day');
+    } finally {
+      stubGetCurrentHour.restore();
+    }
+  });
+});
+```
```
bashmish, line 78: I see in some of last PRs that docs are not added to the website anymore https://modern-web.dev/

cc @thepassle @koddsson 

maybe good to add some rules about that like if we should always do it and be strict with that
and if there are exceptions, good to document where not to do that?

packages/test-runner-intercept/src/createResolveImport.ts
```diff
@@ -0,0 +1,41 @@
...
+  column?: number;
+  line?: number;
+  resolveOptions?: ResolveOptions;
+}
+export type ResolveImport = (
+  args: ResolveImportArguments,
+) => ResolveResult | Promise<ResolveResult>;
+
+/**
+ * TODO: check if `resolveImport()` can be provied by `@web/dev-server-core`'s API
```
bashmish, line 18: just to add options to the table: if it should stay private, we can also make a private shared library in the monorepo that can be used by 2 packages without releasing/exporting to public
it's not the first time I personally run into this, so was curious what others think about that

packages/test-runner-intercept/README.md
tmsns: Actually, we do. In `@web/test-runner-commands` we have this:
image
As this project only has 1 plugin, I used the singular form. The default entry point is used for the browser.

For the name, I tried to make it sweet and short, like other names in the repo: `mocks`, `test-runner-commands`. Both of them also don't mention more context. (ie. are mocks the mocks for wtr or something else; how does commands relate to wtr)

But as such, if more people think it should change, it definitely can, maybe `@web/test-runner-module-interceptor` to stay in line with other plugin names in the repo? `...-loader`, `...-reporter`.

packages/test-runner-intercept/browser/index.js
tmsns: Yup, that's good. I'll push a change that modifies all three instances.

tmsns: done! :-)

packages/test-runner-intercept/package.json
tmsns: Sure :-)

tmsns: done!

packages/test-runner-intercept/README.md
bashmish: I'd say a convention is smth used across all packages, which I checked and found it only in the example you posted, so it seems more of a choice and majority chose to have a single entry point, except where a separate one is required by other tools (e.g. `preset.js` is needed by Storybook or we have some browser-specific entry points)

in your case it's unclear to me how it benefits users?

take for example this exception:

```js
  } catch (e) {
    throw new Error(
      `Module interception is not active. Make sure the \`interceptModulePlugin\` of \`@web/test-runner-intercept\` is added to the Test Runner config.`,
    );
  }
```

I'll need to read docs to understand where to import it from

so for simplicity I'd not introduce extra entry points unless really necessary

bashmish: naming wise, I think we the following package convention

```js
@web/%tool-name%-%plugin-name%
```

so both `@web/test-runner-intercept` and `@web/test-runner-intercept-modules` or even `@web/test-runner-module-interceptor` fit it, jus the question is what describes better the plugin name part

packages/test-runner-intercept/browser/index.js
bashmish: thanks!

packages/test-runner-intercept/src/interceptModulePlugin.ts
tmsns: Resolve plugins can use ANSI colors in their error messages (eg. https://github.com/modernweb-dev/web/blob/3a9d0073ac39dec75b01a3a422584706dd064d9e/packages/dev-server-rollup/src/rollupAdapter.ts#L247C90-L247C90). Unfortunately these color codes are only working in terminal output, and not in browser's console.

Maybe I should add a comment to detail this here?

bashmish: > Unfortunately these color codes are only working in terminal output, and not in browser's console.

don't they? I think DevTools console supports colors too
but thanks for explaining, not sure it's good idea to copy color even if it does, up to you

the only thing I'm concerned is that you need to implement this utility yourself, it doesn't seem like the most straightforward thing to me to maintain, but added a 3rd party dep to remove it is also not good, so you can keep things as is if you decided it's needed

bashmish: a comment might be good in this case indeed

packages/test-runner-intercept/src/interceptModulePlugin.ts
tmsns: This path can be retrieved by multiple tabs. The generated code is concurrency proof. The result of the generation will only base itself on the original code of the module.
Any tab specific / test specific change is done in the browser context of a single tab. The changed code for an intercepted  module is in memory.

packages/test-runner-intercept/browser/index.js
tmsns: This code is browser code. I took over the convention of `@web/test-runner-commands` to only have the server-side code in TS. The script of @web as well don't take into account a mix of browser and server-sider code in TS.

packages/test-runner-intercept/src/interceptModulePlugin.ts
bashmish: let me clarify how it works and correct me where I'm wrong

- your test in Tab 1 requests intercepted module which makes a request to backend (WDS) which caches intercepted module in memor
- your test in Tab 2 fetches same module, although didn't request the intercepted one, but gets intercepted, because WDS had this cached in memory

isn't that an issue?

packages/test-runner-intercept/src/createResolveImport.ts
tmsns: What do you mean with "it should stay private"? Do you mean the comment?

bashmish: I meant `export resolveImport` might not be available via any public export, yet can be reused internally by multiple packages

this is needed sometimes

packages/test-runner-intercept/README.md
tmsns: > I'd say a convention is smth used across all packages, which I checked and found it only in the example you posted, so it seems more of a choice and majority chose to have a single entry point, except where a separate one is required by other tools (e.g. preset.js is needed by Storybook or we have some browser-specific entry points)
I'm not sure the reason of two entry points are clear. So, I'll explain why it's split. One entry point is for the browser (the default entry point). In this entry points I cannot add the plugin in there, as it would import modules with server side code in the browser. The second one, obviously now, is for the server side code only.
It's not the only package doing this. `@web/test-runner-core` is also splitting their exports in two (but actually 3) entries. And the browser split is also the reason of doing it here.

packages/test-runner-intercept/src/interceptModulePlugin.ts
tmsns: > don't they? I think DevTools console supports colors too

It does, but not using the terminal code convention.

>  not sure it's good idea to copy color even if it does, up to you

The idea is not to copy the color. :-) If I don't strip the codes, the error looks broken in the browser.
image

> a comment might be good in this case indeed

Will add one.

tmsns: Done!

packages/test-runner-intercept/src/interceptModulePlugin.ts
tmsns: Ah yes, in that sense.

Indeed, tab 2 will get the intercepted module. It should not pose a problem though. Tab 2 receives the interception, but the generated code of the intercepted module doesn't alter anything by default. The rewiring is done at runtime within the browser

packages/test-runner-intercept/browser/index.js
tmsns: Yup, indeed. :-)

But not sure "want" is the correct term here. ;-) I just reused the same pattern to not get into trouble.

packages/test-runner-intercept/src/createResolveImport.ts
tmsns: We can avoid this by removing this export altogether. Currently, this export (`resolveImport`) is created to workaround a limitation in the dev server core API. Currently there is no way to continue a resolution in an import. This is something which is widely used in Rollup plugins.
A proposal from Lars was to expose a new API object in the start hook with methods like `resolveImport`. I didn't implement this change (in dev server core) in this PR as it changes some things deep in dev server core as well. This could be added in a next PR though.

packages/test-runner-intercept/src/interceptModulePlugin.ts
bashmish: > The rewiring is done at runtime within the browser

right, so Test 2 should not notice any change in the module
lgmt!

packages/test-runner-intercept/src/createResolveImport.ts
bashmish: ok, looks like you are more up-to-date on this than me, feel free to decide here :)

packages/test-runner-intercept/package.json
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "author": "modern-web",
+  "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/test-runner-intercept",
+  "main": "browser/index.js",
+  "exports": {
+    ".": {
+      "types": "./browser/index.d.ts",
+      "default": "./browser/index.mjs"
+    },
+    "./plugin": {
```
thepassle, line 22: Package exports should use file extensions

packages/test-runner-intercept/package.json
```diff
@@ -0,0 +1,64 @@
...
+      "types": "./plugin.d.ts",
+      "import": "./plugin.mjs",
+      "require": "./dist/interceptModulePlugin.js"
```
thepassle, line 25: We should just go ESM-only, we're already looking into moving everything to ESM anyway

packages/test-runner-intercept/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "module",
+    "interception",
+    "intercept"
+  ],
+  "dependencies": {
+    "@web/dev-server-core": "^0.7.0",
+    "es-module-lexer": "^1.3.1"
+  },
+  "devDependencies": {
+    "@web/test-runner": "^0.18.0",
```
thepassle, line 59: you shouldnt have to add these dev deps here if they're already part of the monorepo package.json right?

packages/test-runner-intercept/tsconfig.json
```diff
@@ -0,0 +1,77 @@
...
+      "path": "../dev-server-legacy/tsconfig.json"
+    },
+    {
+      "path": "../dev-server/tsconfig.json"
+    },
+    {
+      "path": "../test-runner-commands/tsconfig.json"
+    },
+    {
+      "path": "../test-runner-saucelabs/tsconfig.json"
```
thepassle, line 62: Do we need all these references in this tsconfig? I dont think thats needed

packages/test-runner-intercept/package.json
tmsns: Can you mark this as resolved? :)

packages/test-runner-intercept/browser/index.d.ts
tmsns: No, this is a browser file, so there is no src/dist for it. Also, there are no scripts for browser folders.
This is the same convention as used in `test-runner-commands`, `test-runner-core` and `test-runner-visual-regression`.

packages/test-runner-intercept/README.md
tmsns: Although I agree it should be in the website docs, I would do this once the package has been a bit more tested. And maybe even with a 1.0.0 release.

packages/test-runner-intercept/package.json
tmsns: I tend to agree, but I reused the same convention used in `test-runner-commands`, `test-runner-core` and `test-runner-visual-regression`. Maybe this should be done as a general overhaul (eg. the PR to move to es modules?)

tmsns: Some scripts right now, are based on these types of conventions within the project.

packages/test-runner-intercept/package.json
tmsns: I think doing this before the PR to ESM might be more tricky. Because there are scripts that need to be adapted for it.

packages/test-runner-intercept/package.json
tmsns: Do you want me to use relative dependencies instead? I see both ways throughout the repo.

packages/test-runner-intercept/tsconfig.json
tmsns: I think you're right. I'll see if I can clean them. :-)

packages/test-runner-intercept/package.json
tmsns: I have updated the devDependencies to use the one that was used (`@web/test-runner-core` instead of `@web/test-runner`). This is similar to the deps of eg. `@web/dev-server-rollup`).

packages/test-runner-intercept/tsconfig.json
tmsns: This file is set by the `generate-ts-configs.mjs` script, but it bases itself on the deps and devDeps of the package. As I switched to `@web/test-runner-core` the list is now a lot smaller.
(fyi it also takes in transitive dependencies)

packages/test-runner-intercept/package.json
bashmish: if you are referring to `storybook-framework-web-components`, specifically `/preset` path, then IIRC it's a convention of the storybook
so if not required by other tools, I'd say you should add an extension

scripts indeed will mess it up a little bit, because they will ad a double extension to the generated files, but I don't think that's an issue, if it is, it shouldn't be hard to fix

packages/test-runner-intercept/README.md
bashmish: not sure where we are on this right now
IIRC, there were suggestion to use `mock` in some way in all names including package name
I personally think the calling this mock is not correct since this plugin doesn't create mocks, it enables mocking
so you can play with names like `enable-module-mocks` or smth like that, but I don't want to force you too :)
what I do want to improve is the genericness of the name, the `-intercept` is not sufficient, because it's not just a generic intercept and as such can be really confused with AJAX

packages/test-runner-intercept/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+const { externalFunction } = await import('external-library');
+externalFunction; // () => console.log('hello world')
+```
+
+### Caveats
+
+#### Order of imports
+
+In order to intercept a module, the module should not be referenced yet. (Once a module is loaded, the module loader also loads all its dependent modules) This means the module containing the system under test should be loaded _after_ the module interception. As interception is achieved using a function, this also means the system under test should _always_ be loaded using a dynamic import in order to be done after the interception.
```
thepassle, line 28: It's not entirely clear to me what is referenced here as being "the system under test", can you clarify this? Or rather, what does this mean for when you're testing a library that you want to mock in a component?

Say you have a component, that in the connectedCallback calls some library that you want to "intercept". Does that mean that in your test file, you have to lazy load your component file?

packages/test-runner-intercept/README.md
thepassle: I agree with Mikhail, Im still not convinced that `-intercept` adequately describes what this does. _Technically_ the plugin does _intercept_ the module, but that's not really the _point_ of this project, the point is to make modules mockable/overrideable. So maybe it should be more something like `@web/test-runner-mockable-modules` or something like that

packages/test-runner-intercept/README.md
```diff
@@ -0,0 +1,141 @@
...
+  return 'day';
+}
+```
+
+```js
+// test/getTimeOfDay.test.js
+import { interceptModule } from '@web/test-runner-intercept';
+
+const timeLibrary = await interceptModule('time-library');
+const { getTimeOfDay } = await import('../src/getTimeOfDay.js');
```
thepassle, line 95: Ok this example clarifies my question above, i'll resolve that after reviewing the rest

packages/test-runner-intercept/package.json
thepassle: How so? We did already move some of the projects to ESM, like dev-server-storybook. Which scripts would need to be adjusted?

packages/test-runner-intercept/package.json
thepassle: No I mean, because we're in a monorepo, and we have mocha and chai installed in the monorepo root's package.json, we shouldnt have to also add them explicitly here. This also shouldnt be needed for `@web/dev-server-rollup`

packages/test-runner-intercept/package.json
tmsns: No, I meant these scripts:
- [generate-mjs-dts-entrypoints.mjs](https://github.com/modernweb-dev/web/blob/master/scripts/generate-mjs-dts-entrypoints.mjs)
- [generate-ts-configs.mjs](https://github.com/modernweb-dev/web/blob/master/scripts/generate-ts-configs.mjs)

They rely heavily on projects following the same conventions. :-)

I just realized this entrypoint is a node entrypoint, not a browser entrypoint. Most node projects don't use extensions in their exports. Browser exports ofc do. Do you still want me to update it?

bashmish: yeah, I know the scripts :D I meant `reused the same convention used` was referring to that

my point is: don't look at conventions, because we are now changing them and IIRC the idea is to use extensions always, unless the extension will break something, e.g. in the `storybook-framework-web-components` `/preset` the extension broke the storybook for me

I think I tried the scripts with the exports containing extensions, they worked, and if not, you can fix a specific bug in there

tmsns: This is now updated. :-) Can you check and resolve if ok?

packages/test-runner-intercept/package.json
tmsns: This is now updated. :-) Can you check and resolve if ok?

packages/test-runner-intercept/package.json
tmsns: This is now updated. :-) Can you check and resolve if ok?

packages/test-runner-intercept/README.md
tmsns: So, in your example, the "system under test" is the component. (https://en.wikipedia.org/wiki/System_under_test) That system is loaded through a module. (eg. `my-component.js` or `system-under-test.js` in the readme) In your example the component uses a library (eg. the default export of `external-library`), this is loaded using an import, often a static import.

In order for the test file to intercept the call to the library, it needs to do so before the library is loaded. So, this indeed means that in your test file you need to import the component module file **after** the interception. As the interception is a async/await function call, you indeed need to use a dynamic import in your test file after that interception call in order to reference/use the components in your test.

```js
import { interceptModule } from '@web/test-runner-intercept';

// First, intercept
const externalLibrary = await interceptModule('external-library');

// Second, load the system under test.
const systemUnderTest = await import('../../src/system-under-test.js');

// Run tests
...
```

packages/test-runner-module-mocking/tsconfig.json
```diff
@@ -33,6 +33,9 @@
   ],
   "include": [
     "src",
-    "test/**/*.ts"
   ],
+  "exclude": [
+    "browser",
+    "test"
+  ]
```
bashmish, line 40: why do you exclude these?

tsconfig.node-16-base.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.node-base.json",
+  "compilerOptions": {
+    "module": "node16",
+    "moduleResolution": "node16",
+  }
+}
```
bashmish, line 7: why do you need a separate tsconfig, based specifically on node16?

packages/test-runner-module-mocking/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+// server relative module
+await importMockable('/src/library-to-intercept.js');
+```
+
+In order to use regular relative references, `import.meta.resolve()` and `new URL().pathname` can be used.
+
+```javascript
+// relative module
+await importMockable(new URL(import.meta.resolve('../src/library-to-intercept.js')).pathname);
```
bashmish, line 144: I think for transparency it's good to mention here that other ideas were considered, because the approach with explicit dynamic import is causing 2 issues
- you need to change some static imports to dynamic ones to ensure that the mockable module is not loaded by your component/app code earlier than the test makes it mockable (all such imports that potentially load the same module should be rewritten from static to dynamic to be loaded after the `importMockable` call)
- the relative imports are quite painful which is also seen here

We discussed the alternative proposal which could solve both issues: put the mockable import on top of the test, use static import for it and a special import attribute, e.g.:

```
// top of the test file
import json from "../src/library-to-intercept.js" with { type: "module-mockable" };

// all other imports

// test code
```

this solve the issue with dynamic imports, as well as makes it really easy to make anything mockable without running into the import ordering issue which is not easy to debug

but it can also help to solve the relative import issue (basically improve the DX) by instrumenting the code dynamically based on this special `type: "module-mockable"`, similar to what `importMockable` does internally, e.g. rewriting the code to include a query param in the import with the relative path

since this is quite a magic, for now we decided to go with the current proposal

I think it would be nice to explore the alternative proposal in the future after gathering feedback from users about the current one

packages/test-runner-intercept/README.md
tmsns: Can you resolve this conversation if it's ok? 😃

packages/test-runner-module-mocking/tsconfig.json
tmsns: This was just the same as the other projects. I removed it here, as it seems it can build without the exclude list.

tsconfig.node-16-base.json
tmsns: This is needed for the build command in the root `package.json`. If we use the normal tsconfig, the build fails as the project is of type module and uses ES modules.

packages/test-runner-module-mocking/README.md
tmsns: Fair point! I added a section explaining this. 😃

bashmish (approved):
Thanks for your tremendous work Tom!

packages/test-runner-module-mocking/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "directory": "packages/test-runner-module-mocking"
+  },
+  "author": "modern-web",
+  "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/test-runner-module-mocking",
+  "main": "browser/index.js",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./browser/index.d.ts",
+      "import": "./browser/index.js"
```
thepassle, line 21: this can just be `default`, right?

PR #2561, opened 2023-11-17
feat: expose SnapshotStore for extension

jorenbroekema:
## What I did

1.

### DIFF
--- a/.changeset/tame-books-punch.md
+++ b/.changeset/tame-books-punch.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-commands': patch
+---
+
+Expose SnapshotStore, getSnapshotPath and isSaveSnapshotPayload functions, to make extension easier.
--- a/package-lock.json
+++ b/package-lock.json
@@ -35971,7 +35971,7 @@
     },
     "packages/dev-server-storybook": {
       "name": "@web/dev-server-storybook",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "license": "MIT",
       "dependencies": {
         "@babel/core": "^7.16.0",
--- a/packages/test-runner-commands/plugins.mjs
+++ b/packages/test-runner-commands/plugins.mjs
@@ -11,6 +11,9 @@ const {
   a11ySnapshotPlugin,
   filePlugin,
   snapshotPlugin,
+  SnapshotStore,
+  getSnapshotPath,
+  isSaveSnapshotPayload,
 } = cjsEntrypoint;
 
 export {
@@ -23,4 +26,7 @@ export {
   a11ySnapshotPlugin,
   filePlugin,
   snapshotPlugin,
+  SnapshotStore,
+  getSnapshotPath,
+  isSaveSnapshotPayload,
 };
--- a/packages/test-runner-commands/src/index.ts
+++ b/packages/test-runner-commands/src/index.ts
@@ -6,4 +6,11 @@ export { sendKeysPlugin, SendKeysPayload } from './sendKeysPlugin.js';
 export { sendMousePlugin, SendMousePayload } from './sendMousePlugin.js';
 export { a11ySnapshotPlugin, A11ySnapshotPayload } from './a11ySnapshotPlugin.js';
 export { WriteFilePayload, ReadFilePayload, RemoveFilePayload, filePlugin } from './filePlugin.js';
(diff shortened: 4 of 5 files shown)


jorenbroekema:
https://github.com/amzn/style-dictionary/blob/721959203c2755c62a2daa70a4ed34468dbba412/snapshot-plugin/chai-wtr-snapshot.js  here's the snapshot plugin, still fairly work in progress. For mocha runner in WTR context, not that much is needed that isn't already exposed, but for mocha runner in Node context (see `!isBrowser` branches in the code) I do need to setup the snapshot calls with mocha hooks myself.

I used a patch for now to be able to use the 3 snapshotPlugin internals that need exposing for my chai plugin: https://github.com/amzn/style-dictionary/blob/721959203c2755c62a2daa70a4ed34468dbba412/patches/%40web%2Btest-runner-commands%2B0.9.0.patch

bashmish:
Do I understand correctly that you are building smth like WTR's own `snapshotPlugin` (https://github.com/modernweb-dev/web/blob/%40web/test-runner-commands%400.9.0/packages/test-runner-commands/src/snapshotPlugin.ts#L153) which has a WebSockets-based communication logic to pass snapshots to tests executed in the browsers and you need smth similar for Node?

If so, I have 2 quite different thoughts in my mind:
1. WTR is a browser-based testing solution, so it's not intuitive to me that we start adding features for Node-based testing, right? Node has it's own tools for this sort of stuff, including snapshot-based testing, I remember using smth in the past, so maybe you compared your idea with existing solutions?
2. If we are to make some Node specific features, then I'd rather improve the plugin to be usable in both environments rather than export building blocks which to me seem a little be to much to reexport without first cleaning up public API of them which I think wasn't supposed to be exposed since https://github.com/modernweb-dev/web/pull/1501

jorenbroekema:
> Do I understand correctly that you are building smth like WTR's own `snapshotPlugin`?

Yes pretty much. Currently, snapshotPlugin.ts contains 2 parts:
- SnapshotStore & helpers
- snapshotPlugin for WTR which interfaces with WTR through websockets to send messages to the server which controls the SnapshotStore

I maintain a library (style-dictionary) which should run in both the browser and Node, out of the box, and I want the tests ran in chromium and node simultaneously. What this PR does is expose at least the first part (SnapshotStore) so that I may reuse this part while I create a small interface for Mocha Node Runner to communicate with SnapshotStore. Very similar to what's there in snapshotPlugin but a lot simpler because it's not client-to-server communication, just server-server.

In my view, the SnapshotStore is fundamentally really quite agnostic already, it's just a tool that stores snapshots in memory and reads and writes to `.snap.js` files on the filesystem. The alternative for me is to just fork the SnapshotStore and publish it as a separate NPM package and be on my way, but I prefer not to maintain something that's already in here. I do agree that maybe the public API could be cleaned up a bit, and there's a concept of "sessionId" which is the only part that's not really agnostic in the SnapshotStore class (I just pass a dummy ID to it in Node context), but just exposing it as is, is already quite good if you ask me?

@LarsDenBakker seems you built most of this stuff, what's your 2 cents on this? Keep private / WTR specific or expose it? The docs say the snapshot plugin is quite basic, was there ever an intention to expand on it?

For reference, I'd like to get to a similar experience that [Jest Snapshot testing](https://jestjs.io/docs/snapshot-testing) gives you, I've accomplished this somewhat with my Chai assertion plugin which atm works well for both WTR and Node Mocha Runner. If we can expose the snapshot plugin internals for usage outside of WTR, I'd also be happy to contribute some things for making inline snapshotting work (rather than external .snap.js files)

bashmish:
> "sessionId" which is the only part that's not really agnostic in the SnapshotStore class

This was one of my concerns.
Another was that such building blocks don't seem logical for a package like `@web/test-runner-commands`, I think the purpose of it was to export utils you can use in tests and not building blocks to build other utils.

I totally get that it's convenient to run same tests in node and browser, but it's also quite a burden to deliver on this promise in the long run once we take that route here, that's why I'm hesitating.

jorenbroekema:
You reckon we could publish SnapshotStore helper as a separate package then perhaps? Probably as a standalone thing outside of @web scope since it's not web-specific. Or find an existing well-maintained ESM package that does the same thing (but hopefully even better)? I could take a look and see if I can find what Jest is using under the hood for reading/writing snapshots to test (inline) or snapshot files.

PR #2572, opened 2023-12-07
Fix Windows tests

koddsson:
## What I did

1.


bashmish:
When you fix it, shall we configure the build in such a way that Windows build failure prevents PR merges?

koddsson:
> When you fix it, shall we configure the repository in such a way that the Windows build failure prevents PR merges?

I don't think I have the permissions to do that.

PR #2591, opened 2023-12-28
run typescript tests correctly

koddsson:


koddsson:
This is bad and wrong.

PR #2597, opened 2024-01-04
increase timeout in browser tests

koddsson:
## What I did

1.


koddsson:
This seems to be hanging on mock tests but I feel like this is a step in the right direction.

PR #2600, opened 2024-01-05
Convert everything to ESM

koddsson:
I think we just need to pull the bandaid off and deal with issues that arise. Trying to do this piece meal feels impossible.


Westbrook:
This looks exciting! Need to dig deeper on the code but some questions:

Have you tested the prerelease distributions of these packages in an external project? `next` always releases these.

Would we need to also pull the bandaid in the Open WC packages to satisfy a “full” release of these type of changes?

koddsson:
> Have you tested the prerelease distributions of these packages in an external project? `next` always releases these.

I did run it on one project I got at work but maybe I could do more installs. I was hoping other people might also check it out in their projects and report back if there are any errors.

koddsson:
Whoops!

koddsson:
@bashmish; this is now conflicting. Can you resolve those?

bashmish:
> @bashmish; this is now conflicting. Can you resolve those?

done
I had to remove the "retrocompatibility" for `import { globals } from '@storybook/preview/globals';` and it seems much more difficult to support in the ESM world and not really needed given people will easily upgrade to latest storybook packages to resolve it, and I think it's needed for ESM anyway.

PR #2620, opened 2024-01-24
fix(storybook-builder): import both globals and globalsNameReferenceM…

jdevega:
…ap from @storybook/preview/globals

## What I did

1. Import both globals and globalsNameReferenceMap exports from @storybook/preview/globals.
2. Instead of using globals, only use it when globalsNameReferenceMap has no value to preserve retrocompatibility.

### DIFF
--- a/packages/storybook-builder/src/index.ts
+++ b/packages/storybook-builder/src/index.ts
@@ -1,7 +1,9 @@
 import rollupPluginNodeResolve from '@rollup/plugin-node-resolve';
 import { getBuilderOptions } from '@storybook/core-common';
 import { logger } from '@storybook/node-logger';
-import { globals } from '@storybook/preview/globals';
+// Import both globals and globalsNameReferenceMap to prevent retrocompatibility.
+// @ts-ignore
+import { globals, globalsNameReferenceMap } from '@storybook/preview/globals';
 import type { Builder, Options, StorybookConfig as StorybookConfigBase } from '@storybook/types';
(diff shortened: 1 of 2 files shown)


packages/storybook-builder/src/index.ts
```diff
@@ -1,7 +1,8 @@
 import rollupPluginNodeResolve from '@rollup/plugin-node-resolve';
 import { getBuilderOptions } from '@storybook/core-common';
 import { logger } from '@storybook/node-logger';
-import { globals } from '@storybook/preview/globals';
+// @ts-ignore
+import { globals, globalsNameReferenceMap } from '@storybook/preview/globals';
```
thepassle, line 5: wouldnt just the `globalsNameReferenceMap` be enough? Why are we using both?

jdevega: for retrocompatibility. Those using 7.5.3 or lower will have an error if we don't support the previous behaviour.

thepassle: Ah I see. Could you add a comment to the import that we import both for this reason? Just so we'll know why this is in the future :)

jdevega: Done

thepassle: Thanks :) just running CI to make sure everything passes, then we can merge it

jdevega: Yup, fixed linting problems

PR #2632, opened 2024-02-02
Add turboRepo to manage monorepo

Ramon92:
## What I did

1. Add turborepo to manage the mono-repo more efficiently
2. add github cache to cache turbo repo outputs and use turbo repo caching for tests
3. turbo repo caching will help to not run tests in unchanged packages only changed packages will run tests all other packages tests output will be re-played
4. should shave of significant time on node-tests in the pipeline

### DIFF
--- a/.github/workflows/verify-node.yml
+++ b/.github/workflows/verify-node.yml
@@ -47,6 +47,13 @@ jobs:
       - name: Build specific packages for production
         run: npm run build:production
 
+      - name: Cache turbo build setup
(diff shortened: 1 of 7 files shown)


koddsson:
I'm not convinced that the effort to learn a whole new systems is offset by it's benefits. I'm also really scared of cache in developer setup and CI in case the cache invalidation is misconfigured or buggy.

Are the benefits just that CI will be faster? By how much?

PR #2634, opened 2024-02-03
chore(deps): upgrade `@rocket/cli` to latest version

jonkoops:
Upgrades `@rocket/cli` to latest version, as previous versions depend on native modules that cannot be built on Apple Silicon, causing errors when running `npm install`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       ],
       "dependencies": {
         "@rocket/blog": "^0.4.0",
-        "@rocket/cli": "^0.10.1",
+        "@rocket/cli": "^0.21.0",
         "@rocket/core": "^0.1.2",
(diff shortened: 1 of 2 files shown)


jonkoops:
Closing this PR as it breaks stuff when running `build:site`, looks like this is not part of the CI, so it all seems green. There are too many breaking changes in the various `@rocket/` packages, with too little documentation on how to resolve them for me to get this fixed.

For example, the `@rocket/search` is suddenly no longer exporting the `rocketSearch` function, no idea why. There is nothing about this in the [changelog](https://github.com/modernweb-dev/rocket/blob/main/packages/search/CHANGELOG.md#070) and the link to the migration guide and site are dead.

Closing this PR, as I don't think this will be an easy fix with all the migration that might be required.

PR #2635, opened 2024-02-03
chore: replace `@esm-bundle/chai` with regular `chai` package

### LINKED ISSUE #2633: Use normal `chai` package over `@esm-bundle/chai`
The `@esm-bundle/chai` dependency [recommended in the documentation](https://modern-web.dev/guides/test-runner/getting-started/#setup-tools) has not received maintenance for several years. The official `chai` package has supported ESM since version 4.3.0, so it is therefore no longer required to use this custom build.

jonkoops:
Closes #2633

### DIFF
--- a/docs/blog/introducing-web-test-runner/index.md.review
+++ b/docs/blog/introducing-web-test-runner/index.md.review
@@ -33,7 +33,7 @@ This is the minimal instruction on how to start using the web test runner.
 1. Install the necessary packages
 
    ```
-   npm i --save-dev @web/test-runner @esm-bundle/chai
+   npm i --save-dev @web/test-runner chai
    ```
 
 2. Add a script to your `package.json`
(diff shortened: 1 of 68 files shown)


koddsson:
Looks good to me! I couldn't get this working without converting the package to ESM first so I'm surprised that this is working. If CI is green I'll merge it :)

jonkoops:
> Looks good to me! I couldn't get this working without converting the package to ESM first so I'm surprised that this is working. 

We're using the regular package [in Leaflet](https://github.com/Leaflet/Leaflet/pull/9242), but we are using plain import maps rather than `--node-resolve`, not sure if that affects things.

jonkoops:
Tests [are failing](https://github.com/modernweb-dev/web/actions/runs/7769718134/job/21312343273?pr=2635), looks like the path to `chai` is re-written to `node_modules/chai/index.js`, but this is not correct. The [`main` entry](https://github.com/chaijs/chai/blob/640d93221f1781a8e2fe3f25a892f01bdbeb86ba/package.json#L29) of `chai` is supposed to resolve to `node_modules/chai/chai.js`. Looks like this might be a bug in the Node.js-style resolution?

PR #2642, opened 2024-02-19
docs(test-runner-module-mocking): ✏️ test-moduleMocking details on plugin import order

movrack:
docs(test-runner-module-mocking): ✏️ test-moduleMocking details on plugin import order

Adding details on the order of plugin import in test config. Module mocking must be in first position to avoid conflict with modified path due to other plugins. + refactoring

## What I did

1. Adding doc in readme
2. Small refactoring for easiness code view
3. Adding Sinon.js as plugin dep dependency for new test on relative path and try stubbing mocked dependency

### DIFF
--- a/packages/test-runner-module-mocking/README.md
+++ b/packages/test-runner-module-mocking/README.md
@@ -38,6 +38,8 @@ export default {
 };
 ```
 
+Take care of the plugins order, some plugins can modify path of imported modules and moduleMockingPlugin may be not able to retreive them correctly, so it's advise to use it first in the list.
+
 ### Simple test scenario
 
 ```js
--- a/packages/test-runner-module-mocking/src/moduleMockingPlugin.ts
+++ b/packages/test-runner-module-mocking/src/moduleMockingPlugin.ts
@@ -37,38 +37,7 @@ export function moduleMockingPlugin(): Plugin {
           const namedExports = exports.map(e => e.n).filter(n => n !== 'default');
           const hasDefaultExport = exports.some(e => e.n === 'default');
 
-          body = `
-          import * as original from '${url.pathname}';
-          const newOriginal = {...original};
-
-          ${namedExports.map(x => `export let ${x} = newOriginal['${x}'];`).join('\n')}
-
-          ${
-            hasDefaultExport
-              ? `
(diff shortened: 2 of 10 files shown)


packages/test-runner-module-mocking/src/moduleMockingPlugin.ts
```diff
@@ -37,15 +37,37 @@ export function moduleMockingPlugin(): Plugin {
           const namedExports = exports.map(e => e.n).filter(n => n !== 'default');
           const hasDefaultExport = exports.some(e => e.n === 'default');
 
-          body = `
+          body = generateInterceptedModuleBody(body, url, namedExports, hasDefaultExport);
```
tmsns, line 40: There is no real reason to extract this in a separate method. Can you revert this change?

packages/test-runner-module-mocking/test/moduleMockingPlugin.test.ts
```diff
@@ -20,6 +20,14 @@ describe('moduleMockingPlugin', function test() {
     });
   });
 
+  it('can intercept server relative modules with stub', async () => {
```
tmsns, line 23: All tests are using stubs. :-)
Example: https://github.com/modernweb-dev/web/blob/2d933bcf7d8d4a1b2f67dcd6413985bb4c692356/packages/test-runner-module-mocking/test/fixtures/server-relative/browser-test.js#L16

We don't want to test all different stubbing libraries.

So, you can remove this extra test.

packages/test-runner-module-mocking/README.md
```diff
@@ -38,6 +38,8 @@ export default {
 };
 ```
 
+Take care of the plugins order, some plugins can modify path of imported modules and moduleMockingPlugin may be not able to retreive them correctly, so it's advise to use it first in the list.
```
tmsns, line 41:
```suggestion
Make sure `moduleMockingPlugin()` is the first plugin defined! 
(the interception mechanism of `moduleMockingPlugin()` needs to run before other plugins (eg. `rollupAdapter()`) transform the import resolutions within the `resolveImport()` hook)
```

Issue #2646, opened 2024-02-21
Used Puppeteer version is deprecated


Sanderovich:
I am running into the issue that @web/test-runner-puppeteer only supports Puppeteer version 20.x, which means it is running Chrome version 115. Therefore, I am not able to automatically test my package on the latest version of Chrome (version 121 at the time of writing). Additionally, Puppeteer deprecated all versions lower than 21.5.0.
```bash
npm WARN deprecated puppeteer@20.0.0: < 21.5.0 is no longer supported
```

For now, I force web-test-runner to use the newest Puppeteer by adding `overrides` to my `package.json`.
```json
"overrides": {
    "puppeteer": "22.1.0"
  }
```

To continue, I could not find upcoming plans to start updating Puppeteer or a reason for why Puppeteer is not updated yet, so I made a pull request to update Puppeteer to its latest major (See #2647).

web-padawan:
Fixed by #2647 and released in `@web/test-runner-puppeteer` 0.16.0 (for some reason this issue wasn't closed).

PR #2647, opened 2024-02-21
Update puppeteer to version 22.0.0

Sanderovich:
Updated Puppeteer to its latest major. See [#2646](https://github.com/modernweb-dev/web/issues/2646) for more context


package-lock.json
```diff
@@ -9338,6 +9256,18 @@
         "node": "*"
       }
     },
+    "node_modules/ast-types": {
+      "version": "0.13.4",
+      "resolved": "https://nexus3.topdesk.com/nexus/repository/npm/ast-types/-/ast-types-0.13.4.tgz",
```
koddsson, line 9261: Looks like you got your npm config set to a custom registry

koddsson (changes requested):

packages/browser-logs/test/serialize-deserialize.test.ts
```diff
@@ -14,7 +14,7 @@ describe('serialize deserialize', function () {
   let browser: Browser;
   let page: Page;
   before(async () => {
-    browser = await puppeteer.launch({ headless: 'new' });
+    browser = await puppeteer.launch({ headless: true });
```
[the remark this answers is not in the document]
Sanderovich: It is part of the puppeteer 22.0.0 changelog (https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/CHANGELOG.md#2200-2024-02-05) item "enable the new-headless mode by default [#11815](https://github.com/puppeteer/puppeteer/issues/11815)".

The default headless mode is now 'new', so we can migrate to `headles: true`.

package-lock.json
Sanderovich: Good catch, I completely overlooked this, thanks! I re-created the package-lock.json

Sanderovich:
@koddsson Thanks for the review! I resolved you comments

packages/test-runner-puppeteer/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export { puppeteerLauncher } from './puppeteerLauncher.js';
-export { devices } from 'puppeteer';
+export { KnownDevices } from 'puppeteer';
```
bashmish, line 2: this is kind a breaking change? as such good to document in the migration guide?

but if we do it anyway, I'd rather remove this export, as people should import this directly from `puppeteer`, wdyt?

koddsson:
> @koddsson Thanks for the review! I resolved your comments

The `package-lock.json` was a bit messed up and I just pushed a commit that I think fixes it.

packages/test-runner-puppeteer/src/index.ts
Sanderovich: Good question. Thinking about it, I do not see a clear advantage of reexporting it, so I agree with you that letting people import it from puppeteer directly would be the better option. But this will be a breaking change I guess, so how should I approach it?

koddsson: @bashmish; any input here?

bashmish: actually `@web/test-runner-puppeteer` is on 0.x, so keep it `minor` in the changeset, since in SemVer resolution it will behave like major update
the outcome should be a bump from 0.15.0 to 0.16.0
mention this change in the MIGRATION.md

Sanderovich: @bashmish I removed the KnownDevices export, created the changeset, and added a MIGRATION.md. I could not find an existing MIGRATION.md, so I created a new one and kept it as simple as possible.

Sanderovich: In addition, I had to re-create the package-lock.json again, because it had conflicts with master. Could you double check it is still ok?

MIGRATION.md
```diff
@@ -0,0 +1,9 @@
+# Migration
+
+## From version 0.15.x to 0.16.0
+
+The re-exports of Puppeteer devices is removed and can be imported from Puppeteer directly.
+
+```ts
+import { KnownDevices } from 'puppeteer';
+```
```
bashmish, line 9: I think this migration file needs to be in the package impacted by the change, not in the root

bashmish (changes requested):

MIGRATION.md
Sanderovich: @bashmish Good point, I deleted the MIGRATION.md in the root and created one in the relevant packages

bashmish:
Just realised that we might have a bunch of docs referencing the `devices`, and turned out to be true. Can you please check and update those? I think this means not just removing, but changing the documentation to use new recommended imports from puppeteer.

Sanderovich:
@bashmish I updated the imports in the docs. Additionally, I see the playwright packages are also exporting a devices variable, but I kept those as is to keep the focus of the PR on updating Puppeteer.

.changeset/curly-snails-rush.md
```diff
@@ -0,0 +1,9 @@
+---
+'@web/test-runner-puppeteer': minor
+'@web/test-runner-chrome': minor
+'@web/dev-server-hmr': minor
+'@web/browser-logs': minor
+'@web/dev-server': minor
```
bashmish, line 6: I think you need to bump only

```
'@web/test-runner-puppeteer': minor
'@web/test-runner-chrome': minor
```

?

other packages don't have changes in the API, you only updated the devDependencies

bashmish:
I think it's the very last comment about the version you need to bump, then we can merge this
I wasn't sure what packages would be impacted by the `devices` change, so needed to see your last changes brefore bringing this up

.changeset/curly-snails-rush.md
Sanderovich: Update the changeset :smile:

bashmish:
@Sanderovich thanks a lot!

PR #2648, opened 2024-02-22
Set fetchPriority of polyfills to high

koddsson:
I set the `fetchPriority` of polyfill scripts to `high` to hint to the browser that these are important scripts that should have a high priority.

For browsers that don't support this attribute this operation is a no-op.

## Reading 📖 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority


thepassle (approved):

koddsson:
The Windows tests are failing because they can't download puppeteer. I'm not sure why but I figured it's some server issues and it's fine to merge.

PR #2664, opened 2024-03-11
feat(rollup-plugin-html): resolves assets in styles

bashmish:
## What I did

1. rollup-plugin-html resolves assets in styles


packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -1041,4 +1041,29 @@ describe('rollup-plugin-html', () => {
       ].join(''),
     );
   });
+
+  it.only('includes referenced assets in the bundle', async () => {
```
bashmish, line 1045: this is marked as `only`

to run tests and see the logs of this test, run

```
npm --prefix packages/rollup-plugin-html run test:node
```

packages/rollup-plugin-html/src/output/emitAssets.ts
```diff
@@ -57,7 +82,54 @@ export async function emitAssets(
...
+          const { code } = await transform({
+            filename: basename,
+            code: asset.content,
+            minify: false,
+            visitor: {
+              Url: url => {
+                // Support foo.svg#bar
+                const [filePath, idRef] = url.url.split('#');
+
+                if (shouldHandleAsset(url.url)) {
```
bashmish, line 98: can you pass `filePath` instead of `url` and simplify the extension check for the SVG?

packages/rollup-plugin-html/src/output/emitAssets.ts
```diff
@@ -57,7 +82,54 @@ export async function emitAssets(
...
+      const emittedExternalAssets = new Map();
       if (asset.hashed) {
+        if (basename.endsWith('.css') && options.bundleAssetsFromCss) {
+          let updatedCssSource = false;
+          const { code } = await transform({
+            filename: basename,
+            code: asset.content,
+            minify: false,
+            visitor: {
+              Url: url => {
```
bashmish, line 94: does this recursively follow all css files includes (the `@import` ones)?

if yes, looks like `source = Buffer.from(code);` would not update anything but the entry point

packages/rollup-plugin-html/test/fixtures/resolves-assets-in-styles-images/styles.css
```diff
@@ -0,0 +1,24 @@
+#a {
+background-image: url("images/star.svg");
```
bashmish, line 2: nitpick: can you run prettier over this and other changed files?

packages/rollup-plugin-html/src/output/emitAssets.ts
```diff
@@ -9,6 +11,29 @@ export interface EmittedAssets {
...
+  /.*\.svg(#[A-Za-z][A-Za-z0-9\-_:.]*)?/,
+  /.*\.png/,
+  /.*\.jpg/,
+  /.*\.jpeg/,
+  /.*\.webp/,
+  /.*\.gif/,
+  /.*\.avif/,
+  /.*\.woff2/,
+  /.*\.woff/,
+];
```
bashmish, line 25: why do you want to use the allowed extensions list? are there some extensions that should not be handled?

bashmish:
super nice work Pascal!
I'm just concerned about 2 things
- more about the allowed extensions list
- less about the recursive css imports handling - this one we can also add in the future, but not sure how many need this as css nested imports is not a super nice performance strategy and might not be used much

packages/rollup-plugin-html/src/output/emitAssets.ts
thepassle: The code as currently implemented does not _bundle_ css via `@import`. I looked at both esbuild and lightningcss, and they have a `bundle` for css, but it _only_ follows `@import`, and not `url`, which is what we need. So for now I only added support for assets referenced via `url`, afaik you can't add .css files in `url`. In the future we could consider bundling imported css

packages/rollup-plugin-html/src/output/emitAssets.ts
thepassle: We're not really sure what people may add in `url`, so I figured be on the safe side

packages/rollup-plugin-html/src/output/emitAssets.ts
thepassle: fixed

thepassle:
@bashmish im gonna go ahead and merge this (it will take some time to end up in our internal npm feed as well, and i'd really like to wrap this up today), the other things we can do in another iteration/separate PR

packages/rollup-plugin-html/src/output/emitAssets.ts
bashmish: with current approach we'll need to maintain a list of extensions and update when new types appear or if we forgot to include smth

instead I'd rather prefer to have a `blocklist` and make it configurable, we can do that later too

packages/rollup-plugin-html/src/output/emitAssets.ts
bashmish: clear, makes sense!

Issue #2631, opened 2024-02-01
[rollup-plugin-html]  No clearly documented /  easy way to ignore modules


wlinna:
Problem: there's no documented way to prevent `rollup-plugin-html` from reading modules from an HTML file and bundling them. The module is processed even if I remove the module in `transformHtml`.

Proposal: run `transformHtml` before bundling or even better, introduce an option that lets users prevent the bundling logic from running.

PR #2671, opened 2024-03-15
feat(rollup-plugin-html): glob patterns exclusion for external assets

### LINKED ISSUE #811: External does not seem to work with @web/rollup-plugin-html
I'd like to be able to reference assets (scripts) with absolute paths that should be ignored in the rollup + html build. 

As an example, I have the following HTML code:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <script type="module" src="/path/module.js"></script>
  </head>
  <body>
  </body>
</html>
```

The ES module (a webcomponent) `/path/module.js` should not be touched / bundled by rollup. Defining `external: ['/path/module.js']` in my `rollup.config.js` gives this error:

```
index.html → dist...
[!] Error: Could not resolve entry module (path/module.js).
Error: Could not resolve entry module (path/module.js).
```

A simple test case can be found at https://github.com/peschee/rollup-external-html-test

JulianCataldo:
Fixes #811

## What I did

1. Added a plugin option for providing glob patterns in order to exclude assets from being extracted.

Very useful for stuff that already lives in a `public` folder and are copied as-is.

## Why?

`extractAssets` is all or nothing.

`transformAssets` doesn't provide a way to skip an asset.

Also related: #2631

### DIFF
--- a/packages/rollup-plugin-html/src/assets/utils.ts
+++ b/packages/rollup-plugin-html/src/assets/utils.ts
@@ -1,5 +1,6 @@
 import { Document, Element } from 'parse5';
 import path from 'path';
+import picomatch from 'picomatch';
 import { findElements, getTagName, getAttribute } from '@web/parse5-utils';
 import { createError } from '../utils.js';
 import { serialize } from 'v8';
@@ -142,4 +143,12 @@ export function getSourcePaths(node: Element) {
 
 export function findAssets(document: Document) {
   return findElements(document, isAsset);
+}
+
+// picomatch follows glob spec and requires "./" to be removed for the matcher to work
+// it is safe, because with or without it resolves to the same file
+// read more: https://github.com/micromatch/picomatch/issues/77
+const removeLeadingSlash = (str: string) => (str.startsWith('./') ? str.slice(2) : str);
+export function createAssetPicomatchMatcher(glob?: string | string[]) {
+  return picomatch(glob || [], { format: removeLeadingSlash });
 }
--- a/packages/rollup-plugin-html/test/fixtures/assets/styles-with-referenced-assets.css
+++ b/packages/rollup-plugin-html/test/fixtures/assets/styles-with-referenced-assets.css
@@ -1,0 +1,15 @@
+#a1 {
+  background-image: url('image-a.png');
+}
+
+#a2 {
+  background-image: url('image-a.svg');
+}
+
+#d1 {
+  background-image: url('./image-d.png');
+}
+
+#d2 {
+  background-image: url('./image-d.svg');
+}
--- a/.changeset/nervous-bugs-rhyme.md
+++ b/.changeset/nervous-bugs-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@web/rollup-plugin-html': minor
+---
+
+glob patterns exclusion for external assets
--- a/docs/docs/building/rollup-plugin-html.md
+++ b/docs/docs/building/rollup-plugin-html.md
@@ -362,6 +362,8 @@ export interface RollupPluginHTMLOptions {
   transformHtml?: TransformHtmlFunction | TransformHtmlFunction[];
   /** Whether to extract and bundle assets referenced in HTML. Defaults to true. */
   extractAssets?: boolean;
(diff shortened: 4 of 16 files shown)


bashmish (changes requested):
I need this also in https://github.com/modernweb-dev/web/pull/2677

PR looks good, I think we just need a few things:
- write a simple unit test
- I'd rename `external` to `externalAssets, since the whole config is flat and `external` can refer to anything, so better make it more specific
- use currently installed picomatch version (because I see types version mismatch and I think it's more handy to do an upgrade to picomatch 4 alltogether for all deps)
- update lock file (I think npm install should have done it, but it wasn't included)

can you work on this today-tomorrow? if not, I'd like to work on this myself, since I need this kinda urgently

JulianCataldo:
Hey!
I am on my phone so I must keep it short. 
I had troubles setting up the monorepo so my strategy was to monkey patch the dist then backport the changes to the source. 
I've seen a missing ´lightingcss' dep. too on a (unreleased?) version. 

I agree on your suggestions. 
I was waiting for a first feedback before engaging additional work and polishing. 

If it wasn't for the monorepo setup I would have gave a stab at writing a test too, but I'm a bit lost TBH.

If your familiar with working with this repo., maybe it's better that you continue this PR.

Cya

bashmish:
@JulianCataldo I'm familiar indeed, some parts I maintain since recently, this code is new to me, but you did most of it anyway, the rest I can do quickly, so I'll take it from here then. Thanks!

packages/rollup-plugin-html/src/assets/utils.ts
```diff
@@ -143,3 +144,11 @@ export function getSourcePaths(node: Element) {
...
   return findElements(document, isAsset);
 }
+
+// picomatch follows glob spec and requires "./" to be removed for the matcher to work
+// it is safe, because with or without it resolves to the same file
+// read more: https://github.com/micromatch/picomatch/issues/77
+const removeLeadingSlash = (str: string) => (str.startsWith('./') ? str.slice(2) : str);
+export function createAssetPicomatchMatcher(glob?: string | string[]) {
+  return picomatch(glob || [], { format: removeLeadingSlash });
+}
```
bashmish, line 154: this was pretty unexpected, but the proposed workaround from the picomatch maintainer seems to be good for us
https://github.com/micromatch/picomatch/issues/77

I added different variations in the test, with and without `./`, to be sure it works

packages/rollup-plugin-html/test/fixtures/assets/styles-with-referenced-assets.css
```diff
@@ -0,0 +1,15 @@
...
+#a2 {
+  background-image: url('image-a.svg');
+}
+
+#d1 {
+  background-image: url('./image-d.png');
+}
+
+#d2 {
+  background-image: url('./image-d.svg');
```
bashmish, line 14: some deduplication of images was happening, I think because we try to identify images with same hash and merge them into one, so it's a feature of this plugin
but it made testing the exclusion cumbersome, so I added a new `image-d.svg/png` with the content equal to the file name to get a different hash than for `image-a` and see in the output all expected exclusions

the deduplication continues to work correctly, it's just that test would be hard to read if I mix so much different functionality in one place

bashmish (approved):

packages/rollup-plugin-html/src/assets/utils.ts
```diff
@@ -143,3 +144,11 @@ export function getSourcePaths(node: Element) {
 export function findAssets(document: Document) {
   return findElements(document, isAsset);
 }
+
+// picomatch follows glob spec and requires "./" to be removed for the matcher to work
+// it is safe, because with or without it resolves to the same file
+// read more: https://github.com/micromatch/picomatch/issues/77
+const removeLeadingSlash = (str: string) => (str.startsWith('./') ? str.slice(2) : str);
```
thepassle, line 151: can't we just `path.normalize` it?

bashmish: I wanted to keep modifications to the original paths used in source files to a minimum

PR #2692, opened 2024-04-08
fix(storybook-builder): simplify and speed up the CommonJS to ESM transformation

bashmish:
## What I did

1. simplify and speed up the CommonJS to ESM transformation
2. make React conditional reexports work in production
3. remove `@chialab/esbuild-plugin-commonjs` and use low-level `cjs-module-lexer`

2nd caused a bug when I started working on MDX and created a static build of the storybook, which made React break with missing reexports like `import { createContext } from 'react';`
this was due to a lack of support in `@chialab/esbuild-plugin-commonjs` for conditional CommonJS reexports, specifically in this file https://www.runpkg.com/?react@18.2.0/index.js

```js
// react@18.2.0/index.js
'use strict';

if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react.production.min.js');
} else {
  module.exports = require('./cjs/react.development.js');
}
```

after hours of debugging, seeing the output of `@chialab/esbuild-plugin-commonjs` and trying to make a patch for it, I gave up an decided to try to make smth similar myself with the help of low-level tools, where https://www.npmjs.com/package/cjs-module-lexer seems to be the most mature and was used by `@chialab/esbuild-plugin-commonjs` under-the-hood

I think the end result is better than the previous in all aspects (simplicity, speed and the readability of the output) and it solves the bug of MDX which I'm working on in another branch

### DIFF
--- a/packages/storybook-builder/src/esbuild-plugin-commonjs-named-exports.ts
+++ b/packages/storybook-builder/src/esbuild-plugin-commonjs-named-exports.ts
@@ -1,35 +1,86 @@
 import type { Plugin } from 'esbuild';
+import { readFile } from 'fs-extra';
+import { dirname } from 'path';
 
-export function esbuildPluginCommonjsNamedExports(module: string, namedExports: string[]): Plugin {
+export function esbuildPluginCommonjsNamedExports(modules: string[]): Plugin {
   return {
     name: 'commonjs-named-exports',
-    setup(build) {
-      build.onResolve({ filter: new RegExp(`^${module}$`) }, args => {
+    async setup(build) {
+      const { init, parse } = await import('cjs-module-lexer');
+      await init();
+
+      build.onResolve({ filter: new RegExp(`^(${modules.join('|')})$`) }, async args => {
+        if (args.pluginData?.preventInfiniteRecursion) return;
+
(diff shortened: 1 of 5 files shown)


packages/storybook-builder/src/esbuild-plugin-commonjs-named-exports.ts
```diff
@@ -1,35 +1,76 @@
...
+      build.onResolve({ filter: new RegExp(`^(${modules.join('|')})$`) }, async args => {
+        if (args.pluginData?.preventInfiniteRecursion) return;
+
+        const { path, ...rest } = args;
+        rest.pluginData = { preventInfiniteRecursion: true };
+        const resolveResult = await build.resolve(path, rest);
+        const resolvePath = resolveResult.path;
+
+        // skip if resolved to an ESM file
+        if (resolvePath.endsWith('.mjs')) return;
```
LarsDenBakker, line 21: what if it's a project with type module?

bashmish: it's OK, the lexer would try to parse them and if it works, then it will give the reexports

PR #2700, opened 2024-04-16
test-runner-module-mocking: bug when loaded normally in previous test before intercepted in next

bashmish:
## What I did

1. provided a minimal reproduction of the bug

We discussed a similar situation, but different, when the order of mocking and importing was relevant within a single test.
But between 2 independent test files such problem wasn't discussed thoroughly, so I consider this to be a bug.
I hope my reproduction sheds some light on what goes wrong.

The biggest difference is that within a single test file you can make any order of loading modules you want, therefore this solution was OK.
But between 2 files I can't really guarantee this. In my repository I have tens of test files, many of them import reusable utilities. The module graph can be quite random depending on the unit test entry point.
The only solution I see is to have a special entry point for all unit tests which is guaranteed to be first loaded which will mock all necessary modules beforehand, but I don't think this is an acceptable solution, it scatters the test logic. We can as well just make a configuration for this then, but still same problem with scattering the test logic.

Fun fact: I was even able to use `test-runner-module-mocking` first without any issue, before smth changed in the order of test files loading, and caused this bug. Not so fun actually, as I spent a few hours debugging this and finding the root cause.

### DIFF
--- a/packages/test-runner-module-mocking/test/fixtures/bug-cache/browser-test-1.js
+++ b/packages/test-runner-module-mocking/test/fixtures/bug-cache/browser-test-1.js
@@ -1,0 +1,14 @@
+import { expect } from '../chai.js';
+
+// this module gets improted here first and gets cached by the dev server
+// resulting in `import { getCurrentHour } from './time-library.js';` being cached too
+// the following test intercepts "time-library.js", but that has no impact on the cached "getTimeOfDay.js"
+// so the mocking won't work for any module we need to intercept inside "getTimeOfDay.js" in any following test
+import { getTimeOfDay } from './fixture/getTimeOfDay.js';
+
+describe('1st test', () => {
+  it('works', () => {
+    const timeOfDay = getTimeOfDay();
+    expect(typeof timeOfDay === 'string').to.equal(true);
+  });
+});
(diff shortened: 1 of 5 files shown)


tmsns:
I have seen this problem as well.
It seems that loaded modules are shared across runs of test files, across different wtr-sessions. This is weird, of course, as each session is normally running in its own context.

Do you have the issue when you disable concurrency? (ie set to 1) I remember this worked for me.

bashmish:
> I have seen this problem as well. It seems that loaded modules are shared across runs of test files, across different wtr-sessions. This is weird, of course, as each session is normally running in its own context.
> 
> Do you have the issue when you disable concurrency? (ie set to 1) I remember this worked for me.

Setting `concurrency: 1` doesn't help.
I also tried to disable ETag plugin, to fully disable browser caching, it didn't help either.
Because it's not cached only in the browser, but WDS caches it first time when module is loaded, and if it's mocked after that, then it doesn't transform it again, resulting in mocking not having any effect.

I came across this today https://rollupjs.org/plugin-development/#shouldtransformcachedmodule, maybe we can look into invalidating the cache next time the same module is being loaded, if it needs mocking now. I don't now who exactly caches in WDS, maybe it's one of rollup plugins, maybe it's a WDS plugin. Just throwing ideas here.

PR #2712, opened 2024-04-23
fix: windows tests

thepassle:
## What I did

1.


packages/rollup-plugin-html/test/rollup-plugin-html.test.ts
```diff
@@ -1257,8 +1257,8 @@ describe('rollup-plugin-html', () => {
     expect(output.length).to.equal(8);
 
     const expectedAssets = [
-      'assets/image-a.png',
-      'assets/image-d.png',
+      'image-a.png',
+      'image-d.png',
```
bashmish, line 1261: why is output changed?

thepassle: because we're comparing the `name` (which is 'image-a.png'), instead of the `fileName` (which is 'assets/image-a.png')

PR #2716, opened 2024-04-24
js-tests.md: Refer to chai as an ES module

olleolleolle:
Since 5.0, chai is distributed as an ES module, no need for the wrapper.

## What I did

1. I modernized the reference to `chai` in the example.

### DIFF
--- a/docs/docs/test-runner/writing-tests/js-tests.md
+++ b/docs/docs/test-runner/writing-tests/js-tests.md
@@ -5,7 +5,7 @@ Javascript files are loaded by the test framework that is configured. The defaul
 For example:
 
 ```js
-import { expect } from '@esm-bundle/chai';
+import { expect } from 'chai';
 import { myFunction } from '../src/myFunction.js';
 
 describe('myFunction', () => {


bashmish:
There are many more mentions of `chai` in the code.
In the documentation should be easy to do that.
In the code and tests I think this is a required change too then, also to verify all works with chai 5.x.
I think these changes should go as a whole.

olleolleolle:
@bashmish Thanks! I took a look, learned a little about the structure, and then saw this was too big of a change for me to do without knowing more about the project. 

If someone wants to carry this change, be welcome to it!

bashmish:
I think this is a good first issue. It's impactful, but if chai 5 is compatible, it should be relatively easy to do.
This can help to start changing and prepare a PR https://github.com/modernweb-dev/web/blob/master/CONTRIBUTING.md

bashmish (changes requested):
`@esm-bundle/chai` needs to be replaced with `chai` native ESM everywhere in the docs, code and tests.

PR #2740, opened 2024-05-27
Fix storybook-builder and storybook-framework-web-components on Windows

bashmish:
## What I did

1. setup testing on Windows


robrez:
I encountered the following error today while trying to convert one of my repos to use the builder.  Does this PR fix this error?

> Storybook failed to load the following preset: @web\storybook-framework-web-components\preset

bashmish:
@robrez  it will, but I'm still busy working on that, today gonna have some focus
it's also gonna be solved in Storybook itself, given they triaged and marked it as bug https://github.com/storybookjs/storybook/issues/27227, but our builder is still on Storybook 7, so for that I'll make a workaround, stay tuned :)

bashmish:
Unfortunately I can't find a good workaround, so need to wait for Storybook to fix the "preset" issue first.
And as a result will park this PR until then.

Meanwhile, you can try using WSL to run your project with Storybook on Windows, it works well (I tested), but it's no suprise given it's just Ubuntu under the hood.

PR #2741, opened 2024-05-29
feat(mocks): update mocks addon styles to support dark mode

kateutlik:
## What I did

1. Use hsla values for background and border color so it works for both - light and dark mode:

image

image

2. Slightly improve styles when loading / no mocks:

image

image


thepassle:
Looks much better, thanks kate!

PR #2745, opened 2024-06-02
 added new things as const internalDependencyMap = new Map();

omeratsiz:
## What I did

1. added new things as const internalDependencyMap = new Map();


bashmish:
seems like some ChatGPT/Copilot non-sense, closing

PR #2749, opened 2024-06-07
Update migration-to-storybook-7.md

jdevega:
As Storybook 8 is out there, users need to use the right version to migrate.

## What I did

1. Change `latest` to `7` to avoid using the upgrade command from the wrong CLI.


bashmish (approved):

bashmish:
Thanks!

PR #2752, opened 2024-06-13
fix: replace ip dependency due to security bug CVE-2024-29415

### LINKED ISSUE #2747: [dev-server]  CVE-2024-29415: "ip SSRF improper categorization in isPublic"
This issue is just meant to raise the flag about the current vulnerability with the `ip` package that `@web/dev-server` depends on. It seems that there currently is no fix for this from `ip`.
https://github.com/advisories/GHSA-2p57-rm9w-gvfp

bashmish:
## What I did

Fixes #2747 

`ip` package is now archived on GitHub https://github.com/indutny/node-ip/
PRs with a fix won't be released
image

### DIFF
--- a/packages/dev-server/package.json
+++ b/packages/dev-server/package.json
@@ -65,7 +65,7 @@
     "command-line-usage": "^7.0.1",
     "debounce": "^1.2.0",
     "deepmerge": "^4.2.2",
-    "ip": "^2.0.1",
+    "internal-ip": "^6.2.0",
     "nanocolors": "^0.2.1",
     "open": "^8.0.2",
     "portfinder": "^1.0.32"
--- a/packages/test-runner-browserstack/src/browserstackLauncher.ts
+++ b/packages/test-runner-browserstack/src/browserstackLauncher.ts
@@ -1,7 +1,7 @@
 import { BrowserLauncher, TestRunnerCoreConfig } from '@web/test-runner-core';
 import { WebdriverLauncher } from '@web/test-runner-webdriver';
 import browserstack from 'browserstack-local';
-import ip from 'ip';
+import internalIp from 'internal-ip';
 import {
   registerBrowserstackLocal,
   unregisterBrowserstackLocal,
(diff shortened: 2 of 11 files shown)


thepassle (approved):

thepassle (approved):

packages/test-runner-browserstack/src/browserstackLauncher.ts
```diff
@@ -14,7 +14,7 @@ export interface BrowserstackLauncherArgs {
 }
 
 const REQUIRED_CAPABILITIES = ['name', 'browserstack.user', 'browserstack.key', 'project', 'build'];
-const localIp = ip.address();
+const localIp = internalIpV4Sync() || '0.0.0.0';
```
bashmish, line 17: we never covered the `undefined` case, but new library has types, so the error on this line and in a similar case in `packages/test-runner-saucelabs/src/SauceLabsLauncher.ts` appeared

using fallback `0.0.0.0` seems acceptable for a test runner, although it's not a good practice in general

packages/dev-server/package.json
```diff
@@ -65,7 +65,7 @@
     "command-line-usage": "^7.0.1",
     "debounce": "^1.2.0",
     "deepmerge": "^4.2.2",
-    "ip": "^2.0.1",
+    "internal-ip": "^6.2.0",
```
bashmish, line 68: had to use 6.2.0 since it's a CommonJS module which works with our current CommonJS output
as soon as we go to ESM for our modules, we can try 7.0.0

mgriffin-scottlogic:
Thanks for addressing this, does anyone know how long it might take to make it into a release? I'd like to update my dependencies to avoid this security bug as well.

packages/test-runner-browserstack/src/browserstackLauncher.ts
bashmish: checked with my colleague, and we decided it's not really a good idea
I also checked the sources of the package a bit and see that the "undefined" is more of an edge case for the browser usage, otherwise it's hard to see where else it might happen, so the next best decision is just to throw in case it's undefined

locally the types work with such if condition, hopefully the build passes too, then I merge this

PR #2758, opened 2024-06-21
[feature] Added a CLI flag to allow passing without any test files found

### LINKED ISSUE #1331: Allow to pass if no tests
If I run e.g. `web-test-runner "src/**/*.test.tsx"` and there are no files matching that pattern, I get this error:
```
Error: Could not find any test files with pattern(s): src/**/*.test.tsx
```

Would be nice if there was a flag that could be passed to allow it to pass if no files are found. This is useful for scaffolds and other initial stages.

glektarssza:
## What I did

1. Added a CLI flag that allows passing without any test files being found.
2. Added documentation for the above new CLI flag.
3. Added a Git ignore line for local environment files (e.g. `.envrc`).

Would resolve #1331

### DIFF
--- a/packages/test-runner-core/src/config/TestRunnerCoreConfig.ts
+++ b/packages/test-runner-core/src/config/TestRunnerCoreConfig.ts
@@ -27,6 +27,7 @@ export interface CoverageConfig {
 export interface TestRunnerCoreConfig {
   rootDir: string;
   files?: string | string[];
+  allowPassWithoutTests?: boolean;
   concurrentBrowsers: number;
   concurrency: number;
 
--- a/packages/test-runner/src/config/parseConfig.ts
+++ b/packages/test-runner/src/config/parseConfig.ts
@@ -188,8 +188,10 @@ export async function parseConfig(
       // we can improve this by relying only on groups inside the test runner itself
       if (groupConfig.files == null) {
         groupConfig.files = finalConfig.files;
+        groupConfig.allowPassWithoutTests = finalConfig.allowPassWithoutTests;
       }
       finalConfig.files = undefined;
+      finalConfig.allowPassWithoutTests = undefined;
 
       groupConfigs = [groupConfig];
     }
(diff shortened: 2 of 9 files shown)


Westbrook:
Could you add a change set as [requested above](https://github.com/modernweb-dev/web/pull/2758#issuecomment-2183220312)?

glektarssza:
I will add one after work today.

glektarssza:
Changeset added, @Westbrook. I think the bot's comment is completely out to lunch now, though, because of the merge commit in a957ff9.

Westbrook:
Thanks for getting the changeset it. That is looking fine. Can you check in the CI output? Once that’s green we should be good to edge this.

glektarssza:
Will take a look tonight!

packages/test-runner/src/config/parseConfig.ts
```diff
@@ -188,8 +188,10 @@ export async function parseConfig(
       // we can improve this by relying only on groups inside the test runner itself
       if (groupConfig.files == null) {
         groupConfig.files = finalConfig.files;
+        groupConfig.allowPassWithoutTests = finalConfig.allowPassWithoutTests;
       }
       finalConfig.files = undefined;
+      finalConfig.allowPassWithoutTests = undefined;
```
bashmish, line 194: what is the reason to set it as `undefined`? isn't it more clear to have it always as a boolean and default to `false`?

packages/test-runner-core/src/config/TestRunnerCoreConfig.ts
```diff
@@ -27,6 +27,7 @@ export interface CoverageConfig {
 export interface TestRunnerCoreConfig {
   rootDir: string;
   files?: string | string[];
+  allowPassWithoutTests?: boolean;
```
bashmish, line 30: related to another question: I'd just make it `allowPassWithoutTests: boolean` and default to "false", here and below

PR #2786, opened 2024-08-21
feat(storybook-utils): add MDXFileLoader to load and render non-JS files in MDX

bashmish:
## What I did

1. add MDXFileLoader to load and render non-JS files in MDX
2. fix exports of the utils packages

### DIFF
--- a/packages/storybook-utils/index.mjs
+++ b/packages/storybook-utils/index.mjs
@@ -1,1 +1,1 @@
-export * from './dist/index.js';
+export * from './src/index.js';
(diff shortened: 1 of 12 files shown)


packages/storybook-utils/index.mjs
```diff
@@ -1 +1 @@
-export * from './dist/index.js';
+export * from './src/index.js';
```
bashmish, line 1: not sure how this worked before, but we use this package and exports work, so I assume that TS config changed at some point and stopped copying JS files into `dist`, however this is not needed really and we can just directly import from the `src`

Issue #2793, opened 2024-08-29
Config loader: test fails since Node v22.7.0


web-padawan:
The following error is thrown in CI and can be also reproduced locally using Node 22.7.0.

```
$ npm run test:node -w @web/config-loader

> @web/config-loader@0.3.1 test:node
> mocha test/**/*.test.js --reporter dot

  11 passing (12ms)
  1 failing

  1) cjs package
       throws when loading module-in-.js:

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
```

Looks like it's caused by the fact that [module syntax detection is now enabled by default](https://nodejs.org/uk/blog/release/v22.7.0#module-syntax-detection-is-now-enabled-by-default) and Node is now smarter than we expect it to be, which makes the test fail as no exception is thrown as the file is parsed due to syntax detection:

> Syntax detection attempts to run ambiguous files as CommonJS, and if the module fails to parse as CommonJS due to ES module syntax, Node.js tries again and runs the file as an ES module. Ambiguous files are those with a .js or no extension, where the nearest parent package.json has no "type" field (either "type": "module" or "type": "commonjs").

I think for now we can workaround this by adding `--no-experimental-detect-module` flag. This would make the test pass.

PR #2794, opened 2024-08-29
chore: set type to commonjs in config-loader package.json

web-padawan:
## What I did

Fixes https://github.com/modernweb-dev/web/issues/2793

Set `type` to `commonjs` to avoid test failure due to Node 22.7 experimental module syntax detection.

### DIFF
--- a/packages/config-loader/package.json
+++ b/packages/config-loader/package.json
@@ -14,12 +14,13 @@
   "author": "modern-web",
   "homepage": "https://github.com/modernweb-dev/web/tree/master/packages/config-loader",
   "main": "src/index.js",
+  "type": "commonjs",
   "engines": {
     "node": ">=18.0.0"
   },
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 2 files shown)


packages/config-loader/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "test:node": "mocha test/**/*.test.js --reporter dot",
+    "test:node": "mocha test/**/*.test.js --reporter dot --no-experimental-detect-module",
```
Westbrook, line 22: Is this _roughly the same_ as adding a `"type": "commonjs"` entry?

Westbrook: I'm not 100% sure where this package fits in the ecosystem, but doing so would prevent any consumers from having a similar issue, right?

web-padawan: @Westbrook The package is used by `@web/dev-server` and `@web/test-runner` to load respective configs. 

Updated to use `"type": "commonjs"` and added a changeset. I think it's fine to use patch release for this.

PR #2813, opened 2024-09-23
Bump chokidar v4

talentlessguy:
## What I did

1. Bumped chokidar to v4.0.1 which removes 11 dependencies and has better TypeScript support

local test pass, and it builds locally as well

### DIFF
--- a/.changeset/giant-games-count.md
+++ b/.changeset/giant-games-count.md
@@ -1,0 +1,6 @@
+---
+'@web/test-runner-core': patch
+'@web/dev-server-core': patch
+---
+
+bump chokidar to v4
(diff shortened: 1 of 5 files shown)


bashmish:
> removes 11 dependencies

from the lock file it seems that older chokidar is still used by some packages, which might mean we get more deps as a result on an average project that depends on multiple `@web/` packages
can you please double check that?
it's a pity if we solve this only partially and make deps more of an issue then

talentlessguy:
@bashmish an old version of unplugin is used that depends on v3.6 chokidar. the latest. v1.12.3 of unplugin doesn't depend on chokidar anymore.

PR #2816, opened 2024-09-26
Storybook 8

talentlessguy:
## What I did

1.


packages/dev-server-storybook/package.json
```diff
@@ -59,7 +59,7 @@
     "@rollup/plugin-babel": "^6.0.4",
     "@rollup/plugin-node-resolve": "^15.0.1",
     "@rollup/plugin-terser": "^0.4.4",
-    "@storybook/csf-tools": "^6.4.9",
```
bashmish, line 62: dev-server-storybook is an older pre-builder era package, we don't maintain it actively anymore and no need to update it, actually it should remain on 6.x

packages/storybook-builder/package.json
```diff
@@ -49,10 +49,11 @@
   "dependencies": {
     "@rollup/plugin-node-resolve": "^15.1.0",
     "@rollup/pluginutils": "^5.0.2",
-    "@storybook/core-common": "^7.0.0",
+    "@storybook/channels": "^8.3.3",
```
bashmish, line 52: for consistency across all Storybook packages it's important to have same version
as a library we didn't force any specific one and tried to be compatible with every
updating all storybook packages and keeping them on the same version is hard enough task, so the library shouldn't make it even harder
this is different here
any specific reason?

bashmish:
there are pipeline errors, so those are the minimum that such migration need to cover

in general the Storybook pipeline is pretty reliable in testing even such big changes like this PR, but a few things I always needed to check manually, specifically everything related to path resolution, because paths work very differently when you test within a subpackage which is part of a bigger monorepo like modern-web and inside a project, also differently when it's an actual published NPM package and when it's a symlink, so I'm afraid I'll need to conduct more tests manually after the pipeline is green

my team has been planning to work on this in this autumn, can't give specific timelines, but sooner than later
so we gonna have a closer look into your PR

packages/dev-server-storybook/package.json
talentlessguy: noted

talentlessguy:
Thank you for the review, i didnt have historical context of the monorepo, so i will revert some of the changes for unmaintained tools

packages/storybook-builder/package.json
talentlessguy: not sure what do you mean

I had to add `@storybook/core` because channel transports have been moved there from `@storybooks/channels`

JReinhold: I think what he means is that here you're dependending on `^8.3.3` while below you're depending on `^8.0.0`, and that they should be the same across the board

package.json
```diff
@@ -46,7 +46,8 @@
     "@rocket/cli": "^0.10.1",
     "@rocket/core": "^0.1.2",
     "@rocket/launch": "^0.6.0",
-    "@rocket/search": "^0.6.0"
+    "@rocket/search": "^0.6.0",
+    "@storybook/core": "^8.3.3"
```
JReinhold, line 50: This shouldn't be needed, you should be able to depend on `storybook` instead, I hope

packages/storybook-builder/package.json
```diff
@@ -70,9 +71,10 @@
     "remark-slug": "^6.0.0",
     "rollup": "^4.4.1",
     "rollup-plugin-external-globals": "^0.9.0",
-    "slash": "^5.1.0"
+    "slash": "^5.1.0",
+    "storybook": "^8.3.3"
```
JReinhold, line 75: I don't know much about this project, but in 99 % of the cases, `storybook` should be a peer dependency here, because projects should always have that installed explicitly.

Maybe that's not the case for your setup here though.

talentlessguy: I think I was trying to obtain channels and forgot to uninstall storybook, which depends on `@storybook/core`. I'll remove it

package.json
talentlessguy: `storybook` itself includes a lot of unrelated stuff. I only need `@storybook/core/channels`

packages/storybook-builder/package.json
talentlessguy: ah okay, I will upgrade it to 8.3 everywhere, if I haven't already

package.json
bashmish: shouldn't this be in one of the packages, rather that the monorepo root?

packages/storybook-builder/src/generate-setup-addons-script.ts
```diff
@@ -2,19 +2,11 @@
 
 export async function generateSetupAddonsScript() {
   return `
-import { createChannel as createPostMessageChannel } from '@storybook/channel-postmessage';
-import { createChannel as createWebSocketChannel } from '@storybook/channel-websocket';
```
bashmish, line 6: did you remove these 2 deps? I assumed they moved stuff to core or import from it in their own builders, is that so? if that, I think we don't need to depend on these 2 packages directly anymore

bashmish: > did you remove these 2 deps? I assumed they moved stuff to core or reexport and then import from it in their own builders, is that so? if that, I think we don't need to depend on these 2 packages directly anymore

talentlessguy: one if those channels is not longer maintained, another one is a part of the core now

PR #2818, opened 2024-10-01
first step to get electron browser running, but tests not working yet

trusktr:
TODO:

- [x] enable launching `playwrite._electron` experimental Electron launcher.
- [ ] the window pops open, but it appears the tests are hung, maybe the context/page handling needs to be adjusted, the API is slightly different than other browsers  and there seems to be only a single context and page, not multiple like other browsers, or maybe I missed it.

### DIFF
--- a/packages/test-runner-playwright/src/PlaywrightLauncher.ts
+++ b/packages/test-runner-playwright/src/PlaywrightLauncher.ts
@@ -1,4 +1,10 @@
-import playwright, { Browser, Page, LaunchOptions, BrowserContext } from 'playwright';
+import playwright, {
+  Browser as RegularBrowser,
+  Page,
+  LaunchOptions,
+  BrowserContext,
+  ElectronApplication,
+} from 'playwright';
 import { BrowserLauncher, TestRunnerCoreConfig, CoverageMapData } from '@web/test-runner-core';
 import { PlaywrightLauncherPage } from './PlaywrightLauncherPage.js';
 
@@ -6,7 +12,9 @@ function capitalize(str: string) {
   return `${str[0].toUpperCase()}${str.substring(1)}`;
 }
 
-export type ProductType = 'chromium' | 'firefox' | 'webkit';
+export type ProductType = 'chromium' | 'firefox' | 'webkit' | '_electron';
+
+export type Browser = RegularBrowser | ElectronApplication;
 
 interface CreateArgs {
   browser: Browser;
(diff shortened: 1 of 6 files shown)


packages/test-runner-playwright/src/PlaywrightLauncher.ts
```diff
@@ -57,12 +66,16 @@ export class PlaywrightLauncher implements BrowserLauncher {
   }
 
   async stop() {
-    if (this.browser?.isConnected()) {
-      await this.browser.close();
+    if (this.browser) {
+      if (!('isConnected' in this.browser) || this.browser.isConnected()) {
+        await this.browser.close();
+      }
```
trusktr, line 72: The `ElectronApplication` instance does not have a `isConnected` method.

packages/test-runner-playwright/src/index.ts
```diff
@@ -26,8 +26,20 @@ export function playwrightLauncher(args: PlaywrightLauncherArgs = {}) {
   const {
     product = 'chromium',
     launchOptions = {},
-    createBrowserContext = ({ browser }) => browser.newContext(),
-    createPage = ({ context }) => context.newPage(),
+    createBrowserContext = ({ browser }) => {
+      if ('browserWindow' in browser) {
+        return browser.context();
```
trusktr, line 31: There is no `newContext()` method on the `ElectronApplication` instance. There seems to be a single context for the whole `ElectronApplication` only, so maybe this is why the tests are not working?

packages/test-runner-playwright/src/index.ts
```diff
@@ -26,8 +26,20 @@ export function playwrightLauncher(args: PlaywrightLauncherArgs = {}) {
...
+      } else {
+        return browser.newContext();
+      }
+    },
+    createPage = ({ context, browser }) => {
+      if ('browserWindow' in browser) {
+        return browser.firstWindow();
```
trusktr, line 38: The `firstWindows` is a playwright `Page`. Maybe for Electron we need to open and close windows, instead of making contexts?

packages/test-runner-playwright/test/playwrightLauncher.test.ts
```diff
@@ -1,85 +1,104 @@
...
+//   runIntegrationTests(createConfig, {
+//     basic: true,
+//     many: true,
+//     focus: true,
+//     groups: true,
+//     parallel: true,
+//     testFailure: true,
+//     locationChanged: true,
+//   });
+// });
```
trusktr, line 21: - [ ] TODO re-enable other tests before mering

PR #2821, opened 2024-10-09
`dev-server-core/WebSocketsManager`: send periodic ping over WebSockets to prevent connection closure for inactivity

larabr:
## What I did

1. Added pinging mechanism over the web sockets to prevent unexpected connection closure after ~60s (with websocket error `1006`) when testing on BrowserStack (Safari).

I couldn't get to the bottom of what exactly is causing the timeout, but sending periodic ping messages fixes it. I checked if the `ws` server is closing the connection, but it doesn't seem to be the case; using a ping to keep the connection alive is suggested [here](https://github.com/websockets/ws/issues/1436).

### DIFF
--- a/packages/dev-server-core/src/web-sockets/WebSocketsManager.ts
+++ b/packages/dev-server-core/src/web-sockets/WebSocketsManager.ts
@@ -29,7 +29,10 @@ export class WebSocketsManager extends EventEmitter<Events> {
     });
     this.webSocketServer.on('connection', webSocket => {
       this.openSockets.add(webSocket);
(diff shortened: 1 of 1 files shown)


bashmish:
Tricky one, I would love to have a stable solution for BrowserStack, many times my tests just randomly failed with timeouts in Firefox and Safari on BrowserStack.

However do you think it's a generic issue with WS or is it due to BrowserStack? If the latter, it would be better to implement this in the BrowserStack adapter

larabr:
In this case I think it might be due to the Browserstack testing setup, because I couldn't repro with Playwright in local, but I'm not sure if it's a problem unique to the platform: based on other issues I've found around (incl. the one I linked), other network setups might be affected as well.

Issue #2825, opened 2024-10-13
Code coverage errors when using lodash templates


miyasudokoro:
I am maintaining some legacy code that uses lodash templating. When code coverage gets calculated, the following error is thrown repeatedly:

```
TypeError: Invalid URL
    at new URL (node:internal/url:775:36)
    at v8ToIstanbul (/var/jenkins_home/workspace/(redacted)/node_modules/@web/test-runner-coverage-v8/dist/index.js:52:21)
    ....
```

Debugging this shows that the value of entry.url is the string "lodash.templateSources[0]" for these errors.

I believe the best resolution of this is to skip all invalid URLs after checking them via `URL.canParse`.

PR #2827, opened 2024-10-15
fix(test-runner-coverage-v8): #2825 ignore invalid URLs 

miyasudokoro:
## What I did

1. In test-runner-coverage-v8 index.ts v8ToIstanbul, it now ignores V8Coverage entry.url that are not valid URLs. I did not use URL.canParse because @web/root has engines node: >=18.0.0.
2. I moved the try-catch to surround the entire if-statement as an extra precaution for future issues.


miyasudokoro:
Just giving this a bump because I really do need it to be adopted.

PR #2829, opened 2024-10-16
feat(test-runner-browserstack): Make the local proxy configurable

fbuecklers:
## What I did

1. I introduced a top level local options, which allow you to disable the local proxy setup. This is specially useful when you just want to test public reachable domains
2. I have made the internal IP lookup conditional based on the capability.local flag, to make sure that it will only be run in cases it is really needed. This is also useful in CI environments where the ip command may not be available. Event an import of the module has previously crashed such CI runs.

### DIFF
--- a/packages/test-runner-browserstack/src/browserstackLauncher.ts
+++ b/packages/test-runner-browserstack/src/browserstackLauncher.ts
@@ -11,15 +11,14 @@ import {
 export interface BrowserstackLauncherArgs {
   capabilities: Record<string, unknown>;
   localOptions?: Partial<browserstack.Options>;
+  local?: boolean
 }
 
 const REQUIRED_CAPABILITIES = ['name', 'browserstack.user', 'browserstack.key', 'project', 'build'];
-const localIp = internalIp.v4.sync() as string;
-if (!localIp) {
-  throw new Error('Can not determine the local IP.');
-}
 
 export class BrowserstackLauncher extends WebdriverLauncher {
+  private localIp?: string;
+
   constructor(
     private capabilities: Record<string, unknown>,
     public name: string,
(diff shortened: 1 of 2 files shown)


Westbrook (commented):
Please be sure to update the documentation that distributes [here](https://modern-web.dev/docs/test-runner/browser-launchers/browserstack/#configuration) to support any changes that may be approved here.

packages/test-runner-browserstack/src/browserstackLauncher.ts
```diff
@@ -34,19 +33,35 @@ export class BrowserstackLauncher extends WebdriverLauncher {
...
+        this.capabilities['browserstack.key'] as string,
+        this.localOptions,
+      );
+    }
     await super.initialize(config);
   }
 
   startSession(sessionId: string, url: string) {
-    return super.startSession(sessionId, url.replace(/(localhost|127\.0\.0\.1)/, localIp));
+    if (url === 'localhost' || url === '127.0.0.1') {
```
Westbrook, line 57: Previously this replace these when available, which means `url` likely wasn't exactly `localhost` or `127.0.0.1`, but included them? Be sure that this isn't breaking any functionality on the other side of your change.

fbuecklers:
Hi @Westbrook Thanks for the quick review. I have hopefully addressed your points

packages/test-runner-browserstack/src/browserstackLauncher.ts
fbuecklers: Good point it is fixed

docs/docs/test-runner/browser-launchers/browserstack.md
```diff
@@ -81,3 +91,5 @@ The Browserstack launcher takes two properties, `capabilities` and `localOptions
 `capabilities` are the selenium capabilities used to configure the browser to launch in Browserstack. You can generate most of these on the Saucelabs. It must contain a `browserstack.user` and `browserstack.key` property to authenticate with Browserstack, as well as `name`, `build` and `project` to identify the test run.
```
Westbrook, line 91: See the line above the lists properties.

fbuecklers: Fixed

fbuecklers:
@Westbrook Does it look good for you now?

Issue #1880, opened 2022-02-20
[test runner] coverageConfig includes does not include untested files


jmcgavin:
The `includes` property of the [`coverageConfig` ](https://modern-web.dev/docs/test-runner/cli-and-configuration/#configuration-file) object does not include untested files in the code coverage. Only files visited during tests are included in the coverage.

I think having the option to include the entire codebase in the code coverage output is a good idea. Like the `--all` flag that can be used with [istanbul/nyc](https://github.com/istanbuljs/nyc#selecting-files-for-coverage).

jmcgavin:
The [@web/test-runner-coverage-v8](https://github.com/modernweb-dev/web/blob/master/packages/test-runner-coverage-v8/src/index.ts#L3) uses [the v8-to-istanbul lib which initializes `all` to false](https://github.com/istanbuljs/v8-to-istanbul/blob/master/lib/v8-to-istanbul.js#L43).

jmcgavin:
#1873

u-b-i-p-i-x-e-l-s:
@LarsDenBakker would love to have your feedback on this. 😃

LarsDenBakker:
I'm not sure the "all" flag would change anything here. Did anyone test this? I think we need to see all the files in the browser to make them show up in the coverage.

electrovir:
This may be overkill but here's a workaround:

1. before running web-test-runner, run a new script that finds all your files
2. using that script, generate a new TS/JS file that imports all the found file paths
3. include that new file in the test files that web-test-runner executes
4. now all your files included in the coverage report because the browser imported them all!

mohamedmansour:
> This may be overkill but here's a workaround:
> 
> 
> 
> 1. before running web-test-runner, run a new script that finds all your files
> 
> 2. using that script, generate a new TS/JS file that imports all the found file paths from step 1
> 
> 3. include that new file in the test files that web-test-runner executes
> 
> 4. now all your files are included in the coverage report because the browser imported them all!

Another approach instead of creating a new test file is to add it to Istanbul reporter as 0 coverage for all files found then merge the coverage data found.

PR #2833, opened 2024-10-25
#1880 coverage configuration 'all' option adds coverage for untested files

miyasudokoro:
I implemented #1880, which is to enable the configuration "all" to provide coverage measurement and reporting for untested/untouched files. 

1. The core code change is in test-runner-coverage-v8 index.ts, based on the [c8 implementation for "all"](https://github.com/bcoe/c8/blob/ff146b4dde004c62651b57c33cedd8353c94c423/lib/report.js#L396). It generates empty reports for all included-not-excluded non-test files in the base coverage directory. 
3. To improve performance for large projects, I skip generating empty reports for files present in the local cache of test-runner-coverage-v8 index.ts, because if they are in the local cache, they have already had some coverage measured. 
4. The generated empty reports are then passed through the same v8-to-istanbul code as normal reports to ensure uniform reporting. To more easily reuse the code and improve readability, I split the existing v8ToIstanbul function into multiple smaller functions.

### DIFF
--- a/docs/docs/test-runner/cli-and-configuration.md
+++ b/docs/docs/test-runner/cli-and-configuration.md
@@ -104,6 +104,8 @@ interface CoverageConfig {
   report: boolean;
   reportDir: string;
   reporters?: ReportType[];
+  // whether to measure coverage of untested files
+  all?: boolean;
 }
 
 type MimeTypeMappings = Record<string, string>;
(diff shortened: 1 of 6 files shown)


kaizencode-art:
Requesting core team review. This PR addresses the issue described in [Issue 1880](https://github.com/modernweb-dev/web/issues/1880). Implementation appears correct and resolves the reported problem.

PR #2851, opened 2024-12-23
chore: upgrade selenium-standalone version to v9

web-padawan:
## What I did

1. Upgraded `selenium-standalone` version.


web-padawan:
The `selenium-standalone` v9.1.0 and later includes transitive dependency on `glob` v11 which doesn't support Node 18.

PR #2852, opened 2024-12-24
feat: upgrade WebdriverIO to v9, drop JWP capabilities

web-padawan:
## What I did

1. Upgraded `webdriverio` to v9 and changed the saucelabs launcher accordingly
2. Removed logic and docs related to JWP capabilities as these have been dropped

## Note

Tested locally by running tests for `@web/test-runner-webdriver` package. These don't currently run in CI as they were skipped long ago, in particular due to https://github.com/webdriverio/selenium-standalone/issues/788 - that issue is now fixed but unfortunately the fix landed in a version that dropped Node 18 support.

### DIFF
--- a/.changeset/grumpy-meals-kick.md
+++ b/.changeset/grumpy-meals-kick.md
@@ -1,0 +1,6 @@
+---
+'@web/test-runner-saucelabs': minor
+'@web/test-runner-webdriver': minor
+---
+
+Upgrade WebdriverIO to v9, drop JWP capabilities
--- a/.changeset/grumpy-meals-kick.md
+++ b/.changeset/grumpy-meals-kick.md
@@ -1,4 +1,5 @@
 ---
+'@web/test-runner-browserstack': minor
 '@web/test-runner-saucelabs': minor
 '@web/test-runner-webdriver': minor
 ---
--- a/packages/test-runner-webdriver/src/utils.ts
+++ b/packages/test-runner-webdriver/src/utils.ts
@@ -1,17 +1,15 @@
-import WebDriver from 'webdriver';
-
-function getPlatform(c: WebDriver.DesiredCapabilities): string | undefined {
-  return c.platformName || c.platform;
+function getPlatform(c: WebdriverIO.Capabilities): string | undefined {
+  return c.platformName;
 }
 
-export function getBrowserName(c: WebDriver.DesiredCapabilities): string | undefined {
+export function getBrowserName(c: WebdriverIO.Capabilities): string | undefined {
   return c.browserName;
 }
 
-function getBrowserVersion(c: WebDriver.DesiredCapabilities): string | undefined {
-  return c.browserVersion || c.version;
+function getBrowserVersion(c: WebdriverIO.Capabilities): string | undefined {
+  return c.browserVersion;
 }
 
-export function getBrowserLabel(c: WebDriver.DesiredCapabilities): string {
+export function getBrowserLabel(c: WebdriverIO.Capabilities): string {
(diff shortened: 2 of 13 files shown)


.changeset/grumpy-meals-kick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@web/test-runner-saucelabs': minor
+'@web/test-runner-webdriver': minor
+---
+
+Upgrade WebdriverIO to v9, drop JWP capabilities
```
Westbrook, line 6: "Drop"ing something feels like it should be more than a `minor` release, but I don't know Webdriver that well...

packages/test-runner-webdriver/src/utils.ts
```diff
@@ -1,17 +1,15 @@
-import WebDriver from 'webdriver';
-
-function getPlatform(c: WebDriver.DesiredCapabilities): string | undefined {
-  return c.platformName || c.platform;
+function getPlatform(c: WebdriverIO.Capabilities): string | undefined {
```
Westbrook, line 1: Where does this file get this type value from? If it works, it works, but confused that there is no import. The main question is if we're doing something special somewhere else, or relying on something that I can't see, then is it visible to a more experienced WebdriverIO user in a way that would prevent their consumption of this from causing them issues?

Westbrook (commented):
Seems fine to me. Two questions and then likely good to go.

.changeset/grumpy-meals-kick.md
web-padawan: Sure, but we are still in 0.x so technically every minor can contain breaking changes. 
I wasn't sure whether I should use `major` for this in the changeset.

packages/test-runner-webdriver/src/utils.ts
web-padawan: I tried to use this but got error "Cannot use namespace 'Capabilities' as a type".

```
import type { Capabilities } from '@wdio/types';
```

The namespace is defined [here](https://github.com/webdriverio/webdriverio/blob/133dbeb651fab96391bd0af63dbca4d3d17f7aa0/packages/wdio-types/src/Capabilities.ts#L50-L51), seems like it's done for supporting [custom capabilities](https://webdriver.io/docs/capabilities/#extend-custom-capabilities).

.changeset/grumpy-meals-kick.md
bashmish: minor in 0.x is practically major, so is good

.changeset/grumpy-meals-kick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@web/test-runner-saucelabs': minor
+'@web/test-runner-webdriver': minor
```
bashmish, line 3: shouldn't other packages be updated too?
from the top of my head it's the browserstack one
https://github.com/modernweb-dev/web/blob/master/packages/test-runner-browserstack/package.json
maybe some others, good to check where else the `@web/test-runner-webdriver` is in deps

web-padawan: Good catch, added `@web/test-runner-browserstack`. This is the only remaining package that needs updating (other packages only use webdriver launcher as dev dependency).

.changeset/grumpy-meals-kick.md
Westbrook: Agreed. Thanks for clarifying.

packages/test-runner-webdriver/src/utils.ts
Westbrook: Does it run fine in other projects this way. I'm fine with whatever works, it was just surprising to see and would be annoying if this caused some undocumented requirements on consumers.

PR #2868, opened 2025-01-27
chore: npm update (except chokidar)

bashmish:
## What I did

1. `npm update`
2. then reverted `chokidar@4.0.3` to `chokidar@4.0.1` due to wrong types


bashmish:
Was too much to do in one go. Closing.

PR #2870, opened 2025-01-27
feat: migrate builder to Storybook 8

bashmish:
## What I did

1. Migrated `@web/storybook-builder` and `@web/storybook-framework-web-components` to Storybook 8
2. Made `@web/mocks` and `@web/storybook-utils` compatible with both Storybook 7 and 8 via peerDependencies
3. Updated docs accordingly, including some migration notes for Storybook 8


bashmish:
Windows build is still unstable and the instability is not related to this PR, but it did work multiple times, and since then I only update the docs, so gonna proceed with the merge now.

PR #2874, opened 2025-01-30
fix(test-runner-saucelabs): use correct SauceLabsCapabilities type

sissbruecker:
## What I did

1. Fixed type for providing a `SauceLabsCapabilities` object to `createSauceLabsLauncher`

## Notes

The parameter is specifically for providing SauceLabs capabilities and is destructured into `sauce:options` in the final capabilities object:
https://github.com/modernweb-dev/web/blob/12d0de49e05f38db32745448ec06f5033b07b431/packages/test-runner-saucelabs/src/createSauceLabsLauncher.ts#L48-L51

Thus the type should also reflect that the parameter should contain Sauce Labs capabilities and not all Web Driver capabilities. You can see this being an issue in `packages/test-runner-saucelabs/test-remote/saucelabsLauncher.test.ts` for example, which doesn't compile ATM.

### DIFF
--- a/.changeset/chatty-rings-tease.md
+++ b/.changeset/chatty-rings-tease.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-saucelabs': patch
+---
+
+Use correct type for providing SauceLabsCapabilities object in createSauceLabsLauncher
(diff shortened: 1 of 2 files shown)


bashmish:
thanks!

PR #2876, opened 2025-01-31
feat(test-runner-saucelabs): add support for Sauce Connect 5

sissbruecker:
## What I did

1. Bump `saucelabs` to `9.0.0` to make use of Sauce Connect 5
2. Adapt to breaking changes in newer version

## Notes

For some background, `test-runner-saucelabs` currently uses `saucelabs@^7.2.0`, which downloads Sauce Connect version 4. That version has already reached [EOL](https://docs.saucelabs.com/secure-connections/sauce-connect/lifecycle/) and starting May 5, 2025 SauceLabs will [stop accepting traffic](https://docs.saucelabs.com/secure-connections/sauce-connect/) from that version. That means that at that point `test-runner-saucelabs` would stop working as well.

SauceLabs users need to upgrade to Sauce Connect 5 instead. To support that SauceLabs have now released `saucelabs@9.0.0`, which downloads and runs the newer version. That version contains **breaking changes** in terms of API and configuration options. That means that this version bump is a breaking change for `test-runner-saucelabs` users as well, as they might use options that have been changed.

Bumping the version now would give `test-runner-saucelabs` users ample time to upgrade and adapt their projects before this package stops working completely beginning of May. SauceLabs has published a [migration guide](https://docs.saucelabs.com/secure-connections/sauce-connect-5/migrating/) that helps with upgrading projects.

With the changes in this PR I was able to connect to SauceLabs and run the tests in `test-runner-saucelabs` as well as in https://github.com/vaadin/web-components. I have only marked this a a minor change in the changeset, as the package has not had a major release yet.

### DIFF
--- a/packages/test-runner-saucelabs/src/SauceLabsLauncherManager.ts
+++ b/packages/test-runner-saucelabs/src/SauceLabsLauncherManager.ts
@@ -53,7 +53,7 @@ export class SauceLabsLauncherManager {
     this.connectionPromise = withTimeout(
       this.api.startSauceConnect({
         ...this.connectOptions,
-        noSslBumpDomains: `127.0.0.1,localhost,${internalIp.v4.sync()}`,
+        tlsPassthroughDomains: `^(127\\.0\\.0\\.1|localhost|${internalIp.v4.sync()?.replace(/\./g, '\\.')})$`,
       }),
       '[Saucelabs] Timed out setting up Sauce Connect proxy after 5 minutes.',
     );
--- a/packages/test-runner-saucelabs/src/SauceLabsLauncherManager.ts
+++ b/packages/test-runner-saucelabs/src/SauceLabsLauncherManager.ts
@@ -53,7 +53,9 @@ export class SauceLabsLauncherManager {
     this.connectionPromise = withTimeout(
       this.api.startSauceConnect({
         ...this.connectOptions,
-        tlsPassthroughDomains: `^(127\\.0\\.0\\.1|localhost|${internalIp.v4.sync()?.replace(/\./g, '\\.')})$`,
+        tlsPassthroughDomains: `^(127\\.0\\.0\\.1|localhost|${internalIp.v4
+          .sync()
+          ?.replace(/\./g, '\\.')})$`,
       }),
       '[Saucelabs] Timed out setting up Sauce Connect proxy after 5 minutes.',
     );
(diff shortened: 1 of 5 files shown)


packages/test-runner-saucelabs/src/createSauceLabsLauncher.ts
```diff
@@ -27,9 +34,14 @@ export function createSauceLabsLauncher(
...
 
-  const finalConnectOptions: SauceConnectOptions = { ...sauceConnectOptions };
-  if (typeof finalConnectOptions.tunnelIdentifier !== 'string') {
-    finalConnectOptions.tunnelIdentifier = `web-test-runner-${nanoid()}`;
+  const finalConnectOptions: SauceConnectOptions & LegacySauceConnectOptions = { ...sauceConnectOptions };
+  if (finalConnectOptions?.tunnelIdentifier) {
+    console.warn('The `tunnelIdentifier` option is deprecated. Use `tunnelName` instead.');
+    finalConnectOptions.tunnelName = finalConnectOptions.tunnelIdentifier;
+    delete finalConnectOptions.tunnelIdentifier;
+  }
```
sissbruecker, line 42: This option was an obvious change and it was easy to provide a migration path for it. Up to you if you want to include it.

packages/test-runner-saucelabs/src/SauceLabsLauncherManager.ts
```diff
@@ -53,7 +53,7 @@ export class SauceLabsLauncherManager {
     this.connectionPromise = withTimeout(
       this.api.startSauceConnect({
         ...this.connectOptions,
-        noSslBumpDomains: `127.0.0.1,localhost,${internalIp.v4.sync()}`,
+        tlsPassthroughDomains: `^(127\\.0\\.0\\.1|localhost|${internalIp.v4.sync()?.replace(/\./g, '\\.')})$`,
```
sissbruecker, line 56: All domains now must be passed a regex instead of a glob.

packages/test-runner-saucelabs/src/createSauceLabsLauncher.ts
bashmish: I'd rather not, such changes are expected in minor version in 0.x.x range, and I think we are just a proxy to SauceLabs libs, so we don't need to do more that they do, if they remove it, we should follow and not make own deprecation policy.

Should you fix this, I'll be happy to merge and release this update, good work!

sissbruecker: Alright, I removed it 🙂

bashmish (approved):

PR #2880, opened 2025-02-06
chore(test-runner-playwright): bump playwright

roeniss:
### Summary

@web/test-runner-playwright uses old version of Playwright, which causes problems on Ubuntu 24.04 internally. 

This issue shows the situation well: https://github.com/microsoft/playwright/issues/33051.

### Consideration

- Some other packages are depending on this @web/test-runner-playwright. I'm going to post another PR after this is merged.
- This is my first issue on the project. If I missed something, please let me know :D

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13508,9 +13508,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001607",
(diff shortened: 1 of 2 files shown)


roeniss:
I'm not familiar with the changeset. If I used it wrong, please let me know.

roeniss:
I realized that this patch have almost zero meaning. close.

PR #2904, opened 2025-03-07
feat(test-runner-junit-reporter): Add support for flat test files to junit-reporter

### LINKED ISSUE #2901: [test-runner-junit-reporter] JUnit reporter does not detect tests outside of "describe" suites
Currently, if you have a test file like this:
```javascript
it('under addition', function () {
  chai.expect(1 + 1).to.equal(2);
});

it('null hypothesis', function () {
  chai.expect(true).to.be.true;
});

it('asserts error', function () {
  chai.expect(false).to.be.true;
});
```

Then these test results will not be included in the XML file produced by the JUnit reporter package. The tests still run and can fail, so this format seems to be supported by the testing library without issue!

These tests will start being included in the report as soon as you wrap them in a `describe` block. Would it be possible to make it so the JUnit reporter can detect these tests outside of `describe` blocks as well?

JackRobards:
## What I did

Add support for test files that do not use a `describe` block (aka "flat") to the junit-reporter. 

A file like this would now be included in the report, where before it would not:
```typescript
it('under addition', function () {
  chai.expect(1 + 1).to.equal(2);
});

it('null hypothesis', function () {
  chai.expect(true).to.be.true;
});

it('asserts error', function () {
  chai.expect(false).to.be.true;
});
```

As for what to call the Test Suite - I followed the same pattern that the `toResultsWithMetadataByBrowserTestFileName` function was using. So each of these test suites will be named based on the browser + file being tested.

This should be a non-breaking change from my testing - it is only intended to detect some test cases that wouldn't be included previously!

closes #2901

### DIFF
--- a/.changeset/rotten-brooms-carry.md
+++ b/.changeset/rotten-brooms-carry.md
@@ -1,0 +1,5 @@
+---
+'@web/test-runner-junit-reporter': minor
+---
+
+chore: Add support for flat test files to junit reporter
--- a/packages/test-runner-junit-reporter/src/junitReporter.ts
+++ b/packages/test-runner-junit-reporter/src/junitReporter.ts
@@ -69,7 +69,7 @@ interface TestSuiteXMLAttributes {
 const assignSessionAndSuitePropertiesToTests = ({
   testResults,
   ...rest
-}: TestSession): TestResultWithMetadata[] => {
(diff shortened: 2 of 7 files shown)


JackRobards:
@bashmish Is this something you would have time to take a look at, or are there any changes I should make here? I totally get it if you're just busy with other things, but I wanted to check-in in case there's anything I can do here.

PR #2915, opened 2025-04-02
refactor(config-loader): do not lookup `package.json` with explicit extensions

devversion:
Currently when a config has an explicit extension, the config-loader still tries resolving/traversing for a `package.json` file. This can be slow, and also break in some environments with sandboxing; especially when used in combination with Bazel.

It doesn't seem investigating why the calls are stuck, if the logic should not run in general when an explicit extension is provided.

### DIFF
--- a/.changeset/nice-guests-invent.md
+++ b/.changeset/nice-guests-invent.md
@@ -1,0 +1,9 @@
+---
+'@web/config-loader': patch
+---
+
+Optimizes config loading to not look for `package.json` in file system if the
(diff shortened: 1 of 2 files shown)


bashmish:
@devversion  thanks for the fix! Please add a changeset following the previous comment notes, then I can release it too.

devversion:
@bashmish done!

bashmish:
Thanks! Will release shortly.

devversion:
Thanks for the super quick review & merge!

PR #2932, opened 2025-04-16
fix(summaryReporter): add back broken error reporting and make log reporting optional for summaryReporter

MichaelWest22:
## What I did

1. Added back support for reporting a summary of errors in summaryReporter
2. Added option to turn off error reporting in summaryReporter
3. Added option to turn off log reporting in summaryReporter

I found that in #2126 the option to add log reporting was enabled but it also replaced the logger with the reportTestFileResults BufferedLogger. However this logger flushes its log buffer as soon as reportTestFileResults stage finishes and then this logger is used for Error reporting in onTestRunFinished() and this reporting now just goes to a buffer that is never flushed.  I resolved this issue by moving the error reporting logging back to TestRunnerLogger. This change could be a breaking change to some users who have worked around the lack of error reporting so I have added a flag so it can be turned off if needed but have set error reporting to enabled by default to match the original older working behaviour.

Also it is often useful to combine multiple reporters and if other reporters have log reporting like the defaultReporter it can double up log reporting in some combinations so I added a flag to optionally turn this off if needed.

### DIFF
--- a/packages/test-runner/src/reporter/summaryReporter.ts
+++ b/packages/test-runner/src/reporter/summaryReporter.ts
@@ -29,7 +29,7 @@ const dim = color([2, 0]);
 
 /** Test reporter that summarizes all test for a given run */
 export function summaryReporter(opts: Options): Reporter {
-  const { flatten = false, reportTestLogs = true, reportTestErrors = false } = opts ?? {};
+  const { flatten = false, reportTestLogs = true, reportTestErrors = true } = opts ?? {};
   let args: ReporterArgs;
   let favoriteBrowser: string;
 
--- a/packages/test-runner/src/reporter/summaryReporter.ts
+++ b/packages/test-runner/src/reporter/summaryReporter.ts
@@ -9,10 +9,13 @@ import type {
 import { reportTestsErrors } from './reportTestsErrors.js';
 import { reportTestFileErrors } from './reportTestFileErrors.js';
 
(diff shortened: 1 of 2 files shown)


packages/test-runner/src/reporter/summaryReporter.ts
```diff
@@ -26,10 +29,12 @@ const dim = color([2, 0]);
 
 /** Test reporter that summarizes all test for a given run */
 export function summaryReporter(opts: Options): Reporter {
-  const { flatten = false } = opts ?? {};
+  const { flatten = false, reportTestLogs = true, reportTestErrors = false } = opts ?? {};
```
paladox, line 32: Shouldn't reportTestErrors be made to default to on?

MichaelWest22: Yeah it could be changed so that reportTestErrors is true by default.  I was worried that setting it to true would alter the logging of anyone who has already upgraded to use the broken summaryReporter and maybe they are now used to having no report errors at the bottom or have worked around the issue somehow.  But maybe this is not a big issue and we can just turn it on by default.

bashmish: Maybe good to reflect the change in the description here:

>This change could be a breaking change to some users who have worked around the lack of error reporting so I had to add a flag to enable it that defaults to false to retain existing broken behavior.

bashmish:
Released in `@web/test-runner@0.20.2`

PR #2963, opened 2025-06-13
fix(test-runner-visual-regression): Fix importing test-runner-visual-regression

### LINKED ISSUE #2804: `visualDiff` of `@web/test-runner-visual-regression` can't be imported
Since version 0.10.0 of `@web/test-runner-visual-regression` typescript complains when trying to import `visualDiff` like: `import { visualDiff } from '@web/test-runner-visual-regression';` (suggested by [docs](https://github.com/modernweb-dev/web/blob/master/packages/test-runner-visual-regression/README.md)).

We assume that the `index.d.ts` which newly contains `.mjs` extension (`export * from './browser/commands.mjs';`) causes the problem. If the extension is removed, it works again.

Dependency update PR on our side with failing typescript:
PR: https://github.com/sbb-design-systems/lyne-components/pull/3069
Logs: https://github.com/sbb-design-systems/lyne-components/actions/runs/10811479138/job/29990965173?pr=3069

paladox:
This fix is done by @melink14.

Fixes #2804


jeripeierSBB:
This problem still exists, also with newest releases. It is still preventing from updating to a newer version (at least without a patch). 
@bashmish Could you have a look at the PR? thank you :)

packages/test-runner-visual-regression/index.d.ts
```diff
@@ -1,2 +1,2 @@
 // this file is autogenerated with the generate-mjs-dts-entrypoints script
-export * from './browser/commands.mjs';
```
bashmish, line 2: we usually have it with a `.js` extension which resolves to `.ts` or `.d.ts` file
not sure why, but for consistency I'd prefer it the same way as everywhere
we have plan to refactor this in the nearest future, so I'd not look too much in to this

only in 2 places somehow we got `.mjs`: here and in `packages/test-runner-commands/index.d.ts`

can you please fix both with a `.js` fix?

Image

paladox: Fixed, thanks!

.changeset/fix-import.md
```diff
@@ -0,0 +1,5 @@
+ ---
+ '@web/test-runner-visual-regression': patch
+ ---
+
+@web/test-runner-visual-regression - fixes importing under typescript
```
bashmish: please also update the changeset

```
 ---
 '@web/test-runner-commands': patch
 '@web/test-runner-visual-regression': patch
 ---

fix importing under typescript
```

paladox: Fixed, thanks!

paladox:
Thanks!

PR #3043, opened 2026-03-16
chore: upgrade TypeScript to 5.9.3 and enable .ts imports

bennypowers:
## Summary

Upgrade to TS 5.9.3 with `module: nodenext`, `rewriteRelativeImportExtensions`, and `.js`->`.ts` imports. CJS interop wrappers added for ESM-incompatible packages.

Drop EOL Node versions: all packages now require `>=22.0.0`. Node 18 and 20 are past EOL. CI matrix tests on 22, 24, and latest.

This is a non-breaking internal change. Published output remains CJS.

## Stack
1. #3052 -- `import type` annotations
2. **#3043** -- TS 5.9, nodenext, `.ts` imports, drop EOL Node versions

## Test plan
- [ ] `npm run build` passes with zero TS errors
- [ ] Lint passes
- [ ] Node tests pass on Node 22, 24, and latest
- [ ] Browser tests pass


packages/dev-server-core/src/middleware/etagCacheMiddleware.ts
```diff
@@ -1,4 +1,4 @@
-import { Middleware } from 'koa';
+import type { Middleware } from 'koa';
```
web-padawan, line 1: note: there are still quite a few `import type` changes in this PR like this one. Consider extracting them into separate PR now when https://github.com/modernweb-dev/web/pull/3052 is merged.

packages/dev-server-core/package.json
```diff
@@ -58,30 +60,29 @@
...
+    "@types/ws": "^8.5.13",
     "@web/parse5-utils": "^2.1.0",
     "chokidar": "^4.0.1",
     "clone": "^2.1.2",
     "es-module-lexer": "^1.0.0",
     "get-stream": "^6.0.0",
     "is-stream": "^2.0.0",
     "isbinaryfile": "^5.0.0",
-    "koa": "^2.16.1",
+    "koa": "^3.1.1",
```
web-padawan, line 72: suggestion: extract `koa` upgrade into separate PR to keep this one about TS upgrade.

packages/dev-server-esbuild/package.json
```diff
@@ -53,7 +52,7 @@
   "dependencies": {
     "@mdn/browser-compat-data": "^4.0.0",
     "@web/dev-server-core": "^0.7.4",
-    "esbuild": "^0.27.0",
+    "esbuild": "^0.25.0",
```
web-padawan, line 55: note: this change should be reverted, let's not downgrade esbuild.

web-padawan (commented):
Please consider splitting this PR further if you'd like to proceed to make reviewing easier.

packages/dev-server-esbuild/package.json
bennypowers: as discussed in chat, i think ts 5.9.3 was premature. i'd rather go for node:test first, then when 22 reaches eol, update minimum node to 24, and only then tackle typescript

playwright is the joker. we might need to deprecate a11ySnapshot

bennypowers:
Closing: maintainers agreed to ship EOL deprecation and node:test migration before TS upgrade. Keeping branch for reference.

PR #3053, opened 2026-03-26
Update to Storybook 9

bashmish:
## What I did

1. Migrated all projects to Storybook 9
2. Had to introduce a new major for `@web/mocks`, because there was a choice of either just migration it to Storybook 9 or decoupling from Storybook., both required a major update, so I went into the 2nd approach
3. As a result of 2, new addon was added `@web/storybook-addon-mocks`

thanks to @GuiAmPm for his contribution in #3039


packages/mocks/MIGRATION.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+### Removed `mockRollupPlugin`
+
+This plugin was mainly used for bundling Storybook, so it's not very useful on it's own, therefore it was deleted from `@web/mocks`.
+
+### Update of `msw`
+
+The version of `msw` was bumped to latest.
+We recommend checking the `msw` changelog to ensure a smooth transition for your specific workflows.
```
thepassle, line 38: Maybe we can add a line explaining why the version of MSW is pinned still, e.g.: "The version of `msw` was bumped to the latest, but is still pinned to a specific version because we've experienced a lot of unintentional breaking changes in minor or patch versions of `msw`, which leads to instability."

packages/mocks/package.json
```diff
@@ -23,19 +23,6 @@
...
-    "./storybook/addon/decorator.js": {
-      "types": "./dist-types/storybook/addon/decorator.d.ts",
-      "default": "./storybook/addon/decorator.js"
-    },
-    "./storybook/addon/manager.js": "./storybook/addon/manager.js",
-    "./storybook/decorator.js": {
-      "types": "./dist-types/storybook/decorator.d.ts",
-      "default": "./storybook/decorator.js"
-    },
-    "./storybook/addon.js": "./storybook/addon.js",
```
thepassle, line 38: this cleans up nicely

packages/mocks/README.md
```diff
@@ -90,175 +90,6 @@ http.get('/api/foo', ({ request, cookies, params }) => {
...
-
-<details>
-<summary>
-  <b><code>mockRollupPlugin</code> configuration options</b>
-</summary>
-The rollup plugin also takes an optional <code>interceptor</code>, which can be useful to handle things like rewriting <code>api.</code> prefixes made in requests, for example:
-
-```js
-mockRollupPlugin({
-  interceptor: `
```
thepassle, line 144: how do we handle interceptor stuff now? e.g. for ING?

packages/mocks/rollup-plugin.js
```diff
@@ -1,35 +0,0 @@
...
-
-const require = createRequire(import.meta.url);
-
-export function mockRollupPlugin(
-  { interceptor } = {
-    interceptor: '',
-  },
-) {
-  return {
-    name: 'rollup-plugin-msw',
```
thepassle, line 15: so the outputting of the service worker (and loading the sw) is now handled by the storybook addon?

bashmish: correct, in practice it was only used by Storybook
also I found this rollup plugin hard to reuse, e.g. in our app it wasn't usable due to the `<base href="/different-base">`, so I think it's best to remove this plugin anyway.

packages/mocks/README.md
bashmish: my analyses indicated that it wasn't really used
people did import some interceptors, but they were not needed

in general both in and outside of Storybook there are easy ways to include arbitrary code to the index.html, I mean this is a convenience API, but there is a lot of other options how to achieve the same, so I think this API wasn't in the right place

packages/mocks/MIGRATION.md
bashmish: will do, thanks!

PR #3069, opened 2026-05-08
Support preserving assets input structure in the output

bashmish:
## What I did

1. Add option `preserveDynamicStructure` to `@web/rollup-plugin-import-meta-assets` to resolve dynamic paths differently by using first asset as a base and preserving the dynamic part in the URL
2. Improve handling of assets in CSS in `@web/rollup-plugin-html` so that same use-case as in `preserveDynamicStructure` is also supported (but it doesn't need any extra option since all paths are static and dynamic paths and dynamic runtime were not an issue)
3. Support transforming of assets found in CSS

so that `assetFileNames` can preserve input structure of assets and no extra overhead is needed (e.g. no dynamic runtime in JS)

Review pointers:

- rollup-plugin-import-meta-assets
  - tests for import-meta-assets were all refactored, so a bit hard to see what's new, you can see it in the diff if you check what "it" lines remained unchanged and what have new names, those will be new and also everything under `describe('preserveDynamicStructure', () => {` is new
  - the algorithm is based on the idea that file paths can be calculated in runtime relative to the first found asset (which is calculated by rollup's own internal algorithm), that made it easier to make it work for  `preserveDynamicStructure`
  - I could solve `preserveDynamicStructure` use-case without the option, but the idea was that we don't want the dynamic runtime with path maps, so the main reason to introduce this option is to prevent that runtime from being generated

- rollup-plugin-html
  - new tests are easy to find since tests structure remained the same, but most important ones to start with are under `describe('preserved output paths', () => {`
  - I noticed while refactoring that we don't transform assets in CSS which I consider a bug, and since I was changing the mechanism from sync to async (AST visitors were sync and I needed async), I saw opportunity to fix the transformations too since they also impact the hash, and since I made everything async, it wasn't hard for me to implement them correctly too with hashing in mind
  - I use xxhash64, not exactly what rollup uses since I just use a simple HEX representation of it, plus it's hardcoded to use 64bit algo to be future proof (as far as I understand the math behind it), I need to calculate the hash from both the content and output path, the latter is a weak point, because the output path as we know from `preserved output paths` is not correct at this point and the CSS content still might change, so in theory there is a possibility for bugs here, but in practice it needs to be a super weird `assetFileNames` configuration that can possibly result in such bug, I wasn't able to create a logical unit test for it (it must be some combination of using hash for CSS file names, but not assets, and then chanign output paths for some assets conditionally in `assetFileNames`, I can't possible imagine who would ever do that), so decided to stop chasing it


packages/rollup-plugin-html/src/output/emitAssets.ts
```diff
@@ -99,69 +112,90 @@ export async function emitAssets(
...
+              const ref = this.emitFile({
+                type: 'asset',
+                name: extractAssetsLegacyCss ? `assets/${basename}` : basename,
+                originalFileName: assetInCssAbsPath,
+                source: transformedContent,
+              });
+              assetInCss.ref = ref;
+              assetInCss.outputPath = this.getFileName(ref);
+              if (!extractAssetsLegacyCss) {
+                assetInCss.hash = create64()
```
thepassle: can't we just use node:crypto? it would achieve the same goal right? its not entirely clear to me why we need a dep to create the hash

bashmish: @thepassle xxhash is super fast, it's used by rollup internally too (although they made their own bindings for Rust functions) and recommended for non-crypto hashes in such use-cases

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
```diff
@@ -66,13 +66,50 @@ function getImportMetaUrlType(node) {
  * @param {string|string[]} [options.exclude] A picomatch pattern, or array of patterns, which specifies the files in the build the plugin should _ignore_. By default no files are ignored.
  * @param {boolean} [options.warnOnError] By default, the plugin quits the build process when it encounters an error. If you set this option to true, it will throw a warning instead and leave the code untouched.
  * @param {function} [options.transform] A function to transform assets.
+ * @param {boolean} [options.preserveDynamicStructure] When enabled, emits dynamic assets and rewrites the URL pattern to resolve the original dynamic path relative to the first emitted asset. Requires that the output preserves both filenames (no hashing) and directory structure from the dynamic expression onwards.
```
bashmish, line 69: @thepassle do you think `preserveDynamicStructure` is a suitable name given the behavior and purpose?

packages/rollup-plugin-html/src/output/emitAssets.ts
thepassle: is it just perf related then? Does node:crypto impact perf that much? surely theres overhead loading and instantiating the wasm too right?

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
thepassle: I was gonna leave a comment about the name, but I also can't come up with anything different myself. I am a bit worried it makes sense to me because we've talked about it, but it might not be super clear to people finding out about it for the first time. But then again, the documentation and the requirements are clearly listed/explained, so maybe it just is what it is

packages/rollup-plugin-html/src/output/emitAssets.ts
bashmish: wasm is instantiated once and then the instance is reused
this depends on the amount of assets and their sizes
not sure what will be the best way to benchmark it
I can run some tests on our internal app, but it's not gonna represent a typical project I think

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
bashmish: `preserveInputStructure` or `expectPreservedInputStructure`?
2nd one is so far the closest to the intention

thepassle (approved):

packages/rollup-plugin-html/src/output/emitAssets.ts
bashmish: I tested on one internal project with many images on my Mac M1 Pro
xxhash64 was same as node:crypto sha256
we are talking milliseconds here, e.g. in this project around 13ms
if we take into account the wasm init, xxhash64 becomes a bit slower, although init is also within a millisecond

we use this hashing for references in CSS files only, I don't think there is gonna be a lot in real life for most projects, so wasm init might even make it relatively slower, but anyhow still talking milliseconds

I think the maintenance of a 3rd party dependency is a bigger issue, so prefer to switch to sha256 (and updated code in the PR with that and removed xxahsh dependency)
of course there are other algos in node:crypto, but sha256 seems well established and common, and 256 bits is maybe a bit too much for the task, but not sure I want to use sha1, so sha256 it is

packages/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js
bashmish: on a 2nd (3rd, 4th, 5th 🤣  ) thought, I'd keep current name

PR #3074, opened 2026-05-25
feat!: drop support for Node.js 18 and 20

bennypowers:
## Summary

- Minimum supported Node.js version is now **22.0.0**
- Node.js 18 EOL'd April 2025, Node.js 20 EOL'd April 2026
- CI matrix tests on Node 22 and 24
- Bump GitHub Actions to latest (checkout v6, setup-node v6, upload-artifact v7)

## Node 26 / latest

Pinned to 24 instead of `latest` due to [Playwright/yauzl hang on Node 26](https://github.com/microsoft/playwright/issues/40724). Will re-add `latest` once Playwright ships the fix.

## ts-node workaround

Added `NODE_OPTIONS='--no-experimental-strip-types'` to CI test steps. Node 22.18+ enables native type stripping, which preempts ts-node and breaks mocha tests. This workaround will be removed when mocha is replaced with `node:test` in a follow-up PR.

## Test plan

- [ ] CI green on Node 22 and 24 (Linux + Windows)
- [ ] Lint passes
- [ ] Browser tests pass
- [ ] Storybook builder verification passes

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,14 +12,14 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - name: Setup Node 20
-        uses: actions/setup-node@v4
+      - name: Setup Node 22
+        uses: actions/setup-node@v6
         env:
           FORCE_COLOR: 0
         with:
-          node-version: 20
+          node-version: 22
           cache: npm
 
       - name: Install Dependencies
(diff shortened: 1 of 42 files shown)


.github/workflows/lint.yml
```diff
@@ -12,14 +12,14 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - name: Setup Node 20
-        uses: actions/setup-node@v4
+      - name: Setup Node 22
```
web-padawan, line 17: nit: could also consider splitting GitHub Actions updates and switching to Node 22 for lint / verify to a separate PR, and keep this one purely about breaking change.

web-padawan (approved):

.github/workflows/lint.yml
bennypowers: You're right, I had thought about that too. Low risk, though.

PR #3081, opened 2026-05-31
refactor(rollup-plugin-import-meta-assets): migrate tests to node:test

bennypowers:
## Summary

- Migrate `@web/rollup-plugin-import-meta-assets` tests from mocha/chai/hanbi to `node:test` + `node:assert/strict` + `node:test` mock API
- Update `test:node` script from mocha to `node --test --test-force-exit`
- CJS test file, no TypeScript, no import path changes needed
- Replaced hanbi stub/spy API with `mock.method` / `mock.restoreAll`
- Removed `test:update-snapshots` script (no longer applicable with node:test)

## Test plan

- [x] 16/16 tests pass on Node 22.22.3
- [x] ESLint clean
- [x] Prettier clean

### DIFF
--- a/packages/rollup-plugin-import-meta-assets/test/integration.test.js
+++ b/packages/rollup-plugin-import-meta-assets/test/integration.test.js
@@ -1,7 +1,7 @@
+const { describe, it, beforeEach, afterEach, mock } = require('node:test');
+const assert = require('node:assert/strict');
 const path = require('path');
 const { rollup } = require('rollup');
-const { expect } = require('chai');
-const hanbi = require('hanbi');
 
 const { importMetaAssets } = require('../src/rollup-plugin-import-meta-assets.js');
 const {
@@ -21,11 +21,11 @@ describe('rollup-plugin-import-meta-assets', () => {
   let consoleStub;
 
   beforeEach(() => {
-    consoleStub = hanbi.stubMethod(console, 'warn');
+    consoleStub = mock.method(console, 'warn');
   });
 
   afterEach(() => {
-    hanbi.restore();
+    mock.restoreAll();
     cleanApp();
   });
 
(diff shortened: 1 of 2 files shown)


packages/rollup-plugin-import-meta-assets/test/integration.test.js
```diff
@@ -1,7 +1,7 @@
+const { describe, it, beforeEach, afterEach, mock } = require('node:test');
+const assert = require('node:assert/strict');
 const path = require('path');
 const { rollup } = require('rollup');
-const { expect } = require('chai');
-const hanbi = require('hanbi');
```
bashmish, line 4: just double checking: you are going to remove `hanbi` and other dependencies which will not be used after all refactorings, in a separate PR once all are complete?

packages/rollup-plugin-import-meta-assets/test/integration.test.js
```diff
@@ -61,10 +61,12 @@ describe('rollup-plugin-import-meta-assets', () => {
...
 
-    expect(Object.keys(chunks)).to.have.lengthOf(1);
-    expect(Object.keys(assets)).to.have.lengthOf(5);
+    assert.equal(Object.keys(chunks).length, 1);
+    assert.equal(Object.keys(assets).length, 5);
 
-    expect(chunks['app.js']).to.equal(js`
+    assert.equal(
+      chunks['app.js'],
+      js`
```
bashmish, line 69: the format is a bit off in here and I think in all other similar cases

looks like prettier doesn't like this as a one-liner

```
assert.equal(chunks['app.js'], js`
```

but then it should change the indentation of code lines which doesn't happen either

can you please check this? if you do this with copilot, maybe you can ask it to fix such indentation issues

looks like in some other PRs the indentation was refactored correctly in similar cases, e.g. in https://github.com/modernweb-dev/web/pull/3097

packages/rollup-plugin-import-meta-assets/test/integration.test.js
bennypowers: yes, after all refactor PRs are complete

packages/rollup-plugin-import-meta-assets/test/integration.test.js
bennypowers: that's just how prettier wanted it. i'm just as happy to launch prettier into the sun and never think about it again.

bashmish: prettier is not perfect when it comes to code blocks indentation, but the indentation needs to be fixed, lack of prettier won't fix it either

bennypowers: reverted the indentation changes for the content lines

if prettier doesn't like that, i think we should remove it

PR #3083, opened 2026-06-01
refactor(dev-server-legacy): migrate tests to node:test

bennypowers:
## Summary

- Migrate `@web/dev-server-legacy` tests from mocha/chai to `node:test` + `node:assert/strict`
- First TypeScript test migration in the series
- 2 TS test files, imports from `dist/` (compiled output)
- `__dirname` replaced with `import.meta.dirname`
- Local `.js` import paths updated to `.ts`

## Test plan

- [x] 20/20 tests pass on Node 22.22.3
- [x] ESLint clean
- [x] Prettier clean
- [x] Code review clean

### DIFF
--- a/packages/dev-server-legacy/test/transform-html.test.ts
+++ b/packages/dev-server-legacy/test/transform-html.test.ts
@@ -1,9 +1,10 @@
-import { expect } from 'chai';
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
 import { createTestServer } from '@web/dev-server-core/test-helpers';
 import { fetchText, expectIncludes } from '@web/dev-server-core/test-helpers';
 
-import { legacyPlugin } from '../src/legacyPlugin.js';
-import { modernUserAgents, legacyUserAgents } from './userAgents.js';
+import { legacyPlugin } from '../dist/legacyPlugin.js';
+import { modernUserAgents, legacyUserAgents } from './userAgents.ts';
 
 const htmlBody = `
 <html>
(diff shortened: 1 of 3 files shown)


packages/dev-server-legacy/test/transform-html.test.ts
```diff
@@ -1,9 +1,10 @@
-import { expect } from 'chai';
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
 import { createTestServer } from '@web/dev-server-core/test-helpers';
 import { fetchText, expectIncludes } from '@web/dev-server-core/test-helpers';
 
-import { legacyPlugin } from '../src/legacyPlugin.js';
-import { modernUserAgents, legacyUserAgents } from './userAgents.js';
+import { legacyPlugin } from '../dist/legacyPlugin.js';
+import { modernUserAgents, legacyUserAgents } from './userAgents.ts';
```
bashmish, line 8: @bennypowers I don't quite get now why we need to import from "dist" while we can import from ".ts"? your earlier reply https://github.com/modernweb-dev/web/pull/3097#discussion_r3383147222 now is not clear to me, I mean it's also a relative import, right? not sure what I am missing

bennypowers: TS 5.0 doesn't support .ts import extensions. You'd need:
  - allowImportingTsExtensions (TS 5.0+) -- but requires noEmit, so no JS output
  - rewriteRelativeImportExtensions (TS 5.7+) -- rewrites .ts -> .js in output, which is what we want, but we agreed to tackle TS after node.

PR #3086, opened 2026-06-01
refactor(polyfills-loader): migrate tests to node:test

bennypowers:
## Summary

- Migrate `@web/polyfills-loader` tests from mocha/chai to `node:test` + `node:assert/strict`
- 3 TS test files, imports from `dist/`, type-only imports for interfaces
- `__dirname` -> `import.meta.dirname`
- `expect(x).to.be.a('string')` -> `assert.equal(typeof x, 'string')`
- `--experimental-strip-types` for CI compat

## Test plan

- [x] 34/34 tests pass on Node 22.22.3 with `NODE_OPTIONS='--no-experimental-strip-types'`
- [x] ESLint clean
- [x] Prettier clean
- [x] Code review clean

### DIFF
--- a/packages/polyfills-loader/package.json
+++ b/packages/polyfills-loader/package.json
@@ -26,9 +26,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "test:node": "mocha \"test/**/*.test.{ts,js,mjs,cjs}\" --require ts-node/register --reporter dot",
-    "test:update-snapshots": "mocha \"test/**/*.test.{ts,js,mjs,cjs}\" --require ts-node/register --update-snapshots",
-    "test:watch": "mocha \"test/**/*.test.{ts,js,mjs,cjs}\" --require ts-node/register --watch --watch-files src,test"
+    "test:node": "node --experimental-strip-types --test --test-force-exit test/**/*.test.ts",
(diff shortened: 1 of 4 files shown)


packages/polyfills-loader/package.json
```diff
@@ -26,9 +26,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "test:node": "mocha \"test/**/*.test.{ts,js,mjs,cjs}\" --require ts-node/register --reporter dot",
-    "test:update-snapshots": "mocha \"test/**/*.test.{ts,js,mjs,cjs}\" --require ts-node/register --update-snapshots",
```
bashmish, line 30: just checking: since we don't have mocha snapshots anymore, we just need to update them manually?
I think we use snapshots in just 3 packages now, so it's not much work to update them

I also migrated 2 packages recently to a new solution where everything is managed in JS, becuase I found it really hard to extends tests with new use-case with folder-based approach, and debugging was very slow, so might be good to look into refactoring remaining tests to same approach

bennypowers: `node:test` has `--experimental-test-snapshots` (landed in Node 22.3). It writes .snapshot files and supports `assert.snapshot()`. That's one option.

bashmish: I made an issue for this https://github.com/modernweb-dev/web/issues/3107

PR #3091, opened 2026-06-01
refactor(test-runner-junit-reporter): migrate tests to node:test

bennypowers:
## Summary

- Migrate `@web/test-runner-junit-reporter` tests from mocha/chai to `node:test` + `node:assert/strict`
- 1 TS file, needs Chrome, XML snapshot comparison
- `import type { TestRunnerCoreConfig }` (type-only)
- Rename `test` to `test:node`
- `--experimental-strip-types` for CI compat

## Test plan

- [x] 3/3 tests pass on Node 22.22.3 with Chrome + `NODE_OPTIONS='--no-experimental-strip-types'`
- [x] ESLint clean
- [x] Prettier clean
- [x] Code review clean

### DIFF
--- a/packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
+++ b/packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
@@ -1,28 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
-  <testsuite name="Chrome_puppeteer_/packages/test-runner-junit-reporter/test/fixtures/multiple/nested-test.js" id="0" tests="1" skipped="0" errors="0" failures="0" time="<<computed>>">
+  <testsuite name="Chrome_puppeteer_/test/fixtures/multiple/nested-test.js" id="0" tests="1" skipped="0" errors="0" failures="0" time="<<computed>>">
     <properties>
-      <property name="test.fileName" value="packages/test-runner-junit-reporter/test/fixtures/multiple/nested-test.js"/>
+      <property name="test.fileName" value="test/fixtures/multiple/nested-test.js"/>
       <property name="browser.name" value="Chrome"/>
       <property name="browser.launcher" value="puppeteer"/>
     </properties>
-    <testcase name="is equivalent to map (f.g) f a" time="<<computed>>" classname="given a functor f and a function a -&gt; b map f map g f a" file="packages/test-runner-junit-reporter/test/fixtures/multiple/nested-test.js"/>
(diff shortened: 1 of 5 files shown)


packages/test-runner-junit-reporter/test/fixtures/multiple/expected.xml
```diff
@@ -1,28 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
-  <testsuite name="Chrome_puppeteer_/packages/test-runner-junit-reporter/test/fixtures/multiple/nested-test.js" id="0" tests="1" skipped="0" errors="0" failures="0" time="<<computed>>">
```
bashmish, line 3: I'm not familiar with junit reporter, so not sure how critical this is, but a little strange to see this path change

is this a result of this change in the implementation?

Image

if yes, was this intentional? maybe best to make it work in such a way that this file doesn't change in the PR

bennypowers: This is actually a fix. the new diff is correct, the old output was wrong.

You can confirm it still works with a normal project like so:

```bash
mkdir /tmp/junit-test && cd /tmp/junit-test
```

package.json:
```json
{
  "name": "junit-test",
  "private": true,
  "devDependencies": {
    "@web/test-runner": "^0.19.0",
    "@web/test-runner-junit-reporter": "^0.8.0",
    "@esm-bundle/chai": "^4.3.4-fix.0"
  }
}
```

web-test-runner.config.mjs:
```js
import { junitReporter } from '@web/test-runner-junit-reporter';

export default {
  files: 'test/**/*.test.js',
  nodeResolve: true,
  reporters: [
    junitReporter({ outputPath: './test-results.xml' }),
  ],
};
```

test/math.test.js:
```js

describe('math', () => {
  it('adds numbers', () => {
    expect(1 + 1).to.equal(2);
  });
});
```

```bash
npm install
npx wtr
cat test-results.xml
```

Output:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="Chrome_puppeteer_/test/math.test.js" ...>
    <properties>
      <property name="test.fileName" value="test/math.test.js"/>
      ...
    </properties>
    <testcase name="adds numbers" ... file="test/math.test.js"/>
  </testsuite>
</testsuites>
```

All paths are `test/...` relative to project root -- matching the updated
snapshots in this PR. The previous `packages/test-runner-junit-reporter/test/...`
paths were specific to running from a monorepo workspace root.

PR #3097, opened 2026-06-02
test(rollup-plugin-html): migrate from mocha/chai to node:test

bennypowers:
## Summary

- Migrate all 9 test files (139 tests) from mocha/chai to node:test and node:assert/strict
- Replace mocha test script with `node --experimental-strip-types --test --test-force-exit`
- Change source imports (`../src/`) to dist imports (`../dist/`)
- Use `import type` for type-only imports (TypeScript strip-types compatibility)
- Replace `__dirname` with `import.meta.dirname`

## Test plan

- [x] All 139 tests pass locally on Node 22.22.3
- [x] Lint passes (prettier formatted)
- [x] Full monorepo build succeeds
- [ ] CI green

### DIFF
--- a/packages/rollup-plugin-html/package.json
+++ b/packages/rollup-plugin-html/package.json
@@ -28,7 +28,8 @@
     "demo:mpa": "rm -rf demo/dist && rollup -c demo/mpa/rollup.config.js --watch & npm run serve-demo",
     "demo:spa": "rm -rf demo/dist && rollup -c demo/spa/rollup.config.js --watch & npm run serve-demo",
     "serve-demo": "node ../dev-server/dist/bin.js --watch --root-dir demo/dist --app-index index.html --compatibility none --open",
-    "test:node": "node --experimental-strip-types --test --test-force-exit 'test/**/*.test.ts'"
+    "test:node": "node --experimental-strip-types --test --test-force-exit 'test/**/*.test.ts'",
+    "test:watch": "node --experimental-strip-types --test --test-force-exit --watch 'test/**/*.test.ts'"
   },
   "files": [
     "*.js",
--- a/packages/rollup-plugin-html/package.json
+++ b/packages/rollup-plugin-html/package.json
@@ -28,8 +28,7 @@
     "demo:mpa": "rm -rf demo/dist && rollup -c demo/mpa/rollup.config.js --watch & npm run serve-demo",
     "demo:spa": "rm -rf demo/dist && rollup -c demo/spa/rollup.config.js --watch & npm run serve-demo",
     "serve-demo": "node ../dev-server/dist/bin.js --watch --root-dir demo/dist --app-index index.html --compatibility none --open",
-    "test:node": "mocha test/**/*.test.ts --require ts-node/register --reporter dot",
-    "test:watch": "mocha test/**/*.test.ts --require ts-node/register --watch --watch-files src,test"
+    "test:node": "node --experimental-strip-types --test --test-force-exit 'test/**/*.test.ts'"
   },
   "files": [
     "*.js",
--- a/packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
+++ b/packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
@@ -1,7 +1,8 @@
-import { expect } from 'chai';
+import assert from 'node:assert/strict';
+import { describe, it, afterEach } from 'node:test';
 import { parse } from 'parse5';
 import path from 'path';
-import { extractAssets } from '../../../../src/input/extract/extractAssets.js';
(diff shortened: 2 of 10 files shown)


packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
```diff
@@ -1,7 +1,8 @@
-import { expect } from 'chai';
+import assert from 'node:assert/strict';
+import { describe, it, afterEach } from 'node:test';
 import { parse } from 'parse5';
 import path from 'path';
-import { extractAssets } from '../../../../src/input/extract/extractAssets.js';
+import { extractAssets } from '../../../../dist/input/extract/extractAssets.js';
```
bashmish, line 5: why do you import from `dist` now?
I saw you did similar change in other PRs, but not 100% clear why

packages/rollup-plugin-html/package.json
```diff
@@ -28,8 +28,7 @@
     "demo:mpa": "rm -rf demo/dist && rollup -c demo/mpa/rollup.config.js --watch & npm run serve-demo",
     "demo:spa": "rm -rf demo/dist && rollup -c demo/spa/rollup.config.js --watch & npm run serve-demo",
     "serve-demo": "node ../dev-server/dist/bin.js --watch --root-dir demo/dist --app-index index.html --compatibility none --open",
-    "test:node": "mocha test/**/*.test.ts --require ts-node/register --reporter dot",
-    "test:watch": "mocha test/**/*.test.ts --require ts-node/register --watch --watch-files src,test"
```
bashmish, line 32: I saw you kept "watch" in other PRs, any specific reason to remove it here?

packages/rollup-plugin-html/test/src/output/css.test.ts
```diff
@@ -51,38 +52,41 @@ describe('replacePlaceholders', () => {
       return undefined;
     };
     const result = replacePlaceholders(css, resolver);
-    expect(result).to.include("url('assets/image1.png')");
-    expect(result).to.include("url('assets/image2.png')");
+    assert.ok(result.includes("url('assets/image1.png')"));
+    assert.ok(result.includes("url('assets/image2.png')"));
```
bashmish: chai gives nice output in such case which
- increased debugging locally
- help understand what failed in the pipeline, especially if it's executed on Windows

I think this is very important and I'd prefer to have a better alternative to this

I quickly checked possibilities and asked copilot and this is the first idea to start this discussion: create a reusable helper like this and put it into `test-utils/assert.js`

```js
function assertIncludes(actual, expected) {
  if (!actual.includes(expected)) {
    const preview = actual.length > 200
      ? actual.slice(0, 200) + '...'
      : actual;

    throw new assert.AssertionError({
      message:
        `Expected substring not found.\n\n` +
        `Expected:\n${expected}\n\n` +
        `Actual (preview):\n${preview}`,
      actual,
      expected,
      operator: 'includes'
    });
  }
}
```

packages/rollup-plugin-html/package.json
bennypowers: will fix

packages/rollup-plugin-html/test/src/output/css.test.ts
bennypowers: we already export `expectIncludes` and `expectNotIncludes`, i can refactor to use that

packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
bennypowers: this requires TypeScript 5.7+ with _rewriteRelativeImportExtensions: true_. That lets source files use .ts extensions in imports. we determined in chat that this migration is not worth doing until after node 22 EOL

packages/rollup-plugin-html/package.json
bennypowers: PTAL 7f524d2e

packages/rollup-plugin-html/test/src/output/css.test.ts
bennypowers: PTAL 7f524d2e

packages/rollup-plugin-html/test/src/input/extract/extractAssets.test.ts
bashmish: got it
then I think we need to run `build` always before tests, shouldn't this be part of the change too? or is that smth for the future PR when all PRs are complete?

bennypowers: it runs before tests in ci

local dev and test watch will remain broken until we finish the job and point tests at src

afaict there's no satisfying way to maintain the current dx while doing this upgrade piecemeal

PR #3098, opened 2026-06-02
test(dev-server-core): migrate from mocha/chai to node:test

bennypowers:
## Summary

- Migrate all 14 test files (97 tests) from mocha/chai/hanbi to node:test/node:assert/strict
- Replace hanbi `stubMethod` with `mock.method` from node:test
- Update test script from mocha to `node --experimental-strip-types --test`
- Update published `src/test-helpers.ts` to use node:assert instead of chai (breaking change, changeset included)

## Breaking change

`src/test-helpers.ts` is exported via `exports["./test-helpers"]`. Its `fetchText` function now throws `node:assert`'s `AssertionError` instead of chai's on non-200 responses. Functionally identical, different error class. Pending team decision on acceptability.

## Test plan

- [x] All 97 tests pass on Node 22
- [x] All 97 tests pass on Node 24
- [ ] CI green

### DIFF
--- a/.changeset/brave-bags-march.md
+++ b/.changeset/brave-bags-march.md
@@ -1,5 +1,0 @@
----
-'@web/dev-server-core': major
----
-
-Replace `chai` with `node:assert/strict` in the published `test-helpers` module. The `fetchText` helper now throws `node:assert`'s `AssertionError` instead of `chai`'s on non-200 responses.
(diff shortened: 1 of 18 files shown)


bashmish:
@bennypowers I made a couple of changes, quite small, but the one that removes the major update changset I think you'd like to check. Let me know if you agree with the changes.

I also have another idea about test-helpers, but it's best to discuss it in a separate PR which I plan to create when this one is merged. Just giving you heads-up upfront that it might not be the final state before we release all of this.

bennypowers:
@bashmish LGTM

PR #3105, opened 2026-06-13
fix: Added missing package.json fields: bugs, homepage

Omission-create:
Auto-fix

Added missing package.json fields: bugs, homepage

Related: #1369


web-padawan:
The root-level package.json has `"private": true` and isn't published, so these fields are not relevant. Closing.

PR #3106, opened 2026-06-13
fix: add missing bugs metadata field to package.json

BWM0223:
Adds missing package metadata fields to `package.json`.

This allows npm tooling and consumers to correctly resolve package metadata.


web-padawan:
The root-level package.json has `"private": true` and isn't published, so this field is not relevant. Closing.

PR #3124, opened 2026-07-02
chore: improve prettier setup

bashmish:
## What I did

1. upgrade prettier to latest
2. make sure all embedded languages are formatted, especially js template literals used in tests
3. make sure imports/exports are consistently ordered
    - there was an incorrect import leading to circular dependencies which I found when formatting imports/exports, so fixed that too
4. exclude snapshots from formatting (also renamed some snapshots to follow same naming convention and apply same filter to exclude them)


thepassle (approved):

thepassle:
@bashmish some failing tests still

bashmish:
@thepassle I saw that
- Node 22 passed
- Node 24 is expected to fail, unrelated to this PR and we need to look into this separetely, we also discussed it yesterday
- Windows passed (0 failures in logs), but the build is red, which happens sometimes and I couldn't find out why that happens (not related to this PR), sometimes I restart them and then it's green on the 2nd try, quite unstable in that regard

Not sure which tests you saw failing, but I'll proceed with merging for now.

PR #3127, opened 2026-07-03
fix: clean up test helpers

bashmish:
## What I did

1. Removed test helpers `timeout`, `fetchText`, `expectIncludes`, `expectNotIncludes`, `virtualFilesPlugin` from `@web/dev-server-core/test-helpers`.
    - but kept `createTestServer` in it as it seems really useful and related to this package
2. Locally exposed there test helpers via global `test-helpers/node-test-helpers.js` module.
3. Other test utils were moved to this `test-helpers/node-test-helpers.js` module too (basically I renamed the old filename, that's all).
4. Renamed `expectIncludes` and `expectNotIncludes` to `assertIncludes` and `assertNotIncludes` to better reflect it's not chai anymore and that we use native tools like `node:assert`.


thepassle:
Nitpick: 
`test-helpers/node-test-helpers.js`
⬇️
`test-helpers/node.js`

